PR #56, opened 2021-03-05
Allow for us to send logs, metrics and traces to a local splunk instance additionally to cloud

atoulme:
This allows to set additional outputs for logs, metrics and traces that would duplicate data to go to a local Splunk Enterprise.

Right now, you have to choose to send logs to a local cloud or the remote ingest. This allows to do both.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -137,6 +137,28 @@ exporters:
     sync_host_metadata: true
   {{- end }}
 
+  {{- if .Values.splunkHEC.metrics.enabled }}
+  splunk_hec/metrics:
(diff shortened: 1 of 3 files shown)


dmitryax:
Hi @atoulme, based on the recent internal discussions, this chart is not going to explicitly support both Observability and Enterprise backends. Even if we don't want to add extra configurations to support the enterprise endpoint, it's still possible to set up the forked output with existing functionality. The following values should help to achieve the goal:
```
otelAgent:
  config:
    exporters:
      splunk_hec/metrics:
        token: <Your Splunk Enterprise HEC token>
        endpoint: <Splunk Enterprise endpoint>
        sourcetype: ...
        index: ...
        insecure_skip_verify: ...
    service:
      pipelines:
      exporters:
        - signalfx
        - splunk_hec/metrics 
```

atoulme:
Closing.

PR #57, opened 2021-03-05
Remove `v2/datapoint` from end of ingest_url

jrcamp:
With `v2/datapoint` set on the end it results in trying to send events to
`/v2/datapoint/v2/event` which is not correct. Removing the path will result in
the right suffix being used for metrics and events.


rmfitzpatrick (approved):

rmfitzpatrick:
This seems like undesired default event client behavior in the SFx exporter.  If v2/datapoint were detected it could replace w/ v2/event.

PR #76, opened 2021-03-26
update to new receivercreator rule syntax

jrcamp:


dmitryax:
Do we need to bump otel collector version with this or 0.21.1 already support this syntax?

jrcamp:
Hmm yeah I should bump it here. I guess I hit it since I had overridden the image version to newer version.

dmitryax (approved):

PR #77, opened 2021-03-26
add eks and gke detectors

jrcamp:


jrcamp:
I need to bump contrib in distro as well.

asuresh4 (approved):

PR #80, opened 2021-04-01
Add provider and distro options

jrcamp:
The previous option of platform has been split into two separate options,
provider and distro. This allows us to distinguish between cloud-managed
clusters and user-managed clusters.

The reason " " (with a space) is allowed as an option is to work around UI
wizard that is not able to omit the option or provide just an empty space.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -86,13 +86,22 @@ processors:
   # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor
   resourcedetection:
     detectors:
+      # Note: Kubernetes distro detectors need to come first so they set the proper cloud.platform
+      # before it gets set later by the cloud provider detector.
       - env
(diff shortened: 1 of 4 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -86,13 +86,22 @@ processors:
...
-      - gce
-      {{- else if eq .Values.platform "aws" }}
+      {{- else if eq (.Values.distro | default "") "eks" }}
       - eks
+      {{- else if eq (.Values.distro | default "") "aks" }}
+      - aks
+      {{- end }}
+      {{- if eq (.Values.provider | default "") "gcp" }}
+      - gce
+      {{- else if eq (.Values.provider | default "") "aws" }}
```
dmitryax, line 101: can we set `""` as the default value in values.yaml instead of `| default ""`?

jrcamp: Yeah I was thinking about case where somebody did

```yaml
provider:
```

Which would set it to null but now I realize having it in the schema as required and non-nullable would prevent that. Seems to work.

flands (approved):
LGTM

PR #81, opened 2021-04-05
Set daemonset collector agent's k8s resource attributes by env

rmfitzpatrick:
These changes add the three resource attributes generally set by the receiver creator for all resource detection processor users: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/aa9e5432748c83db8701f53c3a569f2e3c2559b1/receiver/receivercreator/factory.go#L54.  This way all other receiver types can benefit from these identifiers.

Also updates to no longer use the deprecated env var.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -103,6 +103,8 @@ processors:
       {{- else if eq .Values.provider "azure" }}
       - azure
       {{- end }}
+    # Don't override existing resource attributes to maintain identification of data sources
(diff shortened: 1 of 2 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -103,6 +103,7 @@ processors:
       {{- else if eq .Values.provider "azure" }}
       - azure
       {{- end }}
+    override: false
```
jrcamp, line 106: Can you add a comment explaining why we decided to set to false?

jrcamp (approved):

PR #86, opened 2021-04-07
Add host.name by default to fluentd logs

jrcamp:
Related content uses host.name and previously this was being added by a
org-only index rule.


dmitryax (commented):
LGTM

dmitryax (approved):

PR #88, opened 2021-04-09
Fix issue with traces getting agent's k8s.pod.uid

jrcamp:
In change 3c312b6f4 we added some globally set resource attributes like
k8s.pod.uid. The problem is these attributes would get set before k8s_tagger
was run so you would get traces with the k8s.pod.uid of the agent itself
instead of the pod the trace came from.

Instead add a separate metrics pipeline that adds the agent k8s metadata only
to prometheus metrics.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/daemonset.yaml
+++ b/helm-charts/splunk-otel-collector/templates/daemonset.yaml
@@ -152,10 +152,6 @@ spec:
             value: /hostfs/run
           - name: HOST_DEV
             value: /hostfs/dev
-
-          # Host specific resource attributes
-          - name: OTEL_RESOURCE_ATTRIBUTES
-            value: host.name=$(K8S_NODE_NAME),k8s.node.name=$(K8S_NODE_NAME)
           {{- end }}
 
         readinessProbe:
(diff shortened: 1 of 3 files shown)


helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -155,7 +155,7 @@ spec:
 
           # Host specific resource attributes
           - name: OTEL_RESOURCE_ATTRIBUTES
-            value: host.name=$(K8S_NODE_NAME),k8s.node.name=$(K8S_NODE_NAME),k8s.pod.name=$(K8S_POD_NAME),k8s.pod.uid=$(K8S_POD_UID),k8s.namespace.name=$(K8S_NAMESPACE)
+            value: host.name=$(K8S_NODE_NAME),k8s.node.name=$(K8S_NODE_NAME)
```
rmfitzpatrick, line 158: I think it's desirable to break these out into its own resource processor so that we aren't preventing straightforward usage for `OTEL_RESOURCE_ATTRIBUTES` in all deployments.

dmitryax (approved):

helm-charts/splunk-otel-collector/templates/daemonset.yaml
jrcamp: See what you think about the update.

rmfitzpatrick (approved):

PR #90, opened 2021-04-12
Enable batch for metrics

### LINKED ISSUE #32: Enable batching for metrics
Batching is currently only enabled for traces. Should enable for metrics.

https://github.com/signalfx/o11y-collector-for-kubernetes/blob/50e3132c59aeb6e96fa2cb67803c7fa47c3e5cfb/helm-charts/o11y-collector/templates/config/_otel-collector.tpl#L79

https://github.com/signalfx/o11y-collector-for-kubernetes/blob/master/helm-charts/o11y-collector/templates/config/_otel-agent.tpl#L164

Thanks @mstumpfx for finding.

jrcamp:
Use default config for batch (send every 200ms or 8192 items). This is same
that
https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/config/collector/gateway_config.yaml
is using.

Resolves #32

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -111,8 +111,6 @@ processors:
   {{- include "splunk-otel-collector.otelMemoryLimiterConfig" .Values.otelAgent | nindent 2 }}
 
   batch:
-    timeout: 200ms
-    send_batch_size: 128
 
   resource:
     # General resource attributes that apply to all telemetry passing through the agent.
@@ -188,13 +186,13 @@ service:
       receivers: [otlp, jaeger, zipkin]
       processors:
         - memory_limiter
+        - batch
         - resource
         - resourcedetection
         - k8s_tagger
         {{- if .Values.environment }}
         - resource/add_environment
         {{- end }}
-        - batch
       exporters:
         {{- if .Values.otelCollector.enabled }}
         - otlp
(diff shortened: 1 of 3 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -111,8 +111,6 @@ processors:
   {{- include "splunk-otel-collector.otelMemoryLimiterConfig" .Values.otelAgent | nindent 2 }}
 
   batch:
```
dmitryax, line 113: I remember facing issues with null values in helm. Can we set `batch: {}` instead?

dmitryax: AFAIR you cannot override them in user's values files. Feel free to try, I may be wrong

jrcamp: Seems to work fine merging in values from otelAgent.processors.batch. We do the same thing with zpages extension that was already there so I think you're thinking of a different issue.

jrcamp: I think you might be thinking of a situation where in values.yaml we had to do `foo: {}` instead of `foo:` due to some reason.

dmitryax: Ok, maybe. I don't remember exact reason

dmitryax (approved):

Issue #54, opened 2021-03-04
BUG: Readme does not tell where to get splunkRealm and splunkAccessToken from


tigrannajaryan:
"How to install" section should have a link to the docs or product page where I can get the realm and access token from.

PR #92, opened 2021-04-13
Update README

flands:
- Make consistent with splunk-otel-collector
- Also add icon to chart


dmitryax (approved):

dmitryax:
Looks like this change should also resolve https://github.com/signalfx/splunk-otel-collector-chart/issues/54

PR #97, opened 2021-04-14
Use clusterName to work with gdi ui logs inventory

rmfitzpatrick:
`k8s.cluster.name` isn't used by the wizard's* log queries so we need to work around this.


jrcamp (changes requested):
I think the UI should be updated instead.

rmfitzpatrick:
Accepted by ui team.

PR #98, opened 2021-04-15
Use signalfx-forwarder in default agent trace pipeline

rmfitzpatrick:
This was added to the [default agent config](https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/config/collector/agent_config.yaml#L68), so adopting here

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -57,6 +57,10 @@ receivers:
     extra_metadata_labels:
       - container.id
 
+  smartagent/signalfx-forwarder:
+    type: signalfx-forwarder
+    listenAddress: 0.0.0.0:9080
+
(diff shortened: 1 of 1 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -57,6 +57,9 @@ receivers:
     extra_metadata_labels:
       - container.id
 
+  smartagent/signalfx-forwarder:
+    type: signalfx-forwarder
+
```
asuresh4, line 62: Should `listenAddress` also be configured to match the default agent config (`0.0.0.0:9080`)?

rmfitzpatrick: I wasn't sure tbh since this isn't in the SA config map, but if that's the default everything else uses makes sense.

asuresh4 (approved):

PR #106, opened 2021-04-18
Expose signalfx-forwarder on agent

flands:


dmitryax (approved):

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -110,6 +110,10 @@ otelAgent:
       containerPort: 4317
       hostPort: 4317
       protocol: TCP
+    signalfx-forwarder:
```
dmitryax, line 113: Oh, this name must be less than 16 characters. There is k8s limit on port names. Please rename

dmitryax: the test caught it
```
Error: DaemonSet.apps "splunk-otel-collector-lxt5lfzceb-agent" is invalid: spec.template.spec.containers[1].ports[4].name: Invalid value: "signalfx-forwarder": must be no more than 15 characters
```

dmitryax (approved):

PR #114, opened 2021-04-29
Add an option to disable any of the ports exposed by default

dmitryax:


jrcamp (approved):
I guess we have these enabled fields for other things but I wonder if we could have a more general rule where if it's empty then it's disabled, e.g.:

```
ports:
  zipkin: nil
```

Not sure if `if $port` would evaluate correct in templates for `nil` and `{}` or not. 🤔 

Consider it but if you don't think it's any better then this is ok.

dmitryax:
> Consider it but if you don't think it's any better then this is ok.

Actually `null` does work with the latest helm version. I remember it didn't work with older versions. I'll try this more...

dmitryax:
This can be achieved by `enabled_for` option introduced here https://github.com/signalfx/splunk-otel-collector-chart/pull/144 

Closing this PR

PR #122, opened 2021-05-04
Remove monitor_agent events

dmitryax:
Internal metrics already exposed via Prometheus that should be used for monitoring instead of emitting monitor_agent events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+## [0.24.13] - 2021-05-04
+
+### Changed
+
+- Remove internal fluentd metrics sent as logs with monitor_agent. Prometheus
+  metrics exposed on 0.0.0.0:24231 should be used instead (#122)
+
 ## [0.24.12] - 2021-05-03
 
 ### Fixed
--- a/helm-charts/splunk-otel-collector/Chart.yaml
+++ b/helm-charts/splunk-otel-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: splunk-otel-collector
-version: 0.24.12
+version: 0.24.13
 description: Splunk OpenTelemetry Connector for Kubernetes
 icon: https://github.com/signalfx/splunk-otel-collector-chart/tree/main/splunk.png
 type: application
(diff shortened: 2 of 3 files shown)


jrcamp:
Did you verify that prometheus isn't depending on monitor_agent being configured? It's in their sample config but I don't know if it's required or just a coincidence: https://github.com/fluent/fluent-plugin-prometheus/blob/master/misc/fluentd_sample.conf

If it's required I think we can remove the tag config option to make it stop emitting events: https://docs.fluentd.org/input/monitor_agent#tag

helm-charts/splunk-otel-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: splunk-otel-collector
-version: 0.24.12
+version: 0.24.13
```
rmfitzpatrick, line 3: There are 5 PRs open at this time.  What is the reason to include a version bump along with the change itself as opposed to a separate release commit?

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+## [0.24.13] - 2021-05-04
+
+### Changed
+
+- Remove monitor_agent events (#122)
```
rmfitzpatrick, line 11: Would you mind extrapolating this a bit?  I don't see it documented and it's unclear to me as a user what effect this would have.

dmitryax:
> If it's required I think we can remove the tag config option to make it stop emitting events: https://docs.fluentd.org/input/monitor_agent#tag
@jrcamp I tested it. it's not required

helm-charts/splunk-otel-collector/Chart.yaml
dmitryax: In the helm chart, we don't follow the rule to have a release commit so far. I don't see a reason to do that at the moment. Just to introduce a change that can be utilized right away. Git tags can be used to track history of releases instead of release commits.

@jrcamp what's your thoughts?

CHANGELOG.md
dmitryax: Good point. updated

helm-charts/splunk-otel-collector/Chart.yaml
dmitryax: But I'm open to having release commits if everyone agrees

jrcamp: I think bumping the version in the same PR is fine. If there are multiple changes ready to go out it would be nice to group them into the same release just to avoid superfluous releases. I doubt it's bothering anybody though.

jrcamp (approved):

PR #127, opened 2021-05-06
Disable receivers and pipelines if telemetry type disabled

jrcamp:
Testing in progress.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -19,7 +19,6 @@ receivers:
     endpoint: 0.0.0.0:8006
   {{- end }}
 
-  {{- if .Values.metricsEnabled }}
   # Prometheus receiver scraping metrics from the pod itself
   prometheus/agent:
     config:
@@ -32,6 +31,7 @@ receivers:
           # Fluend metrics collection disabled by default
           # - "${K8S_POD_IP}:24231"
 
+  {{- if .Values.metricsEnabled }}
   hostmetrics:
     collection_interval: 10s
     scrapers:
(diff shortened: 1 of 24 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -14,9 +14,8 @@ extensions:
 
 receivers:
   {{- include "splunk-otel-collector.otelTraceReceivers" . | nindent 2 }}
-  fluentforward:
-    endpoint: 0.0.0.0:8006
```
dmitryax, line 18: It should not be disabled on the agent right?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -14,9 +14,8 @@ extensions:
...
 
+  {{- if .Values.metricsEnabled }}
   # Prometheus receiver scraping metrics from the pod itself
   prometheus/agent:
```
dmitryax, line 20: I think this receiver should be enabled all the time since it's meant to be used for troubleshooting and these metrics are part of APM bundle currently

dmitryax:
what about host ports? Is the `enabled_for` thing not needed?

jrcamp:
> what about exposed host ports? Is the `enabled_for` thing not needed?

Not from what I've seen. Though would be good for it to be correct just so it's not misleading to user. I'll file an issue for it if it turns out not to be needed functionally.

dmitryax:
Let rather file an issue with an option to disable hostNetwork. I assume it'll be needed in that case, right?

dmitryax (approved):

jrcamp:
> Let's rather file an issue to add an option to disable hostNetwork. I assume it'll be needed in that case, right?

#133. I think it'll only be necessary for environments that don't have host access (maybe Fargate?)

dmitryax:
> #133. I think it'll only be necessary for environments that don't have host access (maybe Fargate?)

PR #134, opened 2021-05-11
Add receiver:k8scluster dimension to k8s cluster metrics

jrcamp:


mstumpfx:
I wonder if there’s a more useful / better identifier dimension we could use instead of `unused:1`? If we're adding a new global dimension, might as well make it something more useful? Maybe something like `agent_version` or something along those lines?

jrcamp:
> I wonder if there’s a more useful / better identifier dimension we could use instead of `unused:1`? If we're adding a new global dimension, might as well make it something more useful? Maybe something like `agent_version` or something along those lines?

agent_version would be useful but we don't want these MTS to churn every time they upgrade though :( Also it's only for k8s metrics. There's probably *something* that's better but I lack imagination to think of it.

dmitryax (approved):

mstumpfx (approved):

PR #151, opened 2021-05-25
Add an option to mount extra volumes to the daemonset

dmitryax:
This change allows adding any kind of volume to the deamonset


flands (approved):
LGTM

PR #163, opened 2021-06-15
Make auto-detection of prometheus metrics disabled by default

dmitryax:
Add an additional "autodetect.prometheus" flag to enable automatic prometheus endpoint detection and make it disabled by default.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -6,11 +6,9 @@ The values can be overridden in .Values.otelAgent.config
 extensions:
   health_check:
 
-{{- if .Values.autodetect.prometheus }}
   k8s_observer:
     auth_type: serviceAccount
     node: ${K8S_NODE_NAME}
-{{- end }}
 
   zpages:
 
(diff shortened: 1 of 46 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -6,9 +6,11 @@ The values can be overridden in .Values.otelAgent.config
 extensions:
   health_check:
 
+{{- if .Values.autodetect.prometheus }}
   k8s_observer:
```
jrcamp, line 10: I think we should still configure the observer by default as people have configured their own discovery rules to start things other than prometheus so I don't think it should be tied to this flag.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -49,6 +51,7 @@ receivers:
       # System processes metrics, disabled by default
       # process:
 
+  {{- if .Values.autodetect.prometheus }}
```
jrcamp, line 54: Same here, I think the flag should only control adding a prometheus_simple receiver entry but receiver_creator and k8s_observer should stay.

jrcamp (changes requested):

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: sounds good. changed

jrcamp (approved):

PR #171, opened 2021-07-07
Adjust configuration for istio environment

dmitryax:
Add new config option autodetect.istio, which, if enabled, adds few things to the default configuration:
- Add `service.name` attribute to logs that are composed the same way as done for traces in istio. This enables APM <-> Log Observer correlation OOTB.
- Add automatic istio metrics scraping that are exposed via prometheus endpoints.
- Add `sidecar.istio.io/inject: "false"` annotation to the otel-collector pods to avoid infinite self-tracing.


flands (approved):
LGTM -- I am not sure disabling Istio injection will work for all environments, but likely good enough for now. We should document what these settings do better.

PR #172, opened 2021-07-07
Switch port for internal metrics from 8888 to 8889

dmitryax:
GKE has their own otel-collector installed by default as agent with hostNetwork enabled and using the same default port "8888" for collector's metrics. We have to switch to another port to avoid conflicts in GKE.


flands (approved):
Update change log?

PR #176, opened 2021-07-10
Fix generation of service.name log attribute in istio environment

dmitryax:
Istio has an involved logic for generating service name for traces. So we cannot just rely on "app" pod label. This commit adds the same logic to generate service.name logs attribute as istio uses for traces. It's needed to make correlation between logs and traces work OOTB in Splunk o11y suite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+## [0.29.1] - 2021-07-08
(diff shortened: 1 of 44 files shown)


flands (changes requested):
Ouch... I mean looks good, however this is a custom (not built in parser) right? I would assume this plugin needs to be installed so installer scripts need to be updated. Also unsure how this will work on upgrade...

dmitryax:
@flands this plugin is already installed in the image and being used

flands (approved):
Sweet

PR #181, opened 2021-07-29
Template pod affinity settings from the correct value

KingJ:
Pod affinity settings were erroneously reading from the tolerations value rather than the affinity value for both the collector and cluster receiver deployments. As a result, setting the affinity value would result in an affinity key being templated but with the contents of the tolerations value.

This fix ensures the correct value is read, and also adds a leading dash to the template function to prevent an erroneous newline from being added.


rmfitzpatrick (approved):

dmitryax:
@KingJ thank you for fixing this!

PR #182, opened 2021-08-03
Made Fluentd pod init container image variable

spr-mweber3:


dmitryax:
@spr-mweber3 thank you for the contribution!

Issue #189, opened 2021-09-03
Capture some pod labels by default


mtwo:
A common use case that I've heard at customer meetings is applying a common set of pod labels as dimensions / annotations on metrics, traces, and logs. Users can custom-define these labels, and there are often too many to efficiently apply all of them, but it would be nice if the Collector captured some common / built-in ones. For example, Datadog's agent does a nice job of this: https://docs.datadoghq.com/agent/kubernetes/tag/?tab=containerizedagent

mstumpfx:
Hi @mtwo 

By default, at least for metrics, we do capture labels as [signalfx properties](https://docs.signalfx.com/en/latest/metrics-metadata/metrics-metadata.html#properties) that are synced onto resource UID dimensions.

For example, for pod labels we sync this onto the `k8s.pod.uid` dimension. In the signalfx product you should be able to navigate to the Menu --> Metrics --> Metadata Finder --> search for your k8s UID by typing `k8s.pod.uid: $your_pod_uid` and you should be able to find a list of properties synced onto this dimension.

The way this works in the code (since this is undocumented sort of I'll write some of this out) is that a receiver may implement a [metadata exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/e1914a8d6a07a398c5d8445b818448f70bdbc43a/receiver/k8sclusterreceiver/README.md#metadata_exporters). The Signalfx exporter implements a [consume metadata](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/e1914a8d6a07a398c5d8445b818448f70bdbc43a/exporter/signalfxexporter/exporter.go#L55) function which ends up exporting this metadata to our backend.
In the k8s cluster receiver, each resource type is watched for updates and when updated specific sets of metadata is synced which you can [view here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/e1914a8d6a07a398c5d8445b818448f70bdbc43a/receiver/k8sclusterreceiver/internal/collection/collector.go#L165). You can use that function as a jumping off point to see what properties we sync for various resources onto what dimensions.

Specifically for pod labels, you can see this is [added here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/e1914a8d6a07a398c5d8445b818448f70bdbc43a/receiver/k8sclusterreceiver/internal/collection/pods.go#L145) along with some other properties, such as the owners reference of the pod (e.g. Daemonset, StatefulSet ...)

[...]

PR #192, opened 2021-09-14
Move k8s metadata enrichment from fluentd to otel-collector

dmitryax:
This change is the first step towards moving away from fluentd logs collection.

Fluentd "kubernetes_metadata_filter" plugin is known as a performance bottleneck in k8s logs collection with fluentd. So moving away from it already gives performance benefits. Also it mitigates issues when fluentd ends up hammering k8s API. 

This change is compatible with otel gateway, if `otelCollector.enabled=true` all logs are forwarded through the gateway and logs enrichment is happening there reducing load on k8s API.

Drawbacks:

- `container.image.name` logs attribute cannot be provided at the moment until this issue is resolved: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/5235

Additional changes:

- Extra attribute added by default `k8s.pod.labels.app`. Value for this attribute is taken from pod's "app" label if it's set. The change is needed to support istio use case. We also want to enable more of these attributes going forward https://github.com/signalfx/splunk-otel-collector-chart/issues/189.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Move k8s metadata enrichment from fluentd to otel-collector (#191)
+- Move k8s metadata enrichment from fluentd to otel-collector (#192)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Changed
+
+- Move k8s metadata enrichment from fluentd to otel-collector (#191)
+
 ### Fixed
 
 - smartagent/kubernetes-events now works with `otelK8sClusterReceiver` deployment. (#187)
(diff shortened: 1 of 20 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -531,9 +531,9 @@ image:
 
   otelcol:
     # The registry and name of the opentelemetry collector image to pull
-    repository: quay.io/signalfx/splunk-otel-collector
+    repository: quay.io/signalfx/splunk-otel-collector-dev
```
emaderer, line 534: Why are pointing to dev? wouldn't that be unstable?

dmitryax: I'm not going to release the chart yet. I'll bump the version to the next stable one with the chart release. There is functionality here that depends on the latest changes in https://github.com/signalfx/splunk-otel-collector

emaderer (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Changed
+
+- Move k8s metadata enrichment from fluentd to otel-collector (#191)
```
rmfitzpatrick, line 9:
```suggestion
- Move k8s metadata enrichment from fluentd to otel-collector (#192)
```

rockb1017 (approved):
Looks great

CHANGELOG.md
dmitryax: good catch. thanks!

PR #197, opened 2021-09-22
Add optional native OTel logs collection

dmitryax:
We want to replace fluentd logs collection with native OTel logs collection. For now we adding this as an option. Later it will became default way to collect logs

The configuration for native logs collection mostly borrowed from https://github.com/splunk/sck-otel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Added
+
+- Add native OTel logs collection as an option (#197)
+
 ### Removed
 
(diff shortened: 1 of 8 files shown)


emaderer:
@dmitryax  are there any differences in terms of what it will end up collecting if users enable native log collection?

emaderer:
> @emaderer yes, native logs collection is missing the following attibutes: `container.id` and `container.image.name`. We will keep it optional until [this issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/5235) is resolved and we can add those attributes back

I think that it might be a good idea to update the README file as part of this PR and mention the limitations of using the new option.

dmitryax:
@emaderer good point. I added a section to README, PTAL

emaderer (approved):

PR #224, opened 2021-10-13
Fix Values configuration references in comments

lindhe:
I assume these where old/broken references to the templates/config/*.tpl files. If I'm wrong, I think the comments should be clearer about what files they are referring to.


dmitryax:
Hi @lindhe. This is right correction. Thank you!

PR #232, opened 2021-10-18
Reduce scope of host mounted volumes on linux systems

dmitryax:
Reduce scope of host paths for mounted volumes to the minimum required by hostmetrics receiver - gopsutil library.

This change allows us to run the Splunk OTel Collector in k8s environments where mounting to host "/" path is prohibited.


emaderer (approved):

rmfitzpatrick:
Do we need `HOST_**` env vars to be set with these changes?

PR #258, opened 2021-10-29
field name compatibility for SCK

rockb1017:
backward compatible field names for container metadata.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Added
+
+- Field name compatibility for SCK and index routing precedence (#258)
+
 ## [0.37.1] - 2021-11-01
 
 ### Added
 
 - Add initContainer for log checkpoint migration from Fluentd to Otel agent (#253)
 - Add index routing for Splunk Enterprise/Cloud customers (#256)
-- Field name compatibility for SCK and index routing precedence (#258)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Add initContainer for log checkpoint migration from Fluentd to Otel agent (#253)
 - Add index routing for Splunk Enterprise/Cloud customers (#256)
+- Field name compatibility for SCK and index routing precedence (#258)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Field name compatibility for SCK and index routing precedence (#258)
+- Field name compatibility for SCK (#258)
 
 ## [0.37.1] - 2021-11-01
 
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -109,7 +109,7 @@ resource/logs:
       from_attribute: k8s.cluster.name
       action: upsert
     - key: container_id
-      from_attribute: k8s.container.id
+      from_attribute: container.id
       action: upsert
     - key: pod
       from_attribute: k8s.pod.name
@@ -125,13 +125,12 @@ resource/logs:
       from_attribute: {{ printf "k8s.pod.labels.%s" $label }}
       action: upsert
     {{- end }}
-    {{- end }}
     {{- if not .Values.splunkPlatform.fieldNameConvention.keepOtelContention }}
     - key: k8s.container.name
       action: delete
     - key: k8s.cluster.name
       action: delete
-    - key: k8s.container.id
+    - key: container.id
       action: delete
     - key: k8s.pod.name
       action: delete
@@ -142,6 +141,7 @@ resource/logs:
     {{- range $_, $label := .Values.extraAttributes.podLabels }}
     - key: {{ printf "k8s.pod.labels.%s" $label }}
       action: delete
+    {{- end }}
     {{- end }}
     {{- end }}
 {{- end }}
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -125,7 +125,7 @@ resource/logs:
       from_attribute: {{ printf "k8s.pod.labels.%s" $label }}
       action: upsert
     {{- end }}
-    {{- if not .Values.splunkPlatform.fieldNameConvention.keepOtelContention }}
+    {{- if not .Values.splunkPlatform.fieldNameConvention.keepOtelConvention }}
     - key: k8s.container.name
       action: delete
     - key: k8s.cluster.name
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -101,6 +101,49 @@ resource/logs:
     - key: istio_service_name
       action: delete
     {{- end }}
+    {{- if .Values.splunkPlatform.fieldNameConvention.renameFieldsSck }}
+    - key: container_name
+      from_attribute: k8s.container.name
+      action: upsert
+    - key: cluster_name
+      from_attribute: k8s.cluster.name
+      action: upsert
+    - key: container_id
+      from_attribute: k8s.container.id
+      action: upsert
+    - key: pod
+      from_attribute: k8s.pod.name
+      action: upsert
+    - key: pod_uid
+      from_attribute: k8s.pod.uid
+      action: upsert
+    - key: namespace
+      from_attribute: k8s.namespace.name
+      action: upsert
+    {{- range $_, $label := .Values.extraAttributes.podLabels }}
+    - key: {{ printf "label_%s" $label }}
+      from_attribute: {{ printf "k8s.pod.labels.%s" $label }}
+      action: upsert
+    {{- end }}
+    {{- end }}
+    {{- if not .Values.splunkPlatform.fieldNameConvention.keepOtelContention }}
+    - key: k8s.container.name
+      action: delete
+    - key: k8s.cluster.name
+      action: delete
+    - key: k8s.container.id
+      action: delete
+    - key: k8s.pod.name
+      action: delete
+    - key: k8s.pod.uid
+      action: delete
+    - key: k8s.namespace.name
+      action: delete
+    {{- range $_, $label := .Values.extraAttributes.podLabels }}
+    - key: {{ printf "k8s.pod.labels.%s" $label }}
+      action: delete
+    {{- end }}
+    {{- end }}
 {{- end }}
 
 {{/*
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -303,53 +303,6 @@ processors:
      - k8s.pod.name
      - k8s.pod.uid
   {{- end }}
-  {{- if .Values.splunkPlatform.fieldNameConvention.renameFieldsSck }}
-  resource/sckcompatible:
-    attributes:
-    - key: container_name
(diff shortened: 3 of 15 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -277,11 +277,9 @@ processors:
           tag_name: splunk.com/exclude
           from: pod
         - key: splunk.com/index
-          tag_name: com.splunk.index
-          from: namespace
-        - key: splunk.com/index
-          tag_name: com.splunk.index
           from: pod
+        - key: splunk.com/index
+          from: namespace
```
dmitryax, line 282: Why we need `k8s.namespace.labels.splunk.com/index` instead of setting `com.splunk.index` directly?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -302,6 +300,76 @@ processors:
      - k8s.namespace.name
      - k8s.pod.name
      - k8s.pod.uid
+  {{- else }}
+  resource/splunk:
+    attributes:
+    - key: host.name
+      from_attribute: k8s.node.name
+      action: upsert
```
dmitryax, line 308: `host.name` is already set by resource detection processor. Why it has to be overridden?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -302,6 +300,76 @@ processors:
...
+    - key: com.splunk.sourcetype
+      from_attribute: k8s.pod.annotations.splunk.com/sourcetype
+      action: upsert
```
dmitryax, line 311: This is already defined in `resource/logs` processor

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -302,6 +300,76 @@ processors:
...
+    - key: com.splunk.index
+      from_attribute: k8s.namespace.annotations.splunk.com/index
+      action: upsert
+    - key: service.name
+      from_attribute: k8s.pod.name
+      action: upsert
```
dmitryax, line 317: Why this is needed?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -302,6 +300,76 @@ processors:
...
+  resource/splunk2:
+    attributes:
+    - key: service.name
+      from_attribute: k8s.pod.labels.app
+      action: upsert
```
dmitryax, line 322: this logic already exist in `resource/logs` processor

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -329,6 +329,12 @@ logsCollection:
     excludePaths: []
     # Boolean for ingesting the agent's own log
     excludeAgentLogs: true
+    # Field name conventions to use. (Only for those who are migrating from Splunk Connect for Kubernetes helm chart)
+    fieldNameConvention:
+      # Boolean for renaming pod metadata fields to match to Splunk Connect for Kubernetes helm chart.
+      renameFieldsSck: false
+      # Boolean for keeping Otel convention fields after renaming it
+      keepOtelContention: false
```
dmitryax, line 337: This is applicable to metrics as well, right? I think it should be under `splunkPlatform` config group?

And do we really need an option to send both attributes? Maybe `nameConvention: <otel|sck>`?

rendered/manifests/agent-only/clusterRole.yaml
```diff
@@ -1,83 +0,0 @@
----
```
dmitryax, line 1: Why did you remove all the rendered files?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
rockb1017: in order to guarantee that pod annotation gets the precedence over namespace annotation.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
rockb1017: ah i see. great. i will remove

helm-charts/splunk-otel-collector/values.yaml
rockb1017: okay I will move to splunkPlatform. 
as for the use case for both, maybe for during migration phase, we don't want to break their already set up Splunk queries and stuff, but also they want to correlate logs with data in O11y.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
rockb1017: I thought this is needed for O11y

rendered/manifests/agent-only/clusterRole.yaml
rockb1017: added them back. `make render` only deleted them, but not create new ones.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: It should be already guaranteed. Once attribute from pod is set [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/processor.go#L123), the same attribute from namespace will be _inserted_ [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/processor.go#L133) which is noop if the same key was set from pod

helm-charts/splunk-otel-collector/values.yaml
dmitryax: Ok, sounds good

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: Everything needed for o11y is already here

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: We just need to flip the order here. Please submit another PR for that fix to keep git history clean

rockb1017: okay great. will revert it.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Add initContainer for log checkpoint migration from Fluentd to Otel agent (#253)
 - Add index routing for Splunk Enterprise/Cloud customers (#256)
+- Field name compatibility for SCK and index routing precedence (#258)
```
dmitryax, line 13: Please move this to _Unreleased_ section

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -302,6 +302,64 @@ processors:
...
+  {{- if (eq (include "splunk-otel-collector.splunkPlatformEnabled" .) "true") }}
+  resource/splunk:
+    attributes:
+    - key: com.splunk.index
+      from_attribute: k8s.namespace.annotations.splunk.com/index
+      action: upsert
+    - key: com.splunk.index
+      from_attribute: k8s.pod.annotations.splunk.com/index
+      action: upsert
+  {{- end }}
```
dmitryax, line 315: This section is not needed, right?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -302,6 +302,64 @@ processors:
      - k8s.namespace.name
      - k8s.pod.name
      - k8s.pod.uid
+  {{- else }}
```
dmitryax, line 305: Do we need this `else` here? fluentd logs have the same otel attribute names in this chart, so I think it make sense to apply the SCK conventions uniformly

helm-charts/splunk-otel-collector/values.yaml
dmitryax: Do you still plan to move it under `splunkPlatform` and apply the same renaming to metrics? Or it's needed for logs only?

dmitryax: There is also a typo: conVention

dmitryax: And please make it `true` by default, otherwise all attributes removed

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
rockb1017: oh yea, you are right...

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -303,6 +303,53 @@ processors:
...
+    - key: k8s.pod.name
+      action: delete
+    - key: k8s.pod.uid
+      action: delete
+    - key: k8s.namespace.name
+      action: delete
+    {{- range $_, $label := .Values.extraAttributes.podLabels }}
+    - key: {{ printf "k8s.pod.labels.%s" $label }}
+      action: delete
+    {{- end }}
```
dmitryax, line 351: Do we need two more processors for this? Since it's applied to logs only, we could add these attribute manipulation rules in existing `resource/logs` helper. It should give better performance and no need to copy this logic across agent and gateway.

Also then you'd need to put `resource/logs` after `resource` processor because `resource` processor adds `k8s.cluster.name` that you want to rename.

rockb1017: can even the `removedups` be in a single processor? it wouldn't mess up things by like deleting the attribute before inserting it?

dmitryax: All the attribute actions are evaluated sequentially in resource processor. So just make sure they are put in the right order

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -101,6 +101,49 @@ resource/logs:
...
     {{- end }}
+    {{- if .Values.splunkPlatform.fieldNameConvention.renameFieldsSck }}
+    - key: container_name
+      from_attribute: k8s.container.name
+      action: upsert
+    - key: cluster_name
+      from_attribute: k8s.cluster.name
+      action: upsert
+    - key: container_id
+      from_attribute: k8s.container.id
```
dmitryax, line 112: This attribute is actually recorded as `container.id`. Please also change it below with `delete` action

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -101,6 +101,49 @@ resource/logs:
...
+    - key: namespace
+      from_attribute: k8s.namespace.name
+      action: upsert
+    {{- range $_, $label := .Values.extraAttributes.podLabels }}
+    - key: {{ printf "label_%s" $label }}
+      from_attribute: {{ printf "k8s.pod.labels.%s" $label }}
+      action: upsert
+    {{- end }}
+    {{- end }}
+    {{- if not .Values.splunkPlatform.fieldNameConvention.keepOtelContention }}
```
dmitryax, line 129:
```suggestion
    {{- if not .Values.splunkPlatform.fieldNameConvention.keepOtelConvention }}
```

dmitryax: Please look at the rendered files to verify desired outcome

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Added
+
+- Field name compatibility for SCK and index routing precedence (#258)
```
dmitryax, line 9:
```suggestion
- Field name compatibility for SCK (#258)
```

dmitryax (commented):
One last correction. Otherwise looks good

rockb1017:
Thank you so much for detailed review!

dmitryax (approved):

PR #260, opened 2021-11-01
Add include logs functionality

chaitanyaphalak:
Set splunk.com/include annotation to true on pod and .Values.logsCollection.containers.useSplunkIncludeAnnotation flag to true to include its logs from ingested to your Splunk platform deployment. 
All other logs will be ignored.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -46,6 +46,17 @@ zipkin:
 {{- end }}
 
 {{/*
+Filter Attributes Function
+*/}}
+{{- define "splunk-otel-collector.filterAttr" -}}
+{{- if .Values.logsCollection.containers.useSplunkIncludeAnnotation -}}
+splunk.com/include
+{{- else -}}
+splunk.com/exclude
+{{- end }}
+{{- end }}
+
+{{/*
 Common config for resourcedetection processor
 */}}
 {{- define "splunk-otel-collector.resourceDetectionProcessor" -}}
@@ -94,7 +105,7 @@ resource/logs:
       action: upsert
     - key: k8s.pod.annotations.splunk.com/sourcetype
       action: delete
-    - key: splunk.com/exclude
+    - key: {{ include "splunk-otel-collector.filterAttr" . }}
       action: delete
     {{- if .Values.autodetect.istio }}
     - key: service.name
@@ -158,9 +169,9 @@ Filter logs processor
 # Drop logs coming from pods and namespaces with splunk.com/exclude annotation.
 filter/logs:
   logs:
-    exclude:
+    {{ .Values.logsCollection.containers.useSplunkIncludeAnnotation | ternary "include" "exclude" }}:
       resource_attributes:
-        - key: splunk.com/exclude
+        - key: {{ include "splunk-otel-collector.filterAttr" . }}
           value: "true"
 {{- end }}
 
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -270,11 +270,11 @@ processors:
       annotations:
         - key: splunk.com/sourcetype
           from: pod
-        - key: splunk.com/exclude
-          tag_name: splunk.com/exclude
+        - key: {{ include "splunk-otel-collector.filterAttr" . }}
+          tag_name: {{ include "splunk-otel-collector.filterAttr" . }}
           from: namespace
-        - key: splunk.com/exclude
-          tag_name: splunk.com/exclude
+        - key: {{ include "splunk-otel-collector.filterAttr" . }}
+          tag_name: {{ include "splunk-otel-collector.filterAttr" . }}
           from: pod
         - key: splunk.com/index
           tag_name: com.splunk.index
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
@@ -53,11 +53,11 @@ processors:
       annotations:
         - key: splunk.com/sourcetype
           from: pod
-        - key: splunk.com/exclude
-          tag_name: splunk.com/exclude
+        - key: {{ include "splunk-otel-collector.filterAttr" . }}
+          tag_name: {{ include "splunk-otel-collector.filterAttr" . }}
           from: namespace
-        - key: splunk.com/exclude
-          tag_name: splunk.com/exclude
+        - key: {{ include "splunk-otel-collector.filterAttr" . }}
+          tag_name: {{ include "splunk-otel-collector.filterAttr" . }}
           from: pod
         - key: splunk.com/index
           tag_name: com.splunk.index
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -348,8 +348,6 @@ logsCollection:
     # By the time of reconstructing a multiline log the following information is available to
     # identify source of the logs: namespace, pod and container names. At least one source
     # identifier has to be specified in for each multiline config.
-    multilineConfigs: []
-
     # The following example shows how to setup multiline log processing for logs having subsequent
     # log lines written with an offset. Let's say a k8s deployment called "buttercup-app" is
     # scheduled to run in "default" namespace with a java container called "server", and the
@@ -372,7 +370,8 @@ logsCollection:
     #     containerName:
     #       value: server
     #     firstEntryRegex: ^[^\s].*
-    # Set this flag to `true` to collect logs from pods with `splunk.com/include: true` annotation and ignore others.
+    multilineConfigs: []
+    # Set useSplunkIncludeAnnotation flag to `true` to collect logs from pods with `splunk.com/include: true` annotation and ignore others.
     # All other logs will be ignored.
     useSplunkIncludeAnnotation: false
 
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -349,6 +349,7 @@ logsCollection:
     # identify source of the logs: namespace, pod and container names. At least one source
     # identifier has to be specified in for each multiline config.
     multilineConfigs: []
+
     # The following example shows how to setup multiline log processing for logs having subsequent
     # log lines written with an offset. Let's say a k8s deployment called "buttercup-app" is
     # scheduled to run in "default" namespace with a java container called "server", and the
@@ -371,6 +372,9 @@ logsCollection:
     #     containerName:
     #       value: server
     #     firstEntryRegex: ^[^\s].*
+    # Set this flag to `true` to collect logs from pods with `splunk.com/include: true` annotation and ignore others.
+    # All other logs will be ignored.
+    useSplunkIncludeAnnotation: false
 
   checkpointPath: "/var/addon/splunk/otel_pos"
 
--- a/helm-charts/splunk-otel-collector/values.schema.json
+++ b/helm-charts/splunk-otel-collector/values.schema.json
@@ -482,6 +482,9 @@
                 "type": "string"
               }
             },
+            "useSplunkIncludeAnnotation": {
+              "type": "boolean"
+            },
             "multilineConfigs": {
               "type": "array",
               "items": {


dmitryax (commented):
Overall looks good. 
Please delete the attribute once it's used in the `resource/logs` processor.
2. Please add the same in the otelCollector and disable it in the otelAgent in the `otelCollector.enabled=true`. The reason is that there is no k8s metadata on the agent if the gateway is enabled.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -332,7 +332,6 @@ logsCollection:
     # Extra operators for container logs.
     # https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/docs/operators/README.md#what-operators-are-available
     extraOperators: []
-
```
dmitryax, line 335: nit: unexpected removal? :)

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -103,6 +105,21 @@ resource/logs:
     {{- end }}
 {{- end }}
 
+{{/*
+Include logs processor
+*/}}
+{{- define "splunk-otel-collector.includeLogsProcessors" -}}
+# If .Values.logsCollection.useSplunkIncludeAnnotation is set to true, only logs for pods with splunk.com/include annotation are ingested
```
dmitryax, line 112: `.Values.logsCollection.useSplunkIncludeAnnotation` is wrong I believe. But do we need to mention the condition in the unconditional helper? Maybe just remove this line? Seems like there are enough other comments added already

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -365,6 +365,9 @@ logsCollection:
     #       value: server
     #     firstEntryRegex: ^[^\s].*
 
+    # Set `splunk.com/include` annotation to true on pod and `logsCollection.useSplunkIncludeAnnotation` flag to `true` to include its logs from ingested to your Splunk platform deployment.
+    # All other logs will be ignored. You cant use this feature with the above mentioned exclude feature. You can only use either the include feature or the exclude feature.
```
dmitryax, line 369: Oh missed one last thing. There is nothing "above mentioned". Please rephrase :)

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
```diff
@@ -221,6 +230,9 @@ service:
         - batch
         - filter/logs
         - resource/logs
+        {{- if .Values.logsCollection.containers.useSplunkIncludeAnnotation }}
+        - filter/include_pod_logs
+        {{- end }}
```
dmitryax, line 235: This filter should go before `resource/logs` to apply the include attribute removal. same in `otelCollector`

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -434,6 +443,9 @@ service:
         {{- end }}
         - k8sattributes
         - batch
+        {{- if .Values.logsCollection.containers.useSplunkIncludeAnnotation }}
+        - filter/include_pod_logs
+        {{- end }}
```
dmitryax, line 448: This should go under the next condition `{{- if not .Values.otelCollector.enabled }}`. Otherwise, with `otelCollector=enable`, it won't do anything because there is no k8s metatada available at this point to filter the logs on.

Another question. Do we still need to handle `splunk.com/exclude` if `logsCollection.containers.useSplunkIncludeAnnotation=true`? 

If not, we can avoid setting up unnecessary processors and reduce amount of code by just changing the helper for `filter/logs`:

```
filter/logs:
  logs:
    {{ .Values.logsCollection.containers.useSplunkIncludeAnnotation | "include" "exclude" }}:
      match_type: strict
      resource_attributes:
        - key: splunk.com/{{ .Values.logsCollection.containers.useSplunkIncludeAnnotation | "include" "exclude" }}
          value: "true"
```

What do you think?

chaitanyaphalak: really good idea, i think we should just use your definition of filter/logs

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -282,6 +282,11 @@ processors:
         - key: splunk.com/index
           tag_name: com.splunk.index
           from: pod
+        {{- if .Values.logsCollection.containers.useSplunkIncludeAnnotation }}
+        - key: splunk.com/include
+          tag_name: splunk.com/include
+          from: pod
+        {{- end }}
```
dmitryax, line 289: Can we take `splunk.com/exclude` annotation under `else` clause to avoid unnecessary attributes manipulations?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -365,6 +365,9 @@ logsCollection:
     #       value: server
     #     firstEntryRegex: ^[^\s].*
 
+    # Set `splunk.com/include` annotation to true on pod and `logsCollection.useSplunkIncludeAnnotation` flag to `true` to include its logs from ingested to your Splunk platform deployment.
```
dmitryax, line 368:
```suggestion
    # Set `splunk.com/include` annotation to true on pod and `logsCollection.containers.useSplunkIncludeAnnotation` flag to `true` to include its logs from ingested to your Splunk platform deployment.
```

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: Btw why don't we take `splunk.com/include` from namespace? Just curious

chaitanyaphalak: The feature request from feedback was only for pods. We could add namespace support in future if there is customer demand!

dmitryax: Do you see any issue with having consistent behavior for the filtering from the start?

If you agree, we could even add a helper for the filtering attribute that would return `splunk.com/include` or `splunk.com/exclude` depending on the configuration. Then all the places where `splunk.com/exclude` is used can be just replaced with `{{ include "splunk-otel-collector.filterAttr" . }}`. It'll remove unnecessary attribute manipulations which is always good for performance

chaitanyaphalak: I think we should probably do that later during the refactor.

dmitryax: This doesn't seem to be a big change, not sure why it has to introduce any tech debt

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -153,12 +155,28 @@ Filter logs processor
...
 {{- end }}
 
+{{/*
+Filter Attributes Function
+*/}}
+{{- define "splunk-otel-collector.filterAttr" -}}
+{{- if .Values.logsCollection.containers.useSplunkIncludeAnnotation }}
+- key: splunk.com/include
+  tag_name: splunk.com/include
+  from: pod
```
dmitryax, line 172: My idea was to return only `splunk.com/include` or `splunk.com/exclude` in this helper and update _all_ the places where `splunk.com/exclude` currently is being used:
- resource/logs delete action
- k8sattributes processor
- filter/logs processor

For example
```yaml
resource/logs:
  attributes:
    - key: {{ include "splunk-otel-collector.filterAttr" . }}
       action: delete
```

WDYT? Can we make it this way? And here I'm assuming we apply it consistently to pods and namespaces. I still don't understand why we have to introduce this inconsistency just because a customer wants it to be scraped from pods for now. I believe having it as a consistent config for both pods and namespaces is what can be expected by others and will not introduce any problems to the customer who asked for this feature initially.

chaitanyaphalak:
Hey Dmitrii, updated the PR, with your latest recommendations, lint-test is failing due to some fluentd error, I am not sure why?

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
```diff
@@ -56,15 +56,13 @@ processors:
...
-        - key: splunk.com/exclude
-          tag_name: splunk.com/exclude
-          from: pod
         - key: splunk.com/index
           tag_name: com.splunk.index
           from: namespace
         - key: splunk.com/index
           tag_name: com.splunk.index
           from: pod
+        {{- include "splunk-otel-collector.filterAttr" . | nindent 8 }}
```
dmitryax, line 65: Please update usage here

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -270,18 +270,19 @@ processors:
...
+        - key: {{ include "splunk-otel-collector.filterAttr" . }}
+          tag_name: {{ include "splunk-otel-collector.filterAttr" . }}
           from: pod
         - key: splunk.com/index
           tag_name: com.splunk.index
           from: namespace
         - key: splunk.com/index
           tag_name: com.splunk.index
           from: pod
+        {{- include "splunk-otel-collector.filterAttr" . | nindent 8 }}
```
dmitryax, line 285: this is not needed

rendered/manifests/agent-only/configmap-otel-agent.yaml
```diff
@@ -16,236 +16,4 @@ metadata:
...
-          - k8sattributes
-          - batch
-          - resource
-          - resourcedetection
-          receivers:
-          - otlp
-          - jaeger
-          - smartagent/signalfx-forwarder
-          - zipkin
+    Error: 'error converting YAML to JSON: yaml: line 123: could not find expected '':'''
```
dmitryax, line 19: This doesn't seem to be desirabe

dmitryax (approved):
LGTM. did you run `make render`?

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -45,6 +45,17 @@ zipkin:
 {{- end }}
 {{- end }}
 
+{{/*
+Filter Attributes Function
+*/}}
+{{- define "splunk-otel-collector.filterAttr" -}}
+{{- if .Values.logsCollection.containers.useSplunkIncludeAnnotation }}
```
dmitryax, line 52: I think the problem is because this helper returns `\nsplunk.com/include` or `\nsplunk.com/exclude`. Please trim the space after `if` and `else`

dmitryax:
```suggestion
{{- if .Values.logsCollection.containers.useSplunkIncludeAnnotation -}}
```

chaitanyaphalak: Let me try, thanks for the input!

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -371,7 +371,9 @@ logsCollection:
     #     containerName:
     #       value: server
     #     firstEntryRegex: ^[^\s].*
-
+    # Set `splunk.com/include` annotation to true on pod and `logsCollection.containers.useSplunkIncludeAnnotation` flag to `true` to include its logs from ingested to your Splunk platform deployment.
+    # All other logs will be ignored.
+    useSplunkIncludeAnnotation: false
```
dmitryax, line 376: Looks good to me. Just please add a line breaks above and below this section to avoid mixing with other comments

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -371,7 +371,9 @@ logsCollection:
     #     containerName:
     #       value: server
     #     firstEntryRegex: ^[^\s].*
-
+    # Set `splunk.com/include` annotation to true on pod and `logsCollection.containers.useSplunkIncludeAnnotation` flag to `true` to include its logs from ingested to your Splunk platform deployment.
```
dmitryax, line 374: It's not only about Splunk Platform. How about this wording?
```suggestion
    # Set this flag to `true` to collect logs from pods with `splunk.com/include: true` annotation and ignore others.
```

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -349,6 +349,7 @@ logsCollection:
     # identify source of the logs: namespace, pod and container names. At least one source
     # identifier has to be specified in for each multiline config.
     multilineConfigs: []
+
```
dmitryax, line 352: This is a wrong place to separate the comments :)

Comments below are still applied to multiline config and mixed with comments for `useSplunkIncludeAnnotation`

dmitryax (approved):
LGTM

PR #263, opened 2021-11-02
add initContainer for running as non root user

rockb1017:


README.md
```diff
@@ -265,6 +265,11 @@ There are following known limitations of native OTel logs collection:
   Logs collection with fluentd is still recommended if chart deployed with `autodetect.istio=true`.
 - Journald logs cannot be collected natively by Splunk OTel Collector yet.
 
+### Running it as non root user
```
tmorrisonsfx, line 268: Suggested change:

"Running the container in non-root user mode"

"Collecting logs often requires reading log files that are owned by the `root` user. By default, the container runs with `securityContext.runAsUser = 0`, which gives the `root` user permission to read those files. To run the container in `non-root` user mode, set `.Values.otelAgent.runInitForNonRoot` to `true` to cause the container to run the required file system operations as UID and GID `20000`.

**Note:** `cri-o` container runtime did not work during internal testing."

rockb1017:
@tmorrisonsfx Could you re-review? thanks!

tmorrisonsfx (approved):

tmorrisonsfx:
> @tmorrisonsfx Could you re-review? thanks!

Done! Thanks.

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -124,6 +124,45 @@ spec:
...
+          image: registry.access.redhat.com/ubi8/ubi
+          command: ['sh', '-c', '
+          mkdir -p {{ .Values.logsCollection.checkpointPath }};
+          chown -Rv {{ .Values.otelAgent.securityContext.runAsUser | default 20000 }}:{{ .Values.otelAgent.securityContext.runAsGroup | default 20000 }} {{ .Values.logsCollection.checkpointPath }};
+          chmod -v g+rwxs {{ .Values.logsCollection.checkpointPath }};
+          {{ if .Values.logsCollection.containers.enabled -}}
+          if [ -d "/var/lib/docker/containers" ];
+          then
+              chgrp -Rv {{ .Values.otelAgent.securityContext.runAsGroup | default 20000 }} /var/lib/docker/containers;
+              chmod -R g+rwxs /var/lib/docker/containers;
```
dmitryax, line 138: Why do we need to grant write permissions here? We should never write to these directories, only to checkpoints directory

docs/advanced-configuration.md
```diff
@@ -103,6 +103,12 @@ There are following known limitations of native OTel logs collection:
   Logs collection with fluentd is still recommended if chart deployed with `autodetect.istio=true`.
 - Journald logs cannot be collected natively by Splunk OTel Collector yet.
 
+## Running the container in non-root user mode
+
+Collecting logs often requires reading log files that are owned by the root user. By default, the container runs with `securityContext.runAsUser = 0` which gives the `root` user permission to read those files. To run the container in `non-root` user mode, set `.Values.otelAgent.runInitForNonRoot` to `true`to cause the container to run the required file system operations as UID and GID `20000`.
+
+Note: `cri-o` container runtime did not work during internal testing.
```
dmitryax, line 110: We cannot switch to this approach by default it it breaks the functionality for cri-o runtime. Please investigate further

helm-charts/splunk-otel-collector/ci/logs-only-values.yaml
```diff
@@ -3,5 +3,5 @@ splunkObservability:
   realm: fake-realm
   accessToken: fake-token
 
-tracesEnabled: false
-metricsEnabled: false
+  tracesEnabled: false
+  metricsEnabled: false
```
dmitryax, line 7: Please don't include unrelated changes, better to submit separate PRs even for small fixes

helm-charts/splunk-otel-collector/ci/logs-only-otel-se-o11y-values.yaml
```diff
@@ -0,0 +1,20 @@
...
+  # Required for Splunk Enterprise/Cloud (if `endpoint` is specified). Splunk
+  # HTTP Event Collector token.
+  token: "fake-token"
+
+  index: "main"
+
+  logsEnabled: true
+  metricsEnabled: false
+
+logsEngine: otel
```
dmitryax, line 20: Please create another PRs for this one

helm-charts/splunk-otel-collector/templates/daemonset.yaml
dmitryax: And how will it work for new log files? This init container recursively patches existing log files only once, but new log files (from new pods or rotated files) won't be patched and collector won't be able to read them. Or I'm missing something?

May be some linux capabilities can give us a way to read these files instead of patching them...

rockb1017: right. we don't need write permission. will update it.

docs/advanced-configuration.md
rockb1017: so this is not the default in this PR.

helm-charts/splunk-otel-collector/templates/daemonset.yaml
rockb1017: `setfacl` command takes care of new files added to this directory.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -240,7 +240,14 @@ otelAgent:
       # This value is being used as a source for default memory_limiter processor configurations
       memory: 500Mi
 
-  securityContext: {}
+  # Boolean for running initContainer that will change group ownership of directories for container logs and journald logs.
+  # It is required for running the agent as a non root user. By default, it will run as UID and GID 20000.
+  # Disable it if you would rather run it as root user.
+  runInitForNonRoot: false
```
dmitryax, line 246: Do we need this option?

Maybe we can rely on `securityContext.runAsUser` to be anything other than `0` instead? 

Then we can have
```
#  securityContext:
#    runAsUser: 20000
#    runAsGroup: 20000
```
as a recommended commented out config.

dmitryax: Otherwise any custom setting of `securityContext` that user might have will be ignored with this change by default

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -258,7 +258,12 @@ agent:
       # This value is being used as a source for default memory_limiter processor configurations
       memory: 500Mi
 
-  securityContext: {}
+  # To collect container logs and journald logs, it will run the agent as a root user.
+  # To run it as non root user, uncomment below `securityContext` options.
+
+  # securityContext:
```
dmitryax, line 264: I curios if schema validation will fail if we remove `securityContext: {}` from here. I believe it should be failing because the type is required to be `object`, but now it's changed to `nil`

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -258,7 +258,12 @@ agent:
       # This value is being used as a source for default memory_limiter processor configurations
       memory: 500Mi
 
-  securityContext: {}
+  # To collect container logs and journald logs, it will run the agent as a root user.
+  # To run it as non root user, uncomment below `securityContext` options.
```
dmitryax, line 262: Let's add the a note about the init container. That would likely be the reason why users won't change the default.
Something like:

Setting `runAsUser` and `runAsGroup` to a non root user enables an init container that patches group permissions of container logs directories on the host filesystem to make logs readable by this non root user.

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -138,6 +138,47 @@ spec:
               mountPath: /var/log
             - name: varlibdockercontainers
               mountPath: /var/lib/docker/containers
+        {{- if $agent.securityContext }}
+        {{- if not (eq (toString $agent.securityContext.runAsUser) "0" ) }}
```
dmitryax, line 142: I think we should enable it if both runAsUser and runAsGroup are provided
```suggestion
        {{- if and $agent.securityContext.runAsUser $agent.securityContext.runAsGroup }}
```

To make it work with one if statement I'd recommend bringing back `securityContext: {}` by default instead of removing it

helm-charts/splunk-otel-collector/values.yaml
rockb1017: schema validation doesn't fail with `securityContext: {}`. I think because `{}` is an object. `securityContext:` this fails tho.

helm-charts/splunk-otel-collector/templates/daemonset.yaml
rockb1017: applied

dmitryax (approved):
LGTM

PR #270, opened 2021-11-09
documentation for using proxy

rockb1017:


README.md
```diff
@@ -289,6 +289,10 @@ autodetect:
   prometheus: true
 ```
 
+## Proxy Support
+
+Otel collector in this chart support [proxy environment variables](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter#proxy-support). You can set these proxy environment variables by setting `.Values.otelAgent.extraEnvs` for agent daemonset or `.Values.otelCollector.extraEnvs` for collector deployment.
```
dmitryax, line 294: We have a config example for using proxy https://github.com/signalfx/splunk-otel-collector-chart/tree/main/examples#route-telemetry-data-through-a-proxy-server . Please add a link to it

dmitryax: Also I think this should go in `## Advanced configuration` section

rockb1017: oh i didn't see that example part. i think that is good enough. We shouldn't need another doc at README. what do you think

dmitryax: Sounds good to me. And we already have links to examples in the README so we probably can close this PR

PR #276, opened 2021-11-11
updated sourcetype precedence order

luckyj5:
updated sourcetype precedence order: 

- Sourcetype defined via Pod annotation [1]
- Sourcetype defined via optional config param in helm chart values [2]
- K8s metadata: “kube:container:<CONTAINER_NAME>” for container logs [3]
- Sourcetype for non-container logs/ extrafiles: "defined by the user in the config" [4]


dmitryax:
LGTM. We probably should remove `sourcetype` field from `splunk_hec/platform_logs` and `splunk_hec/platform_metrics` exporters since they don't have any effect anymore.

luckyj5:
> LGTM. We probably should remove `sourcetype` field from `splunk_hec/platform_logs` and `splunk_hec/platform_metrics` exporters since they don't have any effect anymore.

Thanks @dmitryax. 
PR updated to remove not used sourcetype fields.

dmitryax (approved):
LGTM. @rockb1017 can you please take a look as well?

rockb1017 (approved):
Looks good!

PR #282, opened 2021-11-14
ingest metrics to Splunk only when metrics index is given

rockb1017:
when metrics are ingested to event type splunk index, client doesn't error out but it is not right. metrics should be sent to 
Splunk's metric type index. 
this prevents metrics data being sent when `metrics_index` is not specified.

### DIFF
--- a/helm-charts/splunk-otel-collector/values.schema.json
+++ b/helm-charts/splunk-otel-collector/values.schema.json
@@ -1084,30 +1084,32 @@
     {
       "if": {
         "$comment": "if splunkPlatform.metricsEnabled is true and splunkPlatform.endpoint is provided",
+        "properties": {
+          "splunkPlatform": {
             "properties": {
-              "splunkPlatform": {
-                "properties": {
-                  "endpoint": {
-                    "minLength": 1
-                  }
-                }
+              "endpoint": {
+                "minLength": 1
               }
-            },
-              "splunkPlatform": {
-                "properties": {
-                  "metricsEnabled": {
-                    "const": true
-                  }
-                }
-              }
-            },
+            }
+          }
+        }
+      },
       "then": {
         "$comment": "then splunkPlatform.metrics_index must be provided",
         "properties": {
           "splunkPlatform": {
-            "properties": {
-              "metrics_index": {
-                "minLength": 1
+            "if" : {
+              "properties": {
+                "metricsEnabled": {
+                  "const": true
+                }
+              }
+            },
+            "then": {
+              "properties": {
+                "metrics_index": {
+                  "minLength": 1
+                }
               }
             }
           }
--- a/helm-charts/splunk-otel-collector/values.schema.json
+++ b/helm-charts/splunk-otel-collector/values.schema.json
@@ -1115,6 +1115,41 @@
           }
         }
       }
+    },
+    {
+      "if": {
+        "$comment": "if splunkPlatform.logsEnabled is true and splunkPlatform.endpoint is provided",
+        "properties": {
+          "splunkPlatform": {
+            "properties": {
+              "endpoint": {
+                "minLength": 1
+              }
+            }
+          }
+        }
+      },
+      "then": {
+        "$comment": "then splunkPlatform.index must be provided",
+        "properties": {
+          "splunkPlatform": {
+            "if" : {
+              "properties": {
+                "logsEnabled": {
+                  "const": true
+                }
+              }
+            },
+            "then": {
+              "properties": {
+                "index": {
+                  "minLength": 1
+                }
+              }
+            }
+          }
+        }
+      }
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -90,7 +90,7 @@ Whether logs enabled for Splunk Platform.
 Whether metrics enabled for Splunk Platform.
 */}}
 {{- define "splunk-otel-collector.platformMetricsEnabled" -}}
-{{- and (eq (include "splunk-otel-collector.splunkPlatformEnabled" .) "true") .Values.splunkPlatform.metricsEnabled }}
+{{- and (eq (include "splunk-otel-collector.splunkPlatformEnabled" .) "true") .Values.splunkPlatform.metricsEnabled (not (eq .Values.splunkPlatform.metrics_index "")) }}
```
dmitryax, line 93: Why metrics only? Shouldn't it be enforced for logs as well? I see `splunkPlatform.index` is optional. 

I believe this validation should better go into values.schema.json. Otherwise it can be confusing for a user when if they set `splunkPlatform.metricsEnabled=true` and it's silently ignored

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -36,7 +36,7 @@ Create chart name and version as used by the chart label.
 Whether to send data to Splunk Platform endpoint
 */}}
 {{- define "splunk-otel-collector.splunkPlatformEnabled" -}}
-{{- not (eq .Values.splunkPlatform.endpoint "") }}
+{{- and (not (eq .Values.splunkPlatform.endpoint "")) (not (eq .Values.splunkPlatform.token "")) }}
```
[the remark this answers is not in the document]
rockb1017: oh i see. awesome. Thanks!

helm-charts/splunk-otel-collector/templates/_helpers.tpl
rockb1017: when index is not provided, there is still a default index that HEC token is associated with. That is typically event type index but that can be a metric type index too. so it makes sense to have a same behavior. okay i will implement this in `values.schema.json`

rockb1017:
with blank metrics_index, i get this error. 
```
$ helm upgrade olly -f .tmp/values.yaml .
splunk-otel-collector:
- (root): Must validate "then" as "if" was v
- splunkPlatform.metrics_index: String length must be greater than or equal to 1
```

rockb1017:
so the validation is working correctly.

dmitryax:
Should we also add a comment to `metrics_index` saying that it's required when `metricsEnabled=true`?

I still don't get why it's only applicable to metrics. Does HEC exporter set's some default value for logs, but not for metrics?

rockb1017:
add comment and applied same validation to `index`

dmitryax (approved):

dmitryax:
It looks good to me. 

I'm thinking would it be easier for users if we have default value for indexes in values.yaml? I see SCK uses `main` for both logs and metrics. Shouldn't we do similar here? 

I'm ok with any approach, just trying to think of it from end user perspective. Right now we call out only `endpoint` and `token` as required fields, but if user sets them up they'll get this validation error.

rockb1017:
yea, it would work sending all to main index, but metric data shouldn't really go to main index (event type index) 
so even for SCK, we print this warning.
```
##################################################################
####   WARN: You did not set splunk.hec.indexName correctly   ####
##################################################################

Metrics collected by splunk-kubernetes-metrics are supposed to be indexed in
a Metrics Index. But you are using the same index for metrics as for logs
and/or objects. You should use splunk-kubernetes-metrics.splunk.hec.indexName
to set index for metrics.

Follow
http://docs.splunk.com/Documentation/Splunk/latest/Indexer/Setupmultipleindexes#Create_metrics_indexes
to create a Metrics Index in your Splunk.

Then run:

    helm upgrade sck -f values.yaml splunk/splunk-connect-for-kubernetes
```

dmitryax:
Hm.. Ok does setting up metrics index require some manual steps on Splunk side while `main` log index is always available?

I'm thinking maybe we should make `metricsEnabled: false` by default and set default `index: main` for logs?

dmitryax:
I'm merging this PR, but let's try to do something with default values before the release to make it work with `endpoint` and `token` as only required fields

rockb1017:
right. main is always there and it requires users to manually create metric type index. I like your suggestion for default. will create PR

dmitryax:
Thanks @rockb1017, could you please also submit a PR to change this line in the specification https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md?plain=1#L140 with justification that it requires metrics index to be created on Splunk side, so we should keep it disabled by default.

PR #285, opened 2021-11-15
Extract `container.image.tag` attribute from `container.image.name`

dmitryax:
Update k8sClusterReceiver component to export `container.image.tag` attribute from `container.image.name` according to OpenTelemetry semantic conventions https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/resource/semantic_conventions/container.md

This can be removed once this issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/6314 is resolved

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
@@ -94,6 +94,21 @@ processors:
         key: {{ .name }}
         value: {{ .value }}
       {{- end }}
+      # Extract "container.image.tag" attribute from "container.image.name" here until k8scluster
+      # receiver does it natively.
+      - key: container.image.name
+        pattern: ^(?P<container_image_name>[^\:]+)(?:\:(?P<container_image_tag>.*))?
+        action: extract
+      - key: container.image.name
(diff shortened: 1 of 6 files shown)


mstumpfx (approved):

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -94,6 +94,21 @@ processors:
         key: {{ .name }}
         value: {{ .value }}
       {{- end }}
+      # Extract "container.image.tag" attribute from "container.image.name" here until k8scluster
+      # receiver does it natively.
+      - key: container.image.name
+        pattern: ^(?P<container_image_name>[^\:]+)(?:\:(?P<container_image_tag>.*))?
```
rmfitzpatrick, line 100: Should you be using a temp prefix in the unlikely event these are stepping on custom attributes?

dmitryax: You're right. Thanks! Let me update it

dmitryax: done

PR #290, opened 2021-11-16
Add journald config

luckyj5:
Added journald configuration under otel logs collection.


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -235,7 +235,35 @@ receivers:
               from: log
               to: $$
   {{- end }}
-
+  # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/journaldreceiver
+  {{- if .Values.logsCollection.journaldLogs.enabled }}
```
rockb1017, line 239: let's make it into `    {{- if and (eq .Values.logsEngine "otel") .Values.logsCollection.journaldLogs.enabled }}`

rockb1017: actually we also need this `{{- if (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}`

rockb1017 (changes requested):

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
luckyj5: Thanks @rockb1017.

rockb1017 (changes requested):

rendered/manifests/logs-only/configmap-otel-agent.yaml
```diff
@@ -132,39 +132,9 @@ data:
     receivers:
       fluentforward:
         endpoint: 0.0.0.0:8006
-      journald/containerd:
+      journald:
```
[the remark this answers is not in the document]
luckyj5: Thanks @rockb1017 
Updated

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -235,7 +235,37 @@ receivers:
...
-
+  # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/journaldreceiver
+  {{- if (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
+  {{- if and (eq .Values.logsEngine "otel") .Values.logsCollection.journaldLogs.enabled }}
+  {{- if .Values.logsCollection.journaldLogs.units }}
+  {{- range $_, $unit := .Values.logsCollection.journaldLogs.units }}
+  {{- printf "journald/%s:" $unit.name | nindent 2 }}
+    directory: {{ $.Values.logsCollection.journaldLogs.directory }}
+    units: [{{ $unit.name }}]
+    priority: {{ $unit.priority | default $.Values.logsCollection.journaldLogs.defaultPriority }}
```
dmitryax, line 246: Is this field the only reason to have receiver per unit? If you think it's critical to give users an option to customize priority per unit, I recommend grouping them by priority and create receiver per priority, e.g. `journald/info`, `journald/error` etc. I believe usually users don't need different levels for each unit, so it'll be probably just one receiver in most cases --> better for performance. Look at https://masterminds.github.io/sprig/dicts.html and http://masterminds.github.io/sprig/lists.html on how to create the groups from your existing config interface

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -235,7 +235,37 @@ receivers:
...
+      host.name: 'EXPR(env("K8S_NODE_NAME"))'
+      {{- if $.Values.clusterName }}
+      k8s.cluster.name: {{ $.Values.clusterName }}
+      {{- end }}
+      {{- if $.Values.environment }}
+      deployment.environment: {{ $.Values.environment }}
+      {{- end }}
+      {{- if $.Values.extraAttributes.custom }}
+      {{- toYaml $.Values.extraAttributes.custom | nindent 6 }}
+      {{- end }}
```
dmitryax, line 260: You can use existing `resource` processor, after batching it'll be more performant.

Btw you cannot use `Values.extraAttributes.custom` as is here. it has different format

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -532,4 +562,31 @@ service:
         {{- end }}
         {{- end }}
     {{- end }}
+
+    {{- if (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
+    {{- if and (eq .Values.logsEngine "otel") .Values.logsCollection.journaldLogs.enabled }}
+    logs/journald:
```
dmitryax, line 568: I recommend reusing existing `logs/extraFiles` pipeline instead of creating a new one. Just add journald receiver there if enabled. We can rename that pipeline to `logs/host`

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -515,6 +515,12 @@
             }
           }
         },
+        "journaldLogs": {
+          "type": "object",
+          "items": {
+            "type": "object"
```
dmitryax, line 521: Please define schema for `journaldLogs` object here

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -392,6 +392,22 @@ logsCollection:
     # All other logs will be ignored.
     useSplunkIncludeAnnotation: false
 
+# Configuration for collecting journald logs using otel collector
+  journaldLogs:
```
dmitryax, line 396:
```suggestion
  journald:
```

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
rockb1017: it is needed for `cursor` (checkpoint data) If user is migration from SCK, which stores cursor information per service, and in this format, it will be handled seamlessly. 

so this is how journald cursor works. 
Let's say SCK has read these journald logs. 

message1 | kubelet | cursor1
message2 | docker | cursor2
message3 | ssh | cursor3
message4 | kubelet | cursor4
message5 | docker  | cursor5

After checkpoint migration, we have `cursor3` for ssh, `cursor4` for kubelet, and `cursor5` for docker. 

If i use ssh's cursor info and start reading journal logs, I will get message 4 & 5, and so on. so message 4& 5 would be duplicate. 

the only way to not have duplicate is to give cursor for each service. 
but even if we just choose arbitrary cursor from what we have (among cursor 3~5), it could have some data loss or duplicate, but it won't be a lot. I think it should be acceptable for the performance. `kubelet` service is constantly logging in k8s cluster, so I think it is safe to just take kubelet's cursor. 
Should I change the checkpoint migration script for journald?

dmitryax: @rockb1017  I would leave it up to you. Having less receivers is definitely better, but if you think that it'll be hard to ensure checkpoint migration, we can leave it as is and maybe change later

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
luckyj5: Thanks @dmitryax. I'll update the PR to use existing resource processor
Will still need to keep `host.name` as it's not a part of resource processor.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
luckyj5: Sure, will update it. 
BTW - `logs/extraFiles` pipeline didn't exist at the time of the PR I think :)

luckyj5:
Hey @dmitryax @rockb1017, can you guys take a look when get a chance? Let me know if anything. Thanks!

rendered/manifests/agent-only/configmap-agent.yaml
```diff
@@ -188,44 +188,4 @@ data:
...
-          - k8sattributes
-          - batch
-          - resource
-          - resourcedetection
-          receivers:
-          - otlp
-          - jaeger
-          - smartagent/signalfx-forwarder
-          - zipkin
+      pipelines: null
```
dmitryax, line 191: Seems like templating is broken. This is invalid output

luckyj5: Thanks. Updated the PR

dmitryax (commented):
Can you add a changelog item and a section in advanced configuration

luckyj5:
> Can you add a changelog item and a section in advanced configuration

Added. Let me know if any modifications

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -242,12 +242,32 @@ receivers:
...
+    resource:
+      com.splunk.source: {{ $.Values.logsCollection.journald.directory }}
+      com.splunk.sourcetype: 'EXPR("kube:"+$$._SYSTEMD_UNIT)'
+      com.splunk.index: {{ $.Values.logsCollection.journald.index | default $.Values.splunkPlatform.index}}
+      host.name: 'EXPR(env("K8S_NODE_NAME"))'
+  {{- end }}
+  {{- else }}
+  journald:
+    directory: {{- toYaml .Values.logsCollection.journald.directory | nindent 6 }}
+    priority: {{ .Values.logsCollection.journald.defaultPriority }}
```
dmitryax, line 266: How to get into this condition? By setting `logsCollection.journald.units` to `null`? Will this collect logs from all available systemd services in that case?

May be this can be default and we provide a way to configure `units` as an option?

Please document it

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -242,12 +242,32 @@ receivers:
           - move:
               from: log
               to: $$
-
```
dmitryax, line 245: unintended change I believe?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -242,12 +242,32 @@ receivers:
...
+  {{- end }}
+  {{- else }}
+  journald:
+    directory: {{- toYaml .Values.logsCollection.journald.directory | nindent 6 }}
+    priority: {{ .Values.logsCollection.journald.defaultPriority }}
+  {{- end }}
+  {{- end }}
+  {{- end }}
   {{- end }}
-
```
dmitryax, line 250: please revert this line removal

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -489,9 +519,11 @@ service:
         {{- if eq (include "splunk-otel-collector.o11yLogsEnabled" .) "true" }}
         - splunk_hec/o11y
         {{- end }}
+        {{- end }}
     {{- end }}
     {{- end }}
 
+
```
dmitryax, line 526: please revert

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -401,6 +401,22 @@ logsCollection:
...
+    defaultPriority: info
+    # List of service units to collect and configuration for each. To collect all journald services logs, leave units blank.
+    units:
+      - name: kubelet
+        priority: info
+      - name: docker
+        priority: info
+      - name: containerd
+        priority: info
+    # Route journald logs to its own Splunk Index by specifying the index value below, else leave it blank. Please make sure the index exist in Splunk and is configured to receive HEC traffic.
```
dmitryax, line 417:
```suggestion
    # Route journald logs to its own Splunk Index by specifying the index value below, else leave it blank. Please make sure the index exist in Splunk and is configured to receive HEC traffic. Not applicable to Splunk Observability.
```

rendered/manifests/logs-only/configmap-agent.yaml
```diff
@@ -168,6 +168,8 @@ data:
           receivers:
           - fluentforward
           - otlp
+        logs/host:
+          receivers: null
```
dmitryax, line 172: Please always look at rendered files, this clearly means that something is broken

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -469,12 +489,22 @@ service:
         {{- end }}
         {{- end }}
 
-    {{- if .Values.logsCollection.extraFileLogs }}
-    logs/extraFiles:
+    {{- if or (.Values.logsCollection.extraFileLogs) (.Values.logsCollection.journald) }}
```
dmitryax, line 492:
```suggestion
    {{- if or (.Values.logsCollection.extraFileLogs) (.Values.logsCollection.journald.enabled) }}
```

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -410,6 +415,11 @@ spec:
         hostPath:
           path: {{ .Values.logsCollection.checkpointPath }}
           type: DirectoryOrCreate
+      {{- if .Values.logsCollection.journald.enabled}}
+      - name: journaldlogs-otel
```
dmitryax, line 419:
```suggestion
      - name: journaldlogs
```

docs/advanced-configuration.md
```diff
@@ -195,6 +195,30 @@ logsCollection:
...
+    directory: /run/log/journal
+    defaultPriority: info
+    # List of service units to collect and configuration for each. To collect all journald services logs, leave units blank.
+    units:
+      - name: kubelet
+        priority: info
+      - name: docker
+        priority: info
+      - name: containerd
+        priority: info
```
dmitryax, line 217: Why not enable all systemd sources by default, and provide `units` configuration as a commented out option? It would be nice to have only one journald receiver by default, not 3.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: Nvm. Just found out all the answers, and it is documented

docs/advanced-configuration.md
luckyj5: updated

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -469,12 +492,23 @@ service:
         {{- end }}
         {{- end }}
 
-    {{- if .Values.logsCollection.extraFileLogs }}
-    logs/extraFiles:
+    {{- if (eq .Values.logsEngine "otel") }}
+    {{- if or (.Values.logsCollection.extraFileLogs) (.Values.logsCollection.journald.enabled) }}
```
dmitryax, line 496:
```suggestion
    {{- if and (eq .Values.logsEngine "otel") (or .Values.logsCollection.extraFileLogs .Values.logsCollection.journald.enabled) }}
```

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -489,6 +523,8 @@ service:
         {{- if eq (include "splunk-otel-collector.o11yLogsEnabled" .) "true" }}
         - splunk_hec/o11y
         {{- end }}
+        {{- end }}
+        {{- end }}
```
dmitryax, line 527: One `end` is shifted a bit. And if the suggestion above is applied, it can be removed

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -246,6 +246,29 @@ receivers:
   {{- if .Values.logsCollection.extraFileLogs }}
   {{- toYaml .Values.logsCollection.extraFileLogs | nindent 2 }}
   {{- end }}
+
+  # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/journaldreceiver
+  {{- if (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
+  {{- if and (eq .Values.logsEngine "otel") .Values.logsCollection.journald.enabled }}
```
dmitryax, line 252: This whole block is under the following condition:
`{{- if and (eq .Values.logsEngine "otel") .Values.logsCollection.containers.enabled }}` 
starting from line 113. 

It means that if `.Values.logsCollection.containers.enabled=false`, this block will be excluded regardless of `Values.logsCollection.journald.enabled`.

Please update line 113 to 
```smarty
  {{- if and (eq (include "splunk-otel-collector.logsEnabled" .) "true") (eq .Values.logsEngine "otel") }}
  {{- if .Values.logsCollection.containers.enabled }}
```

Add the following line after line 244
```smarty
  {{- end }}
```

and here we should have only one line instead:
```smarty
  {{- .Values.logsCollection.journald.enabled }}
```

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -246,6 +246,29 @@ receivers:
...
+      com.splunk.index: {{ $.Values.logsCollection.journald.index | default $.Values.splunkPlatform.index}}
+      host.name: 'EXPR(env("K8S_NODE_NAME"))'
+  {{- end }}
+  {{- else }}
+  journald:
+    directory: {{- toYaml .Values.logsCollection.journald.directory | nindent 6 }}
+    priority: {{ .Values.logsCollection.journald.defaultPriority }}
+  {{- end }}
+  {{- end }}
+  {{- end }}
```
dmitryax, line 271: Remove one end once the previous comment addressed

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -401,6 +401,22 @@ logsCollection:
     # All other logs will be ignored.
     useSplunkIncludeAnnotation: false
 
+# Configuration for collecting journald logs using otel collector
+  journald:
+    enabled: true
+    directory: /run/log/journal
```
rockb1017, line 407: "The journal service stores log data either persistently below `/var/log/journal` or in a volatile way below `/run/log/journal/`"

for better usability, can we have both fields as default? 
I know that journald receiver doesn't support array of `directory` parameter. 
so it may mean building 2 journald receiver, one for each directory unless there is a nice way to find which directory exists.

dmitryax: I agree that `/var/log/journal` is more common. I never saw anything under `/run/log/journal/`.

How collector with journald receiver behaves if provided directory is not present?

Ideally this should be solved in journald receiver code, it can have some kind of discovery if the field is not specified.

I don't think two receiver is a good workaround. As far as this field easily configurable, we just need to make sure user can read a warning message and update field if needed. By default I would keep `/var/log/journal` as most common to my knowledge.

WDYT?

luckyj5: `How collector with journald receiver behaves if provided directory is not present?` - Currently I think this is a required field for journald to work properly, if wrong directory path is provided, it will not work, it doesn't even error out I think. 

For now, we can do the following: 
- Update default to `/var/log/journald` and add a note to update it to /run/log/journal OR any other value if different? 
- Make it a required field, and comment it out so user has to provide an input for this field for journald to work?
- If we can check which one is available and acc update template with that directory path (not sure if there is an easy way to discover which path is present) 
- add 2 separate receivers?
- or if any other approach?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
luckyj5: Thanks @dmitryax. I'll update it. 
I think in this case we might need a separate pipeline for journald? 
Journald is not dependent on `{{- if .Values.logsCollection.containers.enabled }}` 
And I am guessing `extraFiles` is not dependent on `{{- if .Values.logsCollection.containers.enabled }}`  either.

Please let me know if otherwise.

dmitryax: This why we need to add `{{- end }}` after line 244 to close the second `{{- if .Values.logsCollection.containers.enabled }}` condition and keep the first one

helm-charts/splunk-otel-collector/values.yaml
dmitryax: The only requirement of this helm chart is that it must work OOTB with only two fields set in `splunkObservability` or/and `splunkPlatform`. We cannot introduce any other required fields like directory for journald. To make things simple and make this merged sooner I see only two reasonable options:
1. Disable journald by default and make the directory setting required. So if customer enables the journald they have to provide the path.
2. Keep journald enabled with most commonly used directory, which is `/var/log/journald` I believe. If we go this way we have to make sure that collector won't fail to start if that directory doesn't exist -- Please test this scenario. Ideally it should be a warning message saying that directory is not found.

luckyj5: I agree. 
We can keep the journald enabled and use `/var/log/journald` as default. If directory doesnt exist, collector doesnt fail to run, but it doesn't provide any warning as such. 
Will we need to update the receiver/ input to generate any such warning right now? or can be added later?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
luckyj5: Got it. I'll update this with other suggestions in the PR. thank you!

helm-charts/splunk-otel-collector/values.yaml
dmitryax: It can be added later. Noop if directory doesn't exist is fine for now.

rockb1017 (approved):

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
rockb1017: Hello, @dmitryax @luckyj5 
I put more thought into it and I think checkpoint migration for journald is not crucial. 
by default we are reading journald logs ignoring old records (`--follow` option) 
So for those who are migrating, they can just deploy olly chart then shutdown sck. Then it will have no data loss, only a slight data dups. 
Instead of having some data dups, we can simplify journald data pipelines and improve performance. 
What do you think?

dmitryax: Sounds good to me. I'm always up for simplifying things and improving performance!

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -469,12 +492,22 @@ service:
...
+        {{- end }}
+        {{- end }}
+        {{- if (.Values.logsCollection.journald.enabled)}}
+        {{- if .Values.logsCollection.journald.units }}
+        {{- range $_, $unit := .Values.logsCollection.journald.units }}
+        {{- printf "- journald/%s" $unit.name | nindent 8 }}
+        {{- end }}
+        {{- else }}
+        - journald
+        {{- end }}
```
dmitryax, line 510: One more `{{- end }}` is missed here

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: please remove one `{{- end }}` from here

dmitryax (commented):
Thanks for addressing templating comments. There is one closing tag still in a wrong place, otherwise all good

dmitryax (commented):
I tested it. It works fine but messages delivered to the backend are very verbose. I'm are getting the following json object with every journald log line:

```
{
    _SYSTEMD_UNIT: "kubelet.service",
    __CURSOR: "s=e75f3bdb84544363bdb3f8f12ac1247b;i=931cbd;b=ffaa807ac533404baa42e22954ba1c19;m=6824fca11a8;t=5d3c773f9f0ff;x=3bd4e5033d2206f1",
    MESSAGE: "I1223 03:01:21.265865 5650 kubelet_getters.go:176] "Pod status updated" pod="kube-system/kube-proxy-ip-172-20-44-183.us-west-1.compute.internal" status=Running",
    SYSLOG_FACILITY: "3",
    SYSLOG_IDENTIFIER: "kubelet",
    _MACHINE_ID: "ec24404cc124cb36f7826ac9399193ef",
    _SYSTEMD_CGROUP: "/system.slice/kubelet.service",
    _SYSTEMD_INVOCATION_ID: "55b80459852349758c104ab662cdc2fd",
    _CMDLINE: "/usr/local/bin/kubelet --anonymous-auth=false --authentication-token-webhook=true --authorization-mode=Webhook --cgroup-root=/ --client-ca-file=/srv/kubernetes/ca.crt --cloud-provider=aws --cluster-dns=100.64.0.10 --cluster-domain=cluster.local --enable-debugging-handlers=true --eviction-hard=memory.available<100Mi,nodefs.available<10%,nodefs.inodesFree<5%,imagefs.available<10%,imagefs.inodesFree<5% --hostname-override=ip-172-20-44-183.us-west-1.compute.internal --kubeconfig=/var/lib/kubelet/kubeconfig --network-plugin-mtu=9001 --network-plugin=kubenet --non-masquerade-cidr=100.64.0.0/10 --pod-infra-container-image=k8s.gcr.io/pause:3.2 --pod-manifest-path=/etc/kubernetes/manifests --register-schedulable=true --resolv-conf=/run/systemd/resolve/resolv.conf --v=2 --volume-plugin-dir=/usr/libexec/kubernetes/kubelet-plugins/volume/exec/ --cloud-config=/etc/kubernetes/cloud.config --cni-bin-dir=/opt/cni/bin/ --cni-conf-dir=/etc/cni/net.d/ --tls-cert-file=/srv/kubernetes/kubelet-server.crt --tls-private-key-file=/srv/kubernetes/kubelet-server.key",
    _HOSTNAME: "ip-172-20-44-183",
    _SYSTEMD_SLICE: "system.slice",
    _TRANSPORT: "stdout",
    _UID: "0",
    PRIORITY: "6",
    _BOOT_ID: "ffaa807ac533404baa42e22954ba1c19",
    _CAP_EFFECTIVE: "1ffffffffff",
    _COMM: "kubelet",
    _GID: "0",
    _PID: "5650",
    __MONOTONIC_TIMESTAMP: "7156754157992",
    _EXE: "/usr/local/bin/kubelet",
    _SELINUX_CONTEXT: "unconfined ",
    _STREAM_ID: "182885e1ae7c43b19691d13fb0b13b6c"
}
```

I'm not sure we want all of that data. Does SCK send payload as verbose as this for journald?

luckyj5:
> 

Thanks @dmitryax. I think yes, SCK has the same set of data as this. 
@rockb1017 can you please confirm? Thanks!

luckyj5:
https://github.com/signalfx/splunk-otel-collector-chart/pull/290#pullrequestreview-838975666
Thanks @dmitryax. I think yes, SCK has the same set of data as this.
@rockb1017 can you please confirm? Thanks!

luckyj5:
> I tested it. It works fine but messages delivered to the backend are very verbose. I'm are getting the following json object with every journald log line:
> 
> ```
> {
>     _SYSTEMD_UNIT: "kubelet.service",
>     __CURSOR: "s=e75f3bdb84544363bdb3f8f12ac1247b;i=931cbd;b=ffaa807ac533404baa42e22954ba1c19;m=6824fca11a8;t=5d3c773f9f0ff;x=3bd4e5033d2206f1",
>     MESSAGE: "I1223 03:01:21.265865 5650 kubelet_getters.go:176] "Pod status updated" pod="kube-system/kube-proxy-ip-172-20-44-183.us-west-1.compute.internal" status=Running",
>     SYSLOG_FACILITY: "3",
>     SYSLOG_IDENTIFIER: "kubelet",
>     _MACHINE_ID: "ec24404cc124cb36f7826ac9399193ef",
>     _SYSTEMD_CGROUP: "/system.slice/kubelet.service",
>     _SYSTEMD_INVOCATION_ID: "55b80459852349758c104ab662cdc2fd",
>     _CMDLINE: "/usr/local/bin/kubelet --anonymous-auth=false --authentication-token-webhook=true --authorization-mode=Webhook --cgroup-root=/ --client-ca-file=/srv/kubernetes/ca.crt --cloud-provider=aws --cluster-dns=100.64.0.10 --cluster-domain=cluster.local --enable-debugging-handlers=true --eviction-hard=memory.available<100Mi,nodefs.available<10%,nodefs.inodesFree<5%,imagefs.available<10%,imagefs.inodesFree<5% --hostname-override=ip-172-20-44-183.us-west-1.compute.internal --kubeconfig=/var/lib/kubelet/kubeconfig --network-plugin-mtu=9001 --network-plugin=kubenet --non-masquerade-cidr=100.64.0.0/10 --pod-infra-container-image=k8s.gcr.io/pause:3.2 --pod-manifest-path=/etc/kubernetes/manifests --register-schedulable=true --resolv-conf=/run/systemd/resolve/resolv.conf --v=2 --volume-plugin-dir=/usr/libexec/kubernetes/kubelet-plugins/volume/exec/ --cloud-config=/etc/kubernetes/cloud.config --cni-bin-dir=/opt/cni/bin/ --cni-conf-dir=/etc/cni/net.d/ --tls-cert-file=/srv/kubernetes/kubelet-server.crt --tls-private-key-file=/srv/kubernetes/kubelet-server.key",
>     _HOSTNAME: "ip-172-20-44-183",
>     _SYSTEMD_SLICE: "system.slice",
>     _TRANSPORT: "stdout",
>     _UID: "0",
>     PRIORITY: "6",
>     _BOOT_ID: "ffaa807ac533404baa42e22954ba1c19",
>     _CAP_EFFECTIVE: "1ffffffffff",
>     _COMM: "kubelet",
>     _GID: "0",
>     _PID: "5650",
>     __MONOTONIC_TIMESTAMP: "7156754157992",
>     _EXE: "/usr/local/bin/kubelet",
>     _SELINUX_CONTEXT: "unconfined ",
>     _STREAM_ID: "182885e1ae7c43b19691d13fb0b13b6c"
> }
> ```
> 
> I'm not sure we want all of that data. Does SCK send payload as verbose as this for journald?

Thanks @dmitryax. I think yes, SCK has the same set of data as this.
@rockb1017 can you please confirm? Thanks!

luckyj5:
> > I tested it. It works fine but messages delivered to the backend are very verbose. I'm are getting the following json object with every journald log line:
> > ```
> > {
> >     _SYSTEMD_UNIT: "kubelet.service",
> >     __CURSOR: "s=e75f3bdb84544363bdb3f8f12ac1247b;i=931cbd;b=ffaa807ac533404baa42e22954ba1c19;m=6824fca11a8;t=5d3c773f9f0ff;x=3bd4e5033d2206f1",
> >     MESSAGE: "I1223 03:01:21.265865 5650 kubelet_getters.go:176] "Pod status updated" pod="kube-system/kube-proxy-ip-172-20-44-183.us-west-1.compute.internal" status=Running",
> >     SYSLOG_FACILITY: "3",
> >     SYSLOG_IDENTIFIER: "kubelet",
> >     _MACHINE_ID: "ec24404cc124cb36f7826ac9399193ef",
> >     _SYSTEMD_CGROUP: "/system.slice/kubelet.service",
> >     _SYSTEMD_INVOCATION_ID: "55b80459852349758c104ab662cdc2fd",
> >     _CMDLINE: "/usr/local/bin/kubelet --anonymous-auth=false --authentication-token-webhook=true --authorization-mode=Webhook --cgroup-root=/ --client-ca-file=/srv/kubernetes/ca.crt --cloud-provider=aws --cluster-dns=100.64.0.10 --cluster-domain=cluster.local --enable-debugging-handlers=true --eviction-hard=memory.available<100Mi,nodefs.available<10%,nodefs.inodesFree<5%,imagefs.available<10%,imagefs.inodesFree<5% --hostname-override=ip-172-20-44-183.us-west-1.compute.internal --kubeconfig=/var/lib/kubelet/kubeconfig --network-plugin-mtu=9001 --network-plugin=kubenet --non-masquerade-cidr=100.64.0.0/10 --pod-infra-container-image=k8s.gcr.io/pause:3.2 --pod-manifest-path=/etc/kubernetes/manifests --register-schedulable=true --resolv-conf=/run/systemd/resolve/resolv.conf --v=2 --volume-plugin-dir=/usr/libexec/kubernetes/kubelet-plugins/volume/exec/ --cloud-config=/etc/kubernetes/cloud.config --cni-bin-dir=/opt/cni/bin/ --cni-conf-dir=/etc/cni/net.d/ --tls-cert-file=/srv/kubernetes/kubelet-server.crt --tls-private-key-file=/srv/kubernetes/kubelet-server.key",
> >     _HOSTNAME: "ip-172-20-44-183",
> >     _SYSTEMD_SLICE: "system.slice",
> >     _TRANSPORT: "stdout",
> >     _UID: "0",
> >     PRIORITY: "6",
> >     _BOOT_ID: "ffaa807ac533404baa42e22954ba1c19",
> >     _CAP_EFFECTIVE: "1ffffffffff",
> >     _COMM: "kubelet",
> >     _GID: "0",
> >     _PID: "5650",
> >     __MONOTONIC_TIMESTAMP: "7156754157992",
> >     _EXE: "/usr/local/bin/kubelet",
> >     _SELINUX_CONTEXT: "unconfined ",
> >     _STREAM_ID: "182885e1ae7c43b19691d13fb0b13b6c"
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm not sure we want all of that data. Does SCK send payload as verbose as this for journald?
> 
> Thanks @dmitryax. I think yes, SCK has the same set of data as this. @rockb1017 can you please confirm? Thanks!

Looks like we are filtering in journald events in SCK, will update this PR to have similar filtering. Thanks for confirming @rockb1017

luckyj5:
> > > I tested it. It works fine but messages delivered to the backend are very verbose. I'm are getting the following json object with every journald log line:
> > > ```
> > > {
> > >     _SYSTEMD_UNIT: "kubelet.service",
> > >     __CURSOR: "s=e75f3bdb84544363bdb3f8f12ac1247b;i=931cbd;b=ffaa807ac533404baa42e22954ba1c19;m=6824fca11a8;t=5d3c773f9f0ff;x=3bd4e5033d2206f1",
> > >     MESSAGE: "I1223 03:01:21.265865 5650 kubelet_getters.go:176] "Pod status updated" pod="kube-system/kube-proxy-ip-172-20-44-183.us-west-1.compute.internal" status=Running",
> > >     SYSLOG_FACILITY: "3",
> > >     SYSLOG_IDENTIFIER: "kubelet",
> > >     _MACHINE_ID: "ec24404cc124cb36f7826ac9399193ef",
> > >     _SYSTEMD_CGROUP: "/system.slice/kubelet.service",
> > >     _SYSTEMD_INVOCATION_ID: "55b80459852349758c104ab662cdc2fd",
> > >     _CMDLINE: "/usr/local/bin/kubelet --anonymous-auth=false --authentication-token-webhook=true --authorization-mode=Webhook --cgroup-root=/ --client-ca-file=/srv/kubernetes/ca.crt --cloud-provider=aws --cluster-dns=100.64.0.10 --cluster-domain=cluster.local --enable-debugging-handlers=true --eviction-hard=memory.available<100Mi,nodefs.available<10%,nodefs.inodesFree<5%,imagefs.available<10%,imagefs.inodesFree<5% --hostname-override=ip-172-20-44-183.us-west-1.compute.internal --kubeconfig=/var/lib/kubelet/kubeconfig --network-plugin-mtu=9001 --network-plugin=kubenet --non-masquerade-cidr=100.64.0.0/10 --pod-infra-container-image=k8s.gcr.io/pause:3.2 --pod-manifest-path=/etc/kubernetes/manifests --register-schedulable=true --resolv-conf=/run/systemd/resolve/resolv.conf --v=2 --volume-plugin-dir=/usr/libexec/kubernetes/kubelet-plugins/volume/exec/ --cloud-config=/etc/kubernetes/cloud.config --cni-bin-dir=/opt/cni/bin/ --cni-conf-dir=/etc/cni/net.d/ --tls-cert-file=/srv/kubernetes/kubelet-server.crt --tls-private-key-file=/srv/kubernetes/kubelet-server.key",
> > >     _HOSTNAME: "ip-172-20-44-183",
> > >     _SYSTEMD_SLICE: "system.slice",
> > >     _TRANSPORT: "stdout",
> > >     _UID: "0",
> > >     PRIORITY: "6",
> > >     _BOOT_ID: "ffaa807ac533404baa42e22954ba1c19",
> > >     _CAP_EFFECTIVE: "1ffffffffff",
> > >     _COMM: "kubelet",
> > >     _GID: "0",
> > >     _PID: "5650",
> > >     __MONOTONIC_TIMESTAMP: "7156754157992",
> > >     _EXE: "/usr/local/bin/kubelet",
> > >     _SELINUX_CONTEXT: "unconfined ",
> > >     _STREAM_ID: "182885e1ae7c43b19691d13fb0b13b6c"
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I'm not sure we want all of that data. Does SCK send payload as verbose as this for journald?
> > 
> > 
> > Thanks @dmitryax. I think yes, SCK has the same set of data as this. @rockb1017 can you please confirm? Thanks!
> 
> Looks like we are filtering in journald events in SCK, will update this PR to have similar filtering. Thanks for confirming @rockb1017

@dmitryax @rockb1017 Added filtering to include the below fields and discard the rest
`
"$$._SYSTEMD_UNIT", 
"$$.MESSAGE", and
"$$.PRIORITY"
`

Sample event:
```
{"MESSAGE":"time=\"2022-01-05T01:50:40.894233600Z\" level=info msg=\"ignoring event\" container=b575d0e82ac2c6b0d556e1c1349e98fe38ef63af83d97abbec0e483f4eb1ecef module=libcontainerd namespace=moby topic=/tasks/delete type=\"*events.TaskDelete\"",
"PRIORITY":"6",
"_SYSTEMD_UNIT":"docker.service"}
```

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
rockb1017: > Hello, @dmitryax @luckyj5 I put more thought into it and I think checkpoint migration for journald is not crucial. by default we are reading journald logs ignoring old records (`--follow` option) So for those who are migrating, they can just deploy olly chart then shutdown sck. Then it will have no data loss, only a slight data dups. Instead of having some data dups, we can simplify journald data pipelines and improve performance. What do you think?

Thanks for updating. Could you make this change? making it simplied? so instead of having a journald receiver for each service unit, we would have it grouped by priority level.

luckyj5: Okay. Will group it my priority. Thanks!

dmitryax:
> @dmitryax @rockb1017 Added filtering to include the below fields and discard the rest `"$$._SYSTEMD_UNIT", "$$.MESSAGE", and "$$.PRIORITY"`
> 
> Sample event:
> 
> ```
> {"MESSAGE":"time=\"2022-01-05T01:50:40.894233600Z\" level=info msg=\"ignoring event\" container=b575d0e82ac2c6b0d556e1c1349e98fe38ef63af83d97abbec0e483f4eb1ecef module=libcontainerd namespace=moby topic=/tasks/delete type=\"*events.TaskDelete\"",
> "PRIORITY":"6",
> "_SYSTEMD_UNIT":"docker.service"}
> ```

Given that we have now only predefined fields maybe better to extract `MESSAGE` into the log body, and set `PRIORITY` to `journald.priority.number` and `_SYSTEMD_UNIT` into `journald.unit.name` attributes?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: This can be addressed in another PR. @luckyj5 feel free to submit an issue

rockb1017:
> Given that we have now only predefined fields maybe better to extract `MESSAGE` into the log body, and set `PRIORITY` to `journald.priority.number` and `_SYSTEMD_UNIT` into `journald.unit.name` attributes? cc @rockb1017

Yea makes sense!

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
luckyj5: Created new issue for tracking this - https://github.com/signalfx/splunk-otel-collector-chart/issues/360

dmitryax:
It doesn't for me now. I'm getting these errors

```
2022-01-19T05:40:45.599Z	error	Failed to process entry	{"kind": "receiver", "name": "journald", "operator_id": "$.metadata", "operator_type": "metadata", "error": {"description": "failed to add resource keys to entry: render embedded expression: invalid operation: string + <nil> (1:8)\n | \"kube:\"+$._SYSTEMD_UNIT\n | .......^"}, "action": "send", "entry": {"timestamp":"2022-01-10T12:59:37.414365Z","body":{"MESSAGE":"Disconnected from authenticating user root 101.42.101.139 port 45992 [preauth]","PRIORITY":"6","SYSLOG_FACILITY":"4","SYSLOG_IDENTIFIER":"sshd","SYSLOG_PID":"3043694","SYSLOG_TIMESTAMP":"Jan 10 12:59:37 ","_BOOT_ID":"94dd12ba0583423c8703e9195b8074b6","_GID":"0","_HOSTNAME":"ip-172-20-36-112","_MACHINE_ID":"ec2ffb32d5187405b7eac9b9a1b87a4b","_PID":"3043694","_SOURCE_REALTIME_TIMESTAMP":"1641819577413275","_TRANSPORT":"syslog","_UID":"0","__CURSOR":"s=1e88599f1b0b4b3ab1eb058fa87c8ba5;i=98a81d;b=94dd12ba0583423c8703e9195b8074b6;m=7f4c4d9d511;t=5d539e8b272dd;x=5111cb02c2128825","__MONOTONIC_TIMESTAMP":"8747856024849"},"severity":0}}
```

luckyj5:
> Did you test? It doesn't for me now. I'm getting these errors
> 
> ```
> 2022-01-19T05:40:45.599Z	error	Failed to process entry	{"kind": "receiver", "name": "journald", "operator_id": "$.metadata", "operator_type": "metadata", "error": {"description": "failed to add resource keys to entry: render embedded expression: invalid operation: string + <nil> (1:8)\n | \"kube:\"+$._SYSTEMD_UNIT\n | .......^"}, "action": "send", "entry": {"timestamp":"2022-01-10T12:59:37.414365Z","body":{"MESSAGE":"Disconnected from authenticating user root 101.42.101.139 port 45992 [preauth]","PRIORITY":"6","SYSLOG_FACILITY":"4","SYSLOG_IDENTIFIER":"sshd","SYSLOG_PID":"3043694","SYSLOG_TIMESTAMP":"Jan 10 12:59:37 ","_BOOT_ID":"94dd12ba0583423c8703e9195b8074b6","_GID":"0","_HOSTNAME":"ip-172-20-36-112","_MACHINE_ID":"ec2ffb32d5187405b7eac9b9a1b87a4b","_PID":"3043694","_SOURCE_REALTIME_TIMESTAMP":"1641819577413275","_TRANSPORT":"syslog","_UID":"0","__CURSOR":"s=1e88599f1b0b4b3ab1eb058fa87c8ba5;i=98a81d;b=94dd12ba0583423c8703e9195b8074b6;m=7f4c4d9d511;t=5d539e8b272dd;x=5111cb02c2128825","__MONOTONIC_TIMESTAMP":"8747856024849"},"severity":0}}
> ```
> 
> Looks like `_SYSTEMD_UNIT` is not passed by the operator. Not sure why.

@dmitryax Thanks for checking. Updated the template to remove `_SYSTEMD_UNIT` when no journald.units are provided in the values.yaml config.

dmitryax:
This is sad that not specifying units explicitly doesn't give us `journald.unit.name` nor `com.splunk.sourcetype`. Without those fields it's in a kind of a broken state. For me as a user, I don't think it's even useful if I cannot understand which service a message is coming from.

Do you know if this is a journald limitation, or something that can be fixed in journald operator of otel logs collection library?

If we cannot do anything with that on otel collector side, I would suggest even removing the option to collect logs from all units and always require specifying the units explicitly. We can have 3 units enabled by default as you provided initially in this PR, but that it's introducing 3 new receivers with the current implementation. What if we switch to the 3 units specified explicitly, but keep journald logs collection disabled by default. And once [this PR](https://github.com/signalfx/splunk-otel-collector-chart/issues/360) is resolved, we enable it by default. How does it sound?

luckyj5:
> This is sad that not specifying units explicitly doesn't give us `journald.unit.name` nor `com.splunk.sourcetype`. Without those fields it's in a kind of a broken state. For me as a user, I don't think it's even useful if I cannot understand which service a message is coming from.
> 
> Do you know if this is a journald limitation, or something that can be fixed in journald operator of otel logs collection library?
> 
> If we cannot do anything with that on otel collector side, I would suggest even removing the option to collect logs from all units and always require specifying the units explicitly. We can have 3 units enabled by default as you provided initially in this PR, but that it's introducing 3 new receivers with the current implementation. What if we switch to the 3 units specified explicitly, but keep journald logs collection disabled by default. And once [this issue](https://github.com/signalfx/splunk-otel-collector-chart/issues/360) is resolved, we enable it by default. How does it sound?

@dmitryax
The way journalctl works is `if called without parameters, it will show the full contents of the journal, starting with the oldest entry collected.
If one or more match arguments are passed, the output is filtered accordingly.`
Read more at: https://www.commandlinux.com/man-page/man1/journalctl.1.html

Units are configurable and thats how journald input works as well [Journald input](https://github.com/open-telemetry/opentelemetry-log-collection/blob/0197a907af8c078060d62094ca832bfe038b39c2/operator/builtin/input/journald/journald.go#L52), it provides unit value if its a part of the supplied configuration. 

Units parameter supports PATTERN match, if configured, units parameter show messages for the specified systemd unit UNIT, or for any of the units matched by PATTERN. So potentially we can collect all units (when not configured) by using a PATTERN that will match all units. (This will require a change on journald input operator side) 

Question is, do we want to collect all units by default when units are not configured in values.yaml  There can be many service units enabled and might not all be that useful to the user? Eg:
image


The reason we are not collecting all by default is, it creates lot of noise, and hence we started this PR with `Units` being required field (with default values) which we did not like as we were trying to go away from more required parameters in the configuration. 

Here are the options I think we have:
1. Leave it as is where if units are not configured, unit.name and sourcetype wont have service name. Thats how journalctl works as well. We can have any other value for sourcetype (eg: kube:journald?) 
2. Update the config to have 3 default unit values (make it a required field), keep journald disabled by default and enable it once we update the receivers to be created by priority level.
3. If we are interested in collecting all the units, update journald-input operator and then update the template on our side. 

I know we are trying to have the best possible version of journald data collection with more flexibility and controlled filtering, but if something seems not reasonable or not user friendly, we can always update it to go with the best possible approach based on the feedback we receive from the users or community?

Please let me know what you think and which option seems reasonable as of now? Thank you!

dmitryax:
Thanks for providing the detailed answer.

> 1. Leave it as is where if units are not configured, unit.name and sourcetype wont have service name. Thats how journalctl works as well. We can have any other value for sourcetype (eg: kube:journald?)
As a user, I don't really like this solution. I always want to know which service logs are coming from. And I'm sure this would be expected behavior from user users.

> 2. Update the config to have 3 default unit values (make it a required field), keep journald disabled by default and enable it once this issue is resolved.
I think we should go this route and completely remove an option to specify all units. Once we have a solution on the operator side as use suggested, we can add the all units option back, but still probably not as a default behavior since you mentioned that it's noisy. 

What do you think?

luckyj5:
> What do you think?

Sounds good to me, we can have all units option if/ when there is an ask.

luckyj5:
@dmitryax Updated the PR back to have units configured by default, and disable journald by default in the config. 
Let me know if anything. Thanks!

docs/advanced-configuration.md
```diff
@@ -195,6 +195,29 @@ logsCollection:
...
+    directory: /run/log/journal
+    # List of service units to collect and configuration for each. Please update the list as needed.
+    units:
+      - name: kubelet
+        priority: info
+      - name: docker
+        priority: info
+      - name: containerd
+       priority: info
+    # Route journald logs to its own Splunk Index by specifying the index value below, else leave it blank. Please make sure the index exist in Splunk and is configured to receive HEC traffic.
```
dmitryax, line 217: Can we add "Not applicable to Splunk Observability"?

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -416,6 +421,15 @@ spec:
...
           type: DirectoryOrCreate
+      {{- if .Values.logsCollection.journald.enabled}}
+      - name: journaldlogs
+        hostPath:
+          path: {{.Values.logsCollection.journald.directory}}
+      {{- end}}
+      - name: old-checkpoint
+        hostPath:
+          path: /var/lib/otel_pos
+          type: DirectoryOrCreate
```
dmitryax, line 432: This is something unrelated right? Probably came from a rebase

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Journald support (#290)
```
dmitryax, line 9:
```suggestion
- Journald logs support (#290)
```
?

luckyj5:
@dmitryax updated. 
Let me know if there is anything which needs to be updated. Thanks!

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -272,10 +273,38 @@ receivers:
...
+  {{- range $_, $unit := .Values.logsCollection.journald.units }}
+  {{- printf "journald/%s:" $unit.name | nindent 2 }}
+    directory: {{ $.Values.logsCollection.journald.directory }}
+    units: [{{ $unit.name }}]
+    priority: {{ $unit.priority }}
+    operators:
+    - type: metadata
+      resource:
+        com.splunk.source: {{ $.Values.logsCollection.journald.directory }}
+        com.splunk.sourcetype: 'EXPR("kube:"+$$._SYSTEMD_UNIT)'
```
dmitryax, line 293: Is this the value that SCK uses for sourcetype? Now we have `kube:container:<conatiner_name>` and `kube:<journald_unit_name>`. Looks inconsistent to me. Shouldn't it be `kube:journald:<journald_unit_name>`?

cc @rockb1017

dmitryax (commented):
It looks good to me! Just one last question

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
luckyj5: Yeah, kube:journald: would make more sense. 

SCK has this format -  `kube:docker` for docker.service unit, `kube:kubelet` for kubelet.service unit

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -532,6 +532,28 @@
...
+            "directory": {
+              "type": "string"
+            },
+            "index": {
+              "type": "string"
+            },
+            "units": {
+              "type": "array",
+              "items": {
+                "type": "object"
```
dmitryax, line 552: Hey @luckyj5 , I'm sorry to keep bugging you with these small asks, but I just found this.

We need to specify schema for this object since it can only have two keys: `name` and `priority`.

I believe we can merge the PR after that

dmitryax (approved):
Looks good to me!

PR #291, opened 2021-11-16
improve chart test

rockb1017:


helm-charts/splunk-otel-collector/ci/logs-only-otel-se-o11y-values.yaml
```diff
@@ -0,0 +1,20 @@
...
+  realm: fake-realm
+  accessToken: fake-token
+
+  tracesEnabled: false
+  metricsEnabled: false
+
+splunkPlatform:
+  endpoint: "fake-endpoint"
+  # Required for Splunk Enterprise/Cloud (if `endpoint` is specified). Splunk
+  # HTTP Event Collector token.
```
dmitryax, line 12: Nit: could you please remove the comment since other parts are not commented (and shouldn't be I think)

dmitryax (approved):

PR #292, opened 2021-11-16
edit default checkpoint path

rockb1017:
To make deployment in Autopilot work out of the box

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,11 @@ guidelines](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UP
 
 - Extract `container.image.tag` attribute from `container.image.name` (#285)
 - Upgrade splunk-otel-collector image to 0.38.1 (#284)
+<<<<<<< HEAD
 - Change secret names according to the GDI specification (#295)
+=======
+- Changed the default checkpoint path to `/var/addon/splunk/otel_pos` (#292)
+>>>>>>> f8979ec8b53d (add init container for copying old checkpoint)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,8 @@ guidelines](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UP
 
 - Extract `container.image.tag` attribute from `container.image.name` (#285)
 - Upgrade splunk-otel-collector image to 0.38.1 (#284)
-<<<<<<< HEAD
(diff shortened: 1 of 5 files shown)


dmitryax:
We also need to add a copy command in the initContainer to avoid re-reading all the logs on upgrade. There are few Observability customers using the native logs collection

rockb1017:
> We also need to add a copy command in the initContainer to avoid re-reading all the logs on upgrade. There are few Observability customers using the native logs collection

hmm so mount both directories and copy files if it doesn't already exist in the new path?

rockb1017:
I am using `cp -n` for no overwrite. and making sure it always exit 0 with ` || true`
```
-n, --no-clobber
              do not overwrite an existing file (overrides a previous -i option)
```

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -94,6 +94,19 @@ spec:
             - name: fluentd-config-cri
               mountPath: /fluentd/etc/cri
         {{- else }}
+        - name: copy-old-checkpoint
+          image: {{ template "splunk-otel-collector.image.fluentd" . }}
```
dmitryax, line 98: Let's use splunk-otel-collector image here so we don't pull an extra image if native logs collection enabled

CHANGELOG.md
```diff
@@ -15,7 +15,11 @@ guidelines](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UP
 
 - Extract `container.image.tag` attribute from `container.image.name` (#285)
 - Upgrade splunk-otel-collector image to 0.38.1 (#284)
+<<<<<<< HEAD
 - Change secret names according to the GDI specification (#295)
+=======
+- Changed the default checkpoint path to `/var/addon/splunk/otel_pos` (#292)
+>>>>>>> f8979ec8b53d (add init container for copying old checkpoint)
```
dmitryax, line 22: Please resolve

dmitryax (approved):

dmitryax:
@rockb1017 Follow-up: we have the `checkpointPath` configuration option but all the initContainers use hardcoded `/var/addon/splunk/otel_pos` value instead. 

I believe changing `checkpointPath` breaks the deployment. We need to decide how to solve it:
1. Make sure `checkpointPath` is used in all initContainers
2. Remove `checkpointPath` option and use hardcoded values instead.

I vote for 2. We can always bring that configuration back if there will be a need. Please let me know what do you think

PR #296, opened 2021-11-17
Make `clusterName` configuration parameter generally required

dmitryax:


README.md
```diff
@@ -134,8 +134,9 @@ For Splunk Observability Cloud the following parameters are required:
 - `splunkObservability.realm`: Splunk realm to send telemetry data to.
 - `splunkObservability.accessToken`: Your Splunk Observability org access
   token.
-- `clusterName`: arbitrary value that will identify your Kubernetes cluster in
-  Splunk Observability Cloud.
+
+The following parameter is required for any of the destinations:
+- `clusterName`: arbitrary value that identifies your Kubernetes cluster.
```
rmfitzpatrick, line 139: The description from the schema is helpful imo:

```suggestion
- `clusterName`: arbitrary value that identifies your Kubernetes cluster. The value will be associated with every trace, metric and log as "k8s.cluster.name" attribute.
```

rmfitzpatrick (approved):

PR #305, opened 2021-11-18
use  as default index and disable metrics by default for Splunk platform

rockb1017:


dmitryax (approved):

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -65,7 +65,7 @@ splunkPlatform:
   # Options to disable particular telemetry data types that will be sent to
```
dmitryax, line 65:
```suggestion
  # Options to disable or enable particular telemetry data types that will be sent to
```

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -65,7 +65,7 @@ splunkPlatform:
...
   # Splunk Platform.
```
dmitryax, line 66:
```suggestion
  # Splunk Platform. Only logs collection is enabled by default.
```

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -62,10 +62,10 @@ splunkPlatform:
   #       The file will be stored as a secret in kubernetes.
   caFile: ""
 
-  # Options to disable particular telemetry data types that will be sent to
-  # Splunk Platform.
+  # Options to disable or enable particular telemetry data types that will be sent to
+  # Splunk Platform. Only logs collection is enabled by default.
   logsEnabled: true
-  metricsEnabled: true
+  metricsEnabled: false
```
chaitanyaphalak, line 68: Why are we disabling metrics by default?

rockb1017: because user must create metric type index and provide that name to `splunkPlatform.metrics_index`

dmitryax: Yeah, we want to make sure that chart is working by default just with `token` and `endpoint` as stated in the docs

dmitryax:
```suggestion
  # If you enable metrics collection, make sure that `metrics_index` is provided as well. 
  metricsEnabled: false
```

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -62,10 +62,11 @@ splunkPlatform:
   #       The file will be stored as a secret in kubernetes.
   caFile: ""
 
-  # Options to disable particular telemetry data types that will be sent to
-  # Splunk Platform.
+  # Options to disable or enable particular telemetry data types that will be sent to
+  # Splunk Platform. Only logs collection is enabled by default.
   logsEnabled: true
-  metricsEnabled: true
+  # If you enable metrics collection, make sure that `metrics_index` is provided as well.
```
rockb1017, line 68:
```suggestion
  # If you enable metrics collection, make sure that `metricsIndex` is provided as well.
```

PR #313, opened 2021-11-19
Remove extraArgs config parameter

dmitryax:
With the latest collector releases, all arguments were moved to configuration. And `extraArgs` parameter is already removed from the schema, so it cannot be used since 0.37.0.


rockb1017 (approved):

rockb1017:
looks good. one thing, no need to remove from json schema?

dmitryax:
I mentioned in the PR comment that it was already removed some time ago

PR #319, opened 2021-11-19
Move "Migration from SCK" doc to docs directory

dmitryax:


rockb1017 (approved):

rockb1017:
the link in header needs to be updated if we move it.

dmitryax:
I actually added the link assuming it goes to docs directory. Just missed it during review

PR #323, opened 2021-11-30
Remove redundant secret from fluentd container

dmitryax:
It was used when fluent was sending data directly to the backend, not being used anymore

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/daemonset.yaml
+++ b/helm-charts/splunk-otel-collector/templates/daemonset.yaml
@@ -217,9 +217,6 @@ spec:
           readOnly: true
         - name: fluentd-config
           mountPath: /fluentd/etc
-        - name: secrets
-          mountPath: /fluentd/etc/splunk
-          readOnly: true
       {{- end }}
       - name: otel-collector
         command:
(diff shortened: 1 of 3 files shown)


rmfitzpatrick (approved):

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -388,9 +385,6 @@ spec:
       - name: journallogpath
         hostPath:
           path: {{ .Values.fluentd.config.journalLogPath | quote }}
-      - name: secrets
-        secret:
-          secretName: {{ template "splunk-otel-collector.secret" . }}
```
rockb1017, line 393: this is used for Otel collector as well for CA bundle. (https) line 361

rockb1017 (changes requested):

helm-charts/splunk-otel-collector/templates/daemonset.yaml
dmitryax: That one is different:`name: secret`

rockb1017 (approved):
LGTM

Issue #328, opened 2021-12-01
Allow helm chart to be used as subchart


gusdebbaja:
Hi, 

As this chart has an enforced values schema, it cannot be deployed as a subchart. This is because `global` is not an allowed property in the schema.

Solution: 
Add global property in the value schema: 
Add this in line 11:
```
  "properties": {
    "global": {
      "type": "object"
    },
```

How to reproduce this: 
Create the following folder structure (or just type `helm create parentChart` ): 
- parentChart
  - charts
    - Chart.yaml (see below)
    - values.yaml (contain general values)
  - values: 
    - param1.yaml (param1 specific values)
    - param2.yaml (param2 specific values)
 
Chart.yaml contains:
```
apiVersion: v2
name: parentChart
version: 1.0.0
dependencies:
  - name: splunk-otel-collector
    version: 0.39.0
    repository: https://signalfx.github.io/splunk-otel-collector-chart
```

Value files have this format: 
```
splunk-otel-collector:
  - clusterName: myCluster
  - rest config here....
```

run `helm install parentChart charts --dry-run --values values/param1.yaml`

Output: 
```
splunk-otel-collector:
- (root): Additional property global is not allowed
```

dmitryax:
@illrill thanks for taking care of it

PR #330, opened 2021-12-01
Allow global property in values schema

illrill:
Fix for #328; Enables this chart to be used as a sub chart (dependency).

Same issue and fix as in https://github.com/istio/istio/issues/35495 / https://github.com/istio/istio/pull/35496


dmitryax (approved):
LGTM

PR #331, opened 2021-12-01
updated helm gateway examples

harnitsignalfx:
For the "Deploy gateway-mode Otel Collector only"
We need logsEnabled to true, so it can receive logs over otlp and forward over to splunk_hec/o11y


dmitryax (approved):
LGTM!

PR #333, opened 2021-12-07
Update release instructions

dmitryax:


RELEASE.md
```diff
@@ -1,4 +1,6 @@
 ## Release
-To make a release bump the version in [Chart.yaml](helm-charts/splunk-otel-collector/Chart.yaml). When it gets merged the release will automatically be made and the helm repo updated.
 
-If you have a dependent chart consuming this chart, bump the version dependency if needed and run `helm dep update`. Commit the changed and added files.
+To make a new release of the helm chart:
+1. Bump the version in [Chart.yaml](helm-charts/splunk-otel-collector/Chart.yaml) and create PR.
```
rmfitzpatrick, line 4: When would we update `appVersion`?  What is their relation to each other?

rockb1017: appVersion is the image tag of otel collector

dmitryax: Yes `appVersion` is not necessarily required. We can release a new helm version with helm only updates, usually it goes as a new patch version. I can mention the relation here

rmfitzpatrick: Sry, shouldn't use leading questions. I pieced that together and thought it warranted an addition since `appVersion` is ambiguous imo.

rmfitzpatrick (approved):
thx!

mstumpfx (approved):

PR #338, opened 2021-12-13
Add GKE Autopilot support

dmitryax:
Docs and configuration changes to support GKE Autopilot operation mode. The change depends on a [kubelet read-only endpoint support](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6488) which was added to OpenTelemetry Collector in 0.41.0

### DIFF
--- a/docs/advanced-configuration.md
+++ b/docs/advanced-configuration.md
@@ -80,6 +80,44 @@ splunkPlatform:
   logsEnabled: true
 ```
 
+## GKE Autopilot support
+
+If you want to run Splunk OTel Collector in [Google Kubernetes Engine
+Autopilot](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-overview),
+make sure to set `distribution` setting to `gke/autopilot`:
+
+```yaml
+distribution: gke/autopilot
+```
+
+**NOTE:** Native OTel logs collection is not yet supported in GKE Autopilot.
+
+Sometimes Splunk OTel Collector agent daemonset can have [problems scheduling in
+Autopilot](https://cloud.google.com/kubernetes-engine/docs/concepts/daemonset#autopilot-ds-best-practices)
+If you run into these issues, you can assign the daemonset a higher [priority
+class](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/),
+this will make sure that the daemonset pods are always present on each node:
+
+1. Create a new priority class for Splunk OTel Collector agent:
+
+cat <<EOF | kubectl apply -f -
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: splunk-otel-agent-priority
+value: 1000000
+globalDefault: false
+description: "Higher priority class for Splunk OpenTelemetry Collector pods."
+EOF
+
+2. Use the created priority class in the helm install/upgrade command:
+with `--set="priorityClassName=splunk-otel-agent-priority"` cli argument or add
+the following line to your custom values.yaml:
+
+```yaml
+priorityClassName: splunk-otel-agent-priority
+```
+
 ## Logs collection
 
 The helm chart currently utilizes [fluentd](https://docs.fluentd.org/) for Kubernetes logs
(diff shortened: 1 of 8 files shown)


rockb1017:
in autopilot, is running it as user root possible?

docs/advanced-configuration.md
```diff
@@ -80,6 +80,42 @@ splunkPlatform:
...
+Autopilot](https://cloud.google.com/kubernetes-engine/docs/concepts/daemonset#autopilot-ds-best-practices)
+If you run into these issues, you can assign the daemonset a higher [priority
+class](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/),
+this will make sure that the daemonset pods are always present on each node:
+
+1. Create a new priority class for Splunk OTel Collector agent:
+
+cat <<EOF | kubectl apply -f -
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
```
rmfitzpatrick, line 103: Is this not something the distribution can handle by default with a flag to disable if desired?

docs/advanced-configuration.md
```diff
@@ -103,6 +139,7 @@ There are following known limitations of native OTel logs collection:
   This means that correlation between logs and traces will not work in Splunk Observability.
   Logs collection with fluentd is still recommended if chart deployed with `autodetect.istio=true`.
 - Journald logs cannot be collected natively by Splunk OTel Collector yet.
+- Not yet supported in GKE Autopilot.
```
rmfitzpatrick, line 142: way want to use bold or place in a more disclaimer-ish position near the distribution description.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -77,8 +77,15 @@ receivers:
 
   kubeletstats:
     collection_interval: 10s
+    {{- if eq .Values.distribution "gke/autopilot" }}
+    # GKE Autopilot doesn't allow using the secure kubelet endpoint,
+    # use the read-only endpoint instead.
+    auth_type: none
+    endpoint: ${K8S_NODE_IP}:10255
```
rmfitzpatrick, line 84: Is readonly port usage documented for autopilot? I wonder how dependable usage of this is or if we're entitled to any kind of deprecation notice.

dmitryax:
> in autopilot, is running it as user root possible?
yes, it is

docs/advanced-configuration.md
dmitryax: I started with providing another helm option for this, but not sure if `PriorityClass` is something we would like to manage in the helm chart. Also the issue doesn't happen often, and according to GKE folks it's something temporary, they are going to address it going forward OOTB. So I not sure we should introduce another help option because of this

docs/advanced-configuration.md
```diff
@@ -80,6 +80,42 @@ splunkPlatform:
...
 ```
 
+## GKE Autopilot support
+
+If you want to run Splunk OTel Collector in [Google Kubernetes Engine
+Autopilot](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-overview),
+make sure to set `distribution` setting to `gke/autopilot`:
+
+```yaml
+distribution: gke/autopilot
```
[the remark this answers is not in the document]
dmitryax: It's just something we need to come up with, I don't see any other parts of the configuration with similar meaning. My reasoning for `/` is that it works better for `type/subtype` kind of names. `autopilot` is still `gke`, so I'm thinking that `/` can better indicate this. But I'm good with using `-` instead as well

docs/advanced-configuration.md
dmitryax: This is a short term limitation. Will be addressed in January. But I'll add the same note to the Autopilot section

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: This is kubernetes problem in general, not Autopilot. Kubelet endpoins are not documented. The secure endpoint doesn't provide a way to authorize just GET requests, that's why Autopilot blocks the whole endpoint. There are plans in k8s ecosystem to deprecate the read-only endpoint and provide a granular access to secure endpoint, once it happens we can switch to it.

docs/advanced-configuration.md
dmitryax: added

rmfitzpatrick (approved):

PR #346, opened 2022-01-06
add eks/fargate distribution with 2-replica StatefulSet

rmfitzpatrick:
~~**Please do not merge until Splunk distribution has updated k8s_observer and receiver creator components**~~

These changes introduce a new `eks/fargate` distribution option that adds a receiver creator-controlled kubelet stats receiver to the cluster receiver deployment, as well as a new "cluster receiver observer" deployment to report that collector's kubelet stats metrics since the cluster receiver cannot reach its own kubelet. I chose this dual collector approach over adding the applicable observer/receiver pair to the existing gateway config because there was no clear way to limit discovery to a single gateway* replica and having multiple gateways would lead to redundant metrics (it's also not guaranteed people will use a gateway).

Please see the updated advanced configuration section for a bit more detail.

edit: from @dmitryax's suggestion* I've updated this to be a 2-replica StatefulSet with the same functionality, and pod name based init container logic for node labeling and stripping the cluster receiver on one pod.


helm-charts/splunk-otel-collector/templates/config/_otel-eks-fargate-cluster-receiver-observer.tpl
```diff
@@ -0,0 +1,205 @@
...
+      involvedObjectKind: Pod
+    - reason: FailedCreate
+      involvedObjectKind: Job
+  {{- end }}
+
+  # dynamically created kubeletstats receiver to report kubelet stats for cluster receiver "node"
+  receiver_creator/eks-fargate-cluster-receiver:
+    receivers:
+      kubeletstats:
+        rule: type == "k8s.node" && name contains "fargate" && labels["otel-eks-fargate-is-cluster-receiver-node"] == "true"
```
dmitryax, line 48: Isn't this rule checking for node labels while the label we are looking for is set on a pod?

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -42,6 +50,26 @@ receivers:
     - reason: FailedCreate
       involvedObjectKind: Job
   {{- end }}
+  {{- if eq (include "splunk-otel-collector.distribution" .) "eks/fargate" }}
+  # dynamically created kubeletstats receiver to report all Fargate "node" kubelet stats
+  # with exception of collector "node's" own since Fargate forbids connection.
```
dmitryax, line 55: Does signalfx-agent has the same issue? I see it has only one deployment.

rmfitzpatrick: Yes, this is a general fargate and eks restriction (https://github.com/signalfx/signalfx-agent/tree/main/deployments/k8s#aws-eksfargate) and the eks fargate deployment is limited: no gateway/forwarder support and no self-pod/node metrics.

helm-charts/splunk-otel-collector/templates/config/_otel-eks-fargate-cluster-receiver-observer.tpl
rmfitzpatrick: yikes, good catch.

rmfitzpatrick: @dmitryax I'm not sure of a way that helm can manage this and fargate profiles don't appear to be able add node labels either (only filter from pod labels). Other than requiring users to manually set a label w/ kubectl or similar can you think of an approach?

rmfitzpatrick: related that suggests adequate message passing though pod labels isn't possible: https://github.com/kubernetes/kubernetes/issues/40610

rmfitzpatrick: I was able to resolve this with an init container on the cluster receiver running kubectl but it requires a patch role for nodes, which is likely undesirable.

dmitryax: It's weird that they allow patching nodes in Fargate :) It does seem undesirable and likely can be disabled at some point. If we are going this route we need to make sure that it's possible to patch all the nodes to remove stale labels.

Do they allow running pods in hostNetwork? If so, maybe we can add another static receiver talking to localhost?

rmfitzpatrick: Fargate nodes are ephemeral and are only around as long as they have a pod so cleanup isn't necessary tmk. `Pods running on Fargate can't specify HostPort or HostNetwork in the pod manifest` (confirmed they won't start).

dmitryax: Does it mean that pods in Fargate are always scheduled on dedicated nodes, and the same node cannot have more than one pod? Based on your solution, looks like it's assumed.

Assuming ^ I we can consider another option:

dmitryax: Does it mean that pods in Fargate are always scheduled on dedicated nodes, and the same node cannot have more than one pod? Based on your solution, looks like it's assumed.

I'm also thinking of the another option... if fargate is distro is enabled we:
1. Change clusterReceiver to 2-replicas statefulSet.
2. Add an initContainer that don't patch the node, but prepares the discovery rule config line that can be injected in the config using some config source (from file to env var?). To get the discovery rule it checks which replica it's running on: 
 - the first replica gets the known line.
 - the second replica looks for first replica node's IP address using the k8s API and prepares the discovery rule using that IP address.

What do you think? Can it work?

dmitryax: Also if the If it's not the case that each Fargate node can have only one pod, we need to setup podAntiAffinity for the statefulSet ^

rmfitzpatrick: I'm not seeing any assurances that each fargate node is unique, but in practice this has been what I've observed. Thanks for the statefulSet suggestion, which I'm looking into now.

dmitryax: Also we need to remove `k8s_cluster` receiver from the pipeline of one of the replicas, everything else in the configs should be identical I believe. As an option we can copy configmap's config file to another local volume shared between containers and adjust it accordingly in the init container with a tool like https://mikefarah.gitbook.io/yq/.

rmfitzpatrick: @dmitryax after getting up to speed and [trying this out (ignore the readmes)](https://github.com/signalfx/splunk-otel-collector-chart/commit/663619a864149e775bf51d9aa3dbc423b7ca52bb) I don't think this approach works well:

1. Because they are launched sequentially pod-0 can only know not to watch it itself by filtering its node-0 name from the downward api in the kubelet stats receiver (and not have a cluster receiver). pod-1 knows to watch only node-0's kubelet by querying pod-0 (and have the cluster receiver) in an init container (read: static config).
2. When pod-0/node-0 goes down there is no updating the filter in pod-1 so monitoring pod-0/node-0 is permanently broken and self-monitoring doesn't work, defeating the whole point.

My proposal is now to keep the current node patching logic but have a 2-replica stateful set w/ node labelling init containers to keep the messaging passing dynamic and the collector instances linked. This way when pod-0/node-0 goes down it will reregister itself via node label so that pod-1 tracks the node change.  I think this also gives us the benefit of being able to use parallel pod management since they work independently but still have significant name ordinals (no waiting w/ intermittent cluster reporting).

dmitryax: Yes this is a problem. I think there should be a way to enforce restart of the second pod based on the liveness of the first one. I was under impression that this dependency somehow enforced not only during the first start.

First crazy idea that crosses my mind is having a regular sidecar container on the second pod that would watch for the first pod and kill itself if the first pod is unavailable. Or maybe [custom pod readiness gate](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate) maybe can help. 

I'm not saying we should try anything like this right now. I think we can proceed with what you just suggested (statefulset + patching node) and maybe later come up with another solution to avoid node pathing.

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
```diff
@@ -1,7 +1,12 @@
 {{ $clusterReceiver := fromYaml (include "splunk-otel-collector.clusterReceiver" .) }}
 {{- if and $clusterReceiver.enabled (eq (include "splunk-otel-collector.metricsEnabled" .) "true") }}
 apiVersion: apps/v1
-kind: Deployment
+{{/*
```
dmitryax, line 4: Try this to avoid the added blank lines in the rendered manifests: `{{-/*`

helm-charts/splunk-otel-collector/templates/clusterRole.yaml
```diff
@@ -86,6 +86,14 @@ rules:
   - get
   - list
   - watch
+{{- if eq (include "splunk-otel-collector.clusterReceiverNodeDiscovererInitContainerEnabled" .) "true" }}
```
dmitryax, line 89: This is probably the only place where usage of `clusterReceiverNodeDiscovererInitContainerEnabled` is needed. Both clusterReceiver template files already have `(toString $clusterReceiver.enabled) "true") (eq (toString $o11yMetricsEnabled) "true")` condition on top. So only `eq (include "splunk-otel-collector.distribution" .) "eks/fargate"` can be used there for better readability.

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
```diff
@@ -146,10 +173,45 @@ spec:
...
       {{- end }}
       {{- if $clusterReceiver.priorityClassName }}
       priorityClassName: {{ $clusterReceiver.priorityClassName }}
       {{- end }}
 {{- end }}
+
+{{- if eq (include "splunk-otel-collector.distribution" .) "eks/fargate" }}
+---
+apiVersion: v1
+kind: Service
```
dmitryax, line 207: Why service is added?

rmfitzpatrick: ```StatefulSets currently require a Headless Service to be responsible for the network identity of the Pods. You are responsible for creating this Service.```

- https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
rmfitzpatrick: Thanks, didn't realize whitespace was introduced by comments.

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
rmfitzpatrick: Without a headless service and link:

```
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1.StatefulSetSpec
error running install splunk-otel-collector-chart: exit status 1
```

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
```diff
@@ -0,0 +1,48 @@
...
+  # copy config to meet container command args
+  cp /conf/relay.yaml /splunk-messages/config.yaml
+  exit 0
+fi
+
+# Else we are the second pod (wide kubelet stats) label our node to be monitored by the first pod and disable the k8s_cluster receiver.
+# Update our config to not monitor ourselves
+echo "Labelling our fargate node to denote it hosts the cluster receiver"
+
+# download kubectl (verifying checksum)
```
dmitryax, line 20: Maybe better to use an existing kubectl image instead https://gallery.ecr.aws/bitnami/kubectl ?

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# If we are the first pod (cluster receiver), set the kubelet stats node filter to only follow labelled nodes.
+# This node label will be set by the second pod.
+if [[ "${K8S_POD_NAME}" == *-0 ]]; then
+  echo "will configure kubelet stats receiver to follow node ${FIRST_CR_REPLICA_NODE_NAME}, as well as use cluster receiver."
+  echo "export CR_KUBELET_STATS_NODE_FILTER='&& labels[\"splunk-otel-is-eks-fargate-cluster-receiver-node\"] == \"true\"'" >/splunk-messages/environ
+  cat /splunk-messages/environ
+
+  # copy config to meet container command args
+  cp /conf/relay.yaml /splunk-messages/config.yaml
```
dmitryax, line 12: If we are copying the config anyway, maybe we can adjust the config same way in both place with yq (the discovery rule and receivers list) instead of using /splunk-messages/environ. I think it should make it simpler.

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
rmfitzpatrick: This is the suggested path from aws and helps ensure we are minimally using their offerings: https://docs.aws.amazon.com/eks/latest/userguide/install-kubectl.html

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
```diff
@@ -146,6 +173,26 @@ spec:
           items:
             - key: relay
               path: relay.yaml
+      {{- if eq (include "splunk-otel-collector.distribution" .) "eks/fargate" }}
+      - name: {{ template "splunk-otel-collector.eksFargateClusterReceiverScript" . }}
```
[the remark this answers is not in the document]
rmfitzpatrick: the truncation is the real aim since the character limit is enforced and doing it everywhere seemed undesirable

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
rmfitzpatrick: Using the env vars seemed more accommodating of future changes to the underlying config, but I'm trying this out.

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
rmfitzpatrick: I wasn't able to find an official image w/ kubectl in it and opted not to use a third party for security precautions.

dmitryax: This one https://gallery.ecr.aws/bitnami/kubectl is verified and hosted by AWS, seems safe, but it's up to you

rmfitzpatrick: It's just in the public ecr, w/ a verified bitnami account. If this were for testing I would use it or similar, but not for customer adoption.

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
```diff
@@ -0,0 +1,51 @@
...
+  exit 0
+fi
+
+# Else we are the second pod (wide kubelet stats) label our node to be monitored by the first pod and disable the k8s_cluster receiver.
+# Update our config to not monitor ourselves
+echo "Labelling our fargate node to denote it hosts the cluster receiver"
+
+# download kubectl (verifying checksum)
+curl -o kubectl https://amazon-eks.s3.us-west-2.amazonaws.com/1.16.15/2020-11-02/bin/linux/amd64/kubectl
+curl -o kubectl.sha256 https://amazon-eks.s3.us-west-2.amazonaws.com/1.16.15/2020-11-02/bin/linux/amd64/kubectl.sha256
```
dmitryax, line 24: Why do we fetch the checksum instead of hardcoding it? If we think that https://amazon-eks.s3.us-west-2.amazonaws.com/1.16.15/2020-11-02/bin/linux/amd64/kubectl can be compromised, why we assume that https://amazon-eks.s3.us-west-2.amazonaws.com/1.16.15/2020-11-02/bin/linux/amd64/kubectl.sha256 cannot be?

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
```diff
@@ -0,0 +1,51 @@
...
+  cp /conf/relay.yaml /splunk-messages/config.yaml
+  exit 0
+fi
+
+# Else we are the second pod (wide kubelet stats) label our node to be monitored by the first pod and disable the k8s_cluster receiver.
+# Update our config to not monitor ourselves
+echo "Labelling our fargate node to denote it hosts the cluster receiver"
+
+# download kubectl (verifying checksum)
+curl -o kubectl https://amazon-eks.s3.us-west-2.amazonaws.com/1.16.15/2020-11-02/bin/linux/amd64/kubectl
```
dmitryax, line 23: Why do we use 1.16 version? It's not supported anymore I believe

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
dmitryax: Maybe more transparent but I think it complicates the logic. Would be nice to reduce the number of scripts and configmap mounts

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
rmfitzpatrick: This is the oldest provided version and fargate nodes were supported on 1.16 clusters. I'm not well versed in kubectl compatibility, so what version would you suggest?

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
rmfitzpatrick: good point, updated.

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
dmitryax: Kubernetes client and server minor version skew is +/-1. It doesn't mean difference beyond the skew cannot work, but it's not guaranteed to work.

Probably we can use built-in template objects to get current user kubernetes version https://helm.sh/docs/chart_template_guide/builtin_objects/ but in that case hardcoded checksum worn work :)

dmitryax: From the other hand it's just one operation. So if we can test it manually, it should be ok to use a version somewhere in between 1.16 and the latest kubernetes version 1.23, maybe 1.20

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
rmfitzpatrick: updated to use yq

helm-charts/splunk-otel-collector/scripts/lookup-eks-fargate-receiver-node.sh
rmfitzpatrick: 1.20 it is

jek-bao-choo:
@dmitryax  and @rmfitzpatrick Thank you for this support :)

rendered/manifests/eks-fargate/configmap-cluster-receiver-node-discoverer-script.yaml
```diff
@@ -0,0 +1,62 @@
...
+    
+    echo "Disabling k8s_cluster receiver for this instance"
+    # strip k8s_cluster and its pipeline
+    ./yq e 'del(.service.pipelines.metrics)' /conf/relay.yaml >/splunk-messages/config.yaml
+    ./yq e -i 'del(.receivers.k8s_cluster)' /splunk-messages/config.yaml
+    
+    # set kubelet stats to not monitor ourselves (all other kubelets)
+    echo "Ensuring k8s_observer-based kubeletstats receivers won't monitor own node to avoid Fargate network limitation."
+    ./yq e -i '.receivers.receiver_creator.receivers.kubeletstats.rule = .receivers.receiver_creator.receivers.kubeletstats.rule + " && not ( name contains \"${K8S_NODE_NAME}\" )"' /splunk-messages/config.yaml
+    ./yq e -i '.extensions.k8s_observer.observe_pods = true' /splunk-messages/config.yaml
```
dmitryax, line 62: Sorry I don't understand why this line is needed. Can you please clarify?

rmfitzpatrick: It's added so that cluster receiver config additions for the receiver creator w/ pod/port type rules will only be detected by this instance and not the self-monitoring one:

```
clusterReceiver:
  config:
    receivers:
      receiver_creator:
        receivers:
          smartagent/apache:
            config:
              type: collectd/apache
            rule: type == "port" && pod.name matches "apache"
```

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -95,6 +117,15 @@ processors:
         value: {{ .value }}
       {{- end }}
 
+  resource/k8s_cluster:
```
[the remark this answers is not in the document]
rmfitzpatrick: > But we will still have kubelet metrics from one node coming with this attribute, but removed from others.

Is this the case? the kubelet stats metrics come from receiver instances created by the receiver creator and `resource/k8s_cluster` isn't in its pipeline. This isn't node dependent, or at least not intended to be. For the receiver creator additions I mentioned earlier I don't believe we want to add this attribute.

I think this was intended just for the k8s_receiver: https://github.com/signalfx/splunk-otel-collector-chart/commit/fd6fb18e253af6b31266353fda41edb3be753d63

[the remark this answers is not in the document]
dmitryax: Oh you already split the pipeline. Sorry didn't notice that

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -171,3 +219,30 @@ service:
         {{- end }}
     {{- end }}
 {{- end }}
+
+{{- define "splunk-otel-collector.clusterReceiverInitContainers" -}}
```
dmitryax, line 223: This looks like in a wrong place. It's being used in `helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml` but defined in this config. Please move to the file where it's being used or in common. 

I'm not sure why this needs a function. May be put in the right in place? At least one condition can go away:

```
{{- if (include "splunk-otel-collector.clusterReceiverInitContainers" .) }}
```

rendered/manifests/eks-fargate/configmap-cluster-receiver-node-discoverer-script.yaml
dmitryax: I see. I think this can be confusing to users. They can look for `.extensions.k8s_observer.observe_pods` in config, assume that it's disabled by default and set it to true in their config.

Maybe we should set it to `true` in config and flip one to `false` here?

docs/advanced-configuration.md
```diff
@@ -157,6 +158,37 @@ the following line to your custom values.yaml:
...
+**NOTE:** Fluentd and Native OTel logs collection are not yet automatically configured in EKS with Fargate profiles
+
+This distribution will operate similarly to the `eks` distribution but with the following distinctions:
+
+1. The Collector agent daemonset is not applied since Fargate doesn't support daemonsets. Any desired Collector instances
+running as agents must be configured manually as sidecar containers in your custom deployments. This includes any application
+logging services like Fluentd. We recommend setting the `gateway.enabled` to `true` and configuring your instrumented
+applications to report metrics, traces, and logs to the gateway's `<installed-chart-name>-splunk-otel-collector` service address if no
+agent instances are used in your cluster. Any desired agent instances that would run as a daemonset should instead run as sidecar containers in your pods.
+2. The Collector's ClusterRole for `eks/fargate` will allow the `patch` verb on `nodes` resources for the default API groups. This is to allow
```
dmitryax, line 180: The first item includes a recommendation to users with some specific needs, but the next two are more like implementation details of some peculiarities without mentioning a reason for having them.

Maybe we can group the next two items into one and let user now why this is needed pointing to the EKS/Fargate limitation? I'm thinking of a list of two items, each of them describes a limitation and appropriate customization applied to the chart to resolve it. WDYT?

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
```diff
@@ -146,6 +167,17 @@ spec:
...
             - key: relay
               path: relay.yaml
+      {{- if eq (include "splunk-otel-collector.distribution" .) "eks/fargate" }}
+      - name: init-eks-fargate-cluster-receiver-script
+        configMap:
+          name: {{ template "splunk-otel-collector.clusterReceiverNodeDiscovererScript" . }}
+          items:
+            - key: script
+              path: init-eks-fargate-cluster-receiver.sh
+              mode: 0777
```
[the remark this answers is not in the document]
rmfitzpatrick: I had a really hard time getting things to exec, but will try it out now that the templates have settled a bit.

docs/advanced-configuration.md
rmfitzpatrick: The intention is to describe the objects and the reason for their existing considering that this is a bit unintuitive compared to the deployment. If it's not clear then I can try to make this more readable/information. How's this?

docs/advanced-configuration.md
```diff
@@ -177,17 +177,20 @@ running as agents must be configured manually as sidecar containers in your cust
...
+
+2. Since Fargate nodes use a VM boundary to prevent access to host-based resources used by other pods, pods are not able to reach their own kubelet. The cluster receiver
+for the Fargate distribution has two primary differences between regular `eks` to work around this limitation:
+    * The configured cluster receiver is deployed as a 2-replica StatefulSet instead of a Deployment and uses a
+    [Kubernetes Observer extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/observer/k8sobserver/README.md)
+    that discovers the cluster's nodes and, on the second replica, its pods. It uses this to dynamically create
+    [Kubelet Stats receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kubeletstatsreceiver/README.md)
+    instances that will report kubelet metrics for all observed Fargate nodes. The first replica will monitor the cluster with a `k8s_cluster` receiver
+    and the second will monitor all kubelets except its own (due to an EKS/Fargate networking restriction).
+    The second replica will have the underlying `k8s_cluster` receiver instance.
```
dmitryax, line 189: This line sounds confusing. Is this correct given the we have _The first replica will monitor the cluster with a `k8s_cluster` receiver_ just above?

docs/advanced-configuration.md
dmitryax: Thanks. It's much cleaner now. Added just one comment

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
rmfitzpatrick: I had originally added a value to enable/disable like was done w/ fluentd. Inlined where used.

docs/advanced-configuration.md
rmfitzpatrick: I think this was a remnant from an editing pass. Removed.

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -174,3 +222,18 @@ service:
         {{- end }}
     {{- end }}
 {{- end }}
+
+{{/*
+Pod anti-affinity to prevent eks/fargate replicas from being on same node
+*/}}
+{{- define "splunk-otel-collector.clusterReceiverPodAntiAffinity" -}}
```
rmfitzpatrick, line 229: @dmitryax is this an ok location for the anti-affinity default? I wonder if the convention is to have it in common, though it's not used in any other component.

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
```diff
@@ -52,14 +61,39 @@ spec:
...
         {{ toYaml $clusterReceiver.securityContext | nindent 8 }}
       {{- end }}
+      {{- if eq (include "splunk-otel-collector.distribution" .) "eks/fargate" }}
+      initContainers:
+        - name: cluster-receiver-node-discoverer
+          image: public.ecr.aws/amazonlinux/amazonlinux:latest
+          imagePullPolicy: IfNotPresent
+          command: ["bash", "-c", "/splunk-scripts/init-eks-fargate-cluster-receiver.sh"]
+          securityContext:
+            runAsUser: 0
```
dmitryax, line 79: I'm curios why does the script need to be run as root?

helm-charts/splunk-otel-collector/templates/configmap-gateway.yaml
```diff
@@ -1,5 +1,5 @@
 {{ $gateway := fromYaml (include "splunk-otel-collector.gateway" .) }}
-{{ if $gateway.enabled }}
+{{ if or $gateway.enabled (eq (include "splunk-otel-collector.distribution" .) "eks/fargate") }}
```
dmitryax, line 2: Looks like we don't install gateway by default, so we shouldn't install gateway config either.

docs/advanced-configuration.md
```diff
@@ -157,6 +158,39 @@ the following line to your custom values.yaml:
...
+
+**NOTE:** Fluentd and Native OTel logs collection are not yet automatically configured in EKS with Fargate profiles
+
+This distribution will operate similarly to the `eks` distribution but with the following distinctions:
+
+1. The Collector agent daemonset is not applied since Fargate doesn't support daemonsets. Any desired Collector instances
+running as agents must be configured manually as sidecar containers in your custom deployments. This includes any application
+logging services like Fluentd. We recommend setting the `gateway.enabled` to `true` and configuring your instrumented
+applications to report metrics, traces, and logs to the gateway's `<installed-chart-name>-splunk-otel-collector` service address if no
+agent instances are used in your cluster. Any desired agent instances that would run as a daemonset should instead run as sidecar containers in your pods.
```
dmitryax, line 179: The ending sounds a bit confusing, not sure if it's really needed

```suggestion
applications to report metrics, traces, and logs to the gateway's `<installed-chart-name>-splunk-otel-collector` service address.
Any desired agent instances that would run as a daemonset should instead run as sidecar containers in your pods.
```

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
rmfitzpatrick: removed

docs/advanced-configuration.md
rmfitzpatrick: If users have sidecar agents shouldn't their apps send to them? removed though to avoid further clarification.

rmfitzpatrick:
re: https://github.com/signalfx/splunk-otel-collector-chart/pull/346#discussion_r796281442

I wasn't able to get nested templates to work (`unexpected <define> in command`) and thought named templates needed to be defined in .tpl, but what I have now seems to allow a multiline string definition.

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
rmfitzpatrick: https://github.com/signalfx/splunk-otel-collector-chart/pull/346#issuecomment-1026962298

rmfitzpatrick: actually looks like helm 3.5.4 doesn't like this, so I'm not sure of the "correct" way that would satisfy your objection: https://github.com/signalfx/splunk-otel-collector-chart/runs/5023506147?.

rmfitzpatrick: err, backtick on the variable definition action resolves...

PR #356, opened 2022-01-13
Updating schema validation for imagePullSecrets as expected by chart

nitaliya:
Updated chart validation schema for `imagePullSecrets` to sync with the expectation of the chart
Previously expected validation by the schema is as mentioned below but the chart expects string instead of object type:
```
imagePullSecrets:
  - name: myregistry
```
Expected by the chart:
```
imagePullSecrets:
  - myregistry
```


dmitryax (approved):
Thanks for fixing this!

PR #357, opened 2022-01-14
Remove the quadruple dollar hack

dmitryax:
Double expansion issue was fixed with [the recent splunk-otel-collector upgrade](https://github.com/signalfx/splunk-otel-collector/pull/1099). Now OTel collector configs are expanded only once and hack with `$$$$` is not needed anymore.

+ temporary set the splunk-otel-collector docker image to the latest dev version to keep the main in a working state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Fixed
+
+- Double expansion issue splunk-otel-collector (#357). See [upgrade
+guideline](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md#0410-to-0420)
+
 ### Removed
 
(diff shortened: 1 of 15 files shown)


rockb1017:
I see that `$$$$` is changed to `$$`. 
but I think `$$` should also be changed to `$`.

dmitryax:
I see `$$` only in places like `$$.logtag` which are supposed to work as expected and should stay the same. They didn't require `$$$$` before because double expansion would be the following: `$$.logtag` -> `$.logtag` -> `$.logtag`, while double expansion of `$$$$body`: `$$body` -> `$body` -> `""`.

rockb1017:
oh hmm I see. thanks for explanation. that is why the CI pipeline worked. I was wondering.. LGTM!

rockb1017 (approved):

rockb1017:
so it will work with just one `$` for `$.*` ?

dmitryax:
> so it will work with just one $ for $.* ?

It will, but I would leave it with `$$` for consistency. Otherwise people may be confused when to use one and when two signs.

PR #361, opened 2022-01-19
Adding native otel logs collection for windows nodes

nitaliya:
Native logs support on windows k8s nodes
Ticket: [OTL-1349](https://signalfuse.atlassian.net/browse/OTL-1349)

**Changes:**
- Updating filelog path and regex to parse windows `$$attributes["file.path"]`
- Not creating `initContainers` in case of windows k8s nodes, because fluentd logging agent is not supported on a windows node
- Mounting C:/ProgramData of the host as var-lib-docker-containers in case of windows k8s nodes

**Testing outcome:**
For testing, I have used the following values.yaml input file to Otel helm chart
```
clusterName: nimesh-windows-cluster
splunkObservability:
  realm: <SPLUNK_REALM>
  accessToken: <SPLUNK_ACCESS_TOKEN>
  logsEnabled: true

image:
  otelcol:
    repository: quay.io/signalfx/splunk-otel-collector-windows-dev

logsEngine: "otel"
readinessProbe:
  initialDelaySeconds: 60
livenessProbe:
  initialDelaySeconds: 60

isWindows: true

agent:
  config:
    exporters:
      logging:
        loglevel: debug
    service:
      pipelines:
        logs:
          exporters: [ logging, splunk_hec/o11y ]
```
to generate a log I have created a sample deployment to print random numbers on the log file.

**Output:**
Console log of otel-collector agent.
Screenshot 2022-01-19 at 4 11 51 PM

Splunk Log Observer.
Screenshot 2022-01-19 at 4 09 40 PM

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -41,7 +41,6 @@ Windows support covers traces and metrics collection only, no logs will be colle
 image:
   otelcol:
     repository: quay.io/signalfx/splunk-otel-collector-windows
-logsEnabled: false
 readinessProbe:
   initialDelaySeconds: 60
 livenessProbe:
--- a/examples/README.md
+++ b/examples/README.md
@@ -35,7 +35,7 @@ This configuration will install application for windows worker node only.
 
 All windows images are available in a separate `quay.io` repository: `quay.io/signalfx/splunk-otel-collector-windows`.
 
-Windows support covers traces and metrics collection only, no logs will be collected. As of now, we don’t have fluentd-hec image for the windows worker node.
+Logs collection on Windows is only available using OTel native logs collection. fluentd log engine is not supported Because, as of now we don’t have fluentd-hec image for the windows worker node.
 
 ```yaml
 image:
--- a/examples/README.md
+++ b/examples/README.md
@@ -35,7 +35,7 @@ This configuration will install application for windows worker node only.
 
 All windows images are available in a separate `quay.io` repository: `quay.io/signalfx/splunk-otel-collector-windows`.
 
-Logs collection on Windows is only available using OTel native logs collection. fluentd log engine is not supported Because, as of now we don’t have fluentd-hec image for the windows worker node.
+Logs collection on Windows is only available using OTel native logs collection. Fluentd log engine is not supported.
 
 ```yaml
 image:
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -131,7 +131,7 @@ receivers:
   {{- if and (eq .Values.logsEngine "otel") .Values.logsCollection.containers.enabled }}
   filelog:
     {{- if .Values.isWindows }}
-    include: ["C:/var/log/pods/*/*/*.log"]
+    include: ["C:\\var\\log\\pods\\*\\*\\*.log"]
     {{- else }}
     include: ["/var/log/pods/*/*/*.log"]
     {{- end }}
@@ -140,7 +140,7 @@ receivers:
     exclude:
       {{- if .Values.logsCollection.containers.excludeAgentLogs }}
       {{- if .Values.isWindows }}
-      - C:/var/log/pods/{{ .Release.Namespace }}_{{ include "splunk-otel-collector.fullname" . }}*_*/otel-collector/*.log
+      - "C:\\var\\log\\pods\\{{ .Release.Namespace }}_{{ include "splunk-otel-collector.fullname" . }}*_*\\otel-collector\\*.log"
       {{- else }}
       - /var/log/pods/{{ .Release.Namespace }}_{{ include "splunk-otel-collector.fullname" . }}*_*/otel-collector/*.log
       {{- end }}
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -130,12 +130,20 @@ receivers:
 
   {{- if and (eq .Values.logsEngine "otel") .Values.logsCollection.containers.enabled }}
   filelog:
+    {{- if .Values.isWindows }}
+    include: ["C:/var/log/pods/*/*/*.log"]
+    {{- else }}
     include: ["/var/log/pods/*/*/*.log"]
+    {{- end }}
     # Exclude logs. The file format is
     # /var/log/pods/<namespace_name>_<pod_name>_<pod_uid>/<container_name>/<restart_count>.log
     exclude:
       {{- if .Values.logsCollection.containers.excludeAgentLogs }}
+      {{- if .Values.isWindows }}
+      - C:/var/log/pods/{{ .Release.Namespace }}_{{ include "splunk-otel-collector.fullname" . }}*_*/otel-collector/*.log
+      {{- else }}
       - /var/log/pods/{{ .Release.Namespace }}_{{ include "splunk-otel-collector.fullname" . }}*_*/otel-collector/*.log
+      {{- end }}
       {{- end }}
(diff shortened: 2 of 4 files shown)


helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -387,12 +396,21 @@ spec:
           name: {{ template "splunk-otel-collector.fullname" . }}-fluentd-json
       {{- end}}
       {{- if eq .Values.logsEngine "otel" }}
+      {{- if .Values.isWindows }}
+      - name: varlog
+        hostPath:
+          path: C:/var/log
+      - name: varlibdockercontainers
```
dmitryax, line 403: The name is misleading. Should be `programdata` perhaps.

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -387,12 +396,21 @@ spec:
...
+        hostPath:
+          path: C:/ProgramData
```
dmitryax, line 405: Can you please clarify why the whole `C:/ProgramData` has to be mounted to container? What is it being used for?

nitaliya: @dmitryax 
- When we have `containerRuntime: docker` at that time docker Daemon’s container storage location will be `C:/ProgramData/docker/containers`
- When we have `containerRuntime: containerd` at that time container storage location will be different then docker runtime and it will be something like `C:/ProgramData/containerd/`

Note: I have taken the reference of DataDog helm chart from [hear](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/_container-agent.yaml#L180-L182)

helm-charts/splunk-otel-collector/templates/daemonset.yaml
nitaliya: Thanks! renamed volume name with `programdata`

helm-charts/splunk-otel-collector/templates/daemonset.yaml
dmitryax: Interesting. Ok please add a comment line saying that `C:/ProgramData/docker/containers` and `C:/ProgramData/containerd/root` are being used. Also please verify that it doesn't break metrics collection. Otherwise looks good

nitaliya: Yes, I have checked metrics collection also it is not affecting.
Observability Chart:
Screenshot 2022-01-20 at 10 26 47 AM

dmitryax: Which metric is on the screenshot? It looks like a clusterReceiver metric. We need kubelet metrics. Please check k8s navigator https://app.signalfx.com/#/infra/k8s/map

nitaliya: yes that was hostmetrics here you can find the kubeletstats metrics also being collected.
k8s navigator:
Screenshot 2022-01-20 at 10 47 40 AM

dmitryax: Looks good, thanks for validating it. Please add a comment to C:/ProgramData volume that I mentioned above and this should be good to go

examples/README.md
```diff
@@ -41,7 +41,6 @@ Windows support covers traces and metrics collection only, no logs will be colle
 image:
   otelcol:
     repository: quay.io/signalfx/splunk-otel-collector-windows
-logsEnabled: false
```
dmitryax, line 44: I think we also have a documentation note about this limitation above, please remove it.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -130,13 +130,21 @@ receivers:
 
   {{- if and (eq .Values.logsEngine "otel") .Values.logsCollection.containers.enabled }}
   filelog:
+    {{- if .Values.isWindows }}
+    include: ["C:/var/log/pods/*/*/*.log"]
```
dmitryax, line 134: Should we have `\\` separators consistently for windows or they don't work in all the cases?

examples/README.md
nitaliya: Thanks! Removed that limitation note

examples/README.md
```diff
@@ -35,13 +35,10 @@ This configuration will install application for windows worker node only.
 
 All windows images are available in a separate `quay.io` repository: `quay.io/signalfx/splunk-otel-collector-windows`.
 
-Windows support covers traces and metrics collection only, no logs will be collected. As of now, we don’t have fluentd-hec image for the windows worker node.
```
dmitryax, line 38: Please mention that logs collection on Windows is only available using OTel native logs collection, not fluentd.

examples/README.md
```diff
@@ -35,13 +35,12 @@ This configuration will install application for windows worker node only.
...
+Logs collection on Windows is only available using OTel native logs collection. fluentd log engine is not supported Because, as of now we don’t have fluentd-hec image for the windows worker node.
```
dmitryax, line 38: Don't need to clarify why. Fluentd will go away in future anyway.

```suggestion
Logs collection on Windows is only available using OTel native logs collection. Fluentd log engine is not supported.
```

helm-charts/splunk-otel-collector/templates/daemonset.yaml
nitaliya: Added comment

examples/README.md
nitaliya: Updated documentation section

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -349,12 +349,22 @@ spec:
           readOnly: true
         {{- end }}
         {{- if and (eq (include "splunk-otel-collector.logsEnabled" $) "true") (eq .Values.logsEngine "otel") }}
+        {{- if .Values.isWindows }}
+        - name: varlog
+          mountPath: C:/var/log
+          readOnly: true
+          # From C:/ProgramData Container storage location C:/ProgramData/docker/containers and C:/ProgramData/containerd/root are being used.
```
dmitryax, line 356:
```suggestion
          # C:/ProgramData mount is needed for access to container storage in C:/ProgramData/docker/containers and C:/ProgramData/containerd/root.
```

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
nitaliya: Yes we can use `\\` in `include` section but in `exclude` we need to use `\` to so for consistency I have used `/`.
if you want then I can change separators to `\\` in include and in `exclude` we have to make following change
from
```
exclude: 
  - C:/var/log/pods/{{ .Release.Namespace }}_{{ include "splunk-otel-collector.fullname" . }}*_*/otel-collector/*.log
```
to
```
exclude: [ "C:\\var\\log\\pods\\{{ .Release.Namespace }}_{{ include "splunk-otel-collector.fullname" . }}*_*\\otel-collector\\*.log" ]
```
Should I?

dmitryax: I don't get why exclude section is different? Probably only quotes are needed but list representation can stay the same like
```
- "C:\\var\\log\\pods\\{{ .Release.Namespace }}_{{ include "splunk-otel-collector.fullname" . }}*_*\\otel-collector\\*.log" ]
```

nitaliya: Yes, you are right only quotes are needed in exclude section. 
Now I made a change to make separators consistent with `\\`

dmitryax (approved):
LGTM. Thanks!

PR #375, opened 2022-02-02
Override host and cloud attributes in OTel Collector Agent

dmitryax:
Instrumentation libraries running in container environments cannot always correctly set host attributes while OTel Collector Agent has all required access to underlying host and can be considered as a source of truth for all host and cloud metadata. This change reconfigures Resource Detection processor on the agent to override all host and cloud attributes.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,21 @@
 # Upgrade guidelines
 
+## 0.43.1 to 0.43.2
+
+[#375 Resource detection processor is configured to override all host and cloud
+attributes](https://github.com/signalfx/splunk-otel-collector-chart/pull/375)
+
+If you still want to keep the previous behavior, use the following custom
+values.yaml configuration:
+
+```yaml
+agent:
+  config:
+    processors:
+      resourcedetection:
+        override: false
+```
+
 ## 0.41.0 to 0.42.0
 
 [#357 Double expansion issue in splunk-otel-collector is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+[Upgrade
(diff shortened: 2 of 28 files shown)


UPGRADING.md
```diff
@@ -1,5 +1,21 @@
...
+
+If you still want to keep the previous behavior, use the following custom
+values.yaml configuration:
+
+```yaml
+agent:
+  config:
+    processors:
+      resourcedetection:
+        override: false
```
rmfitzpatrick, line 16: Does the resource processor's new `upsert` prevent complete rollback?

rmfitzpatrick: `--set 'agent.config.processors.resource.attributes[0].action=insert'` doesn't seem to properly merge:

```
      resource:
        attributes:
        - action: insert
```

dmitryax: Not sure I understand what do you mean by complete rollback. `upsert` in resource processor is needed to keep existing behavior of setting `k8s.cluster.name` attribute -- it should be set from `clusterName` values.yaml config option, and `k8s.cluster.name` set by resourcedetector should be ignored.

dmitryax: `k8s.cluster.name` is being set by resourcedetection processor only in some versions of GKE. Making sure this is taken care of in this rollback example, we would need to add something like:

```yaml
  config:
    processors:
      resourcedetection:
        override: false
      resource:
        attributes:
          - action: insert
            key: k8s.node.name
            value: "${K8S_NODE_NAME}"
          - action: upsert
            key: k8s.cluster.name
            value: < your cluster name from .Values.clusterName >
          < extra attribute from .Values.extraAttributes.custom >
```

I'm not sure likelihood of changed `k8s.cluster.name` warrants adding this config. We can probably just mention that `k8s.cluster.name` can be changed in GKE

dmitryax: Hmm... Actually `upsert` is good. It doesn't need to be changed for the rollback. It preserves the required `k8s.cluster.name` value set from `clusterName` before and after the change.

rmfitzpatrick (approved):

PR #384, opened 2022-02-17
Added support to collect control plane component metrics; etcd

jvoravong:
Description:
This PR adds support for collecting metrics from the k8s etcd.

Testing:
I tested and verified these changes with Kubernetes and Openshift.


docs/advanced-configuration.md
```diff
@@ -226,11 +229,57 @@ The default configurations for the control plane receivers can be found in
...
 * [smartagent/kubernetes-scheduler](https://docs.splunk.com/Observability/gdi/kubernetes-scheduler/kubernetes-scheduler.html)
 
+### Setting up Etcd metrics
+
+The etcd metrics cannot be collected out of box because etcd requires TLS authentication for communication.
+
+The easiest way to set up the TLS authentication for etcd metrics is to retrieve the client certificate and key from an
+etcd pod and directly insert them into the values.yaml. The helm chart will set up the rest. The helm chart will add
+the client certificate and key to the otel-collector agent's kubernetes secret and then configure the etcd receiver to
+use them. The etcd TLS client certificate and key play a critical role in the security of the cluster, handle them with
```
rmfitzpatrick, line 245: > The etcd TLS client certificate and key play a critical role in the security of the cluster, handle them with
care and avoid storing them in unsecured locations.

This being said, having them in plaintext in a values file seems less than ideal as a requirement on our part. Do you know if there's a standard way of reading from a file at installation? Or would it make sense to have them create the secrets themselves and specify their name so we can use them in a mount?

dmitryax: I think it's ok to have an option to provide the certificates in values.yaml. We do the same for all other secrets. It's just being used to create k8s secrets. It doesn't need to be persisted. At the same time we should provide an option to use k8s secrets created manually as we do for other secrets in this chart

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Added support to collect control plane component metrics; etcd (#TBD)
 - Added support to collect control plane component metrics; controller-manager, coredns, proxy, scheduler (#383)
```
dmitryax, line 10:
```suggestion
- Control plane metrics support: controller-manager, coredns, etcd, proxy, scheduler (#383, #384)
```

docs/advanced-configuration.md
```diff
@@ -194,7 +194,10 @@ for the Fargate distribution has two primary differences between regular `eks` t
 ## Control Plane metrics
 
 By setting `agent.controlPlaneEnabled=true` the helm chart will set up the otel-collector agent to collect metrics from
-the control plane.
+the control plane. Most metrics can be collected from the control plan with no extra configuration, however, extra
```
dmitryax, line 197:
```suggestion
the control plane. Most metrics can be collected from the control plane with no extra configuration, however, extra
```

docs/advanced-configuration.md
dmitryax: Also etcd certificated must be present on master nodes, right? So I believe users can just mount them to the container if they want to. Can we test that approach and document as an option?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -212,9 +212,28 @@ agent:
   enabled: true
 
   # This Flag enables k8s control plane metric collection.
-  # Details about control plane monitoring and related configurations are located in docs/advanced-configuration.md
+  # Details about control plane monitoring and related configurations are
+  # located in docs/advanced-configuration.md
   controlPlaneEnabled: true
```
dmitryax, line 217: The name sounds a bit confusing. k8s cannot work without control plane, so it's always enabled. Can we rename it?

What about the following?

```yaml
controlPlaneMetrics:
  coredns:
    enabled: true
  controller-manager:
    enabled: true
  ...
  etcd:
    enabled: false
    clientCert:
    clientKey:
    caFile:
```

etcd certificates must be optional in case user wants to set them up manually. we should rely on `controlPlaneMetrics.etcd.enabled` instead

helm-charts/splunk-otel-collector/templates/secret.yaml
```diff
@@ -26,4 +26,13 @@ data:
...
   {{- end }}
+  {{- with .Values.agent.etcd.caCert }}
+  etcd_ca_cert: {{ . | b64enc }}
+  {{- end }}
+  {{- with .Values.agent.etcd.clientCert }}
+  etcd_client_cert: {{ . | b64enc }}
+  {{- end }}
+  {{- with .Values.agent.etcd.clientKey }}
+  etcd_client_key: {{ . | b64enc }}
+  {{- end }}
```
dmitryax, line 37: I'm not sure we should mix Splunk related secrets with etcd certificates. Maybe better to create another k8s secret just for etcd?

docs/advanced-configuration.md
dmitryax: Also etcd certificated must be present on master nodes, right? So I believe users can just mount them to the container if they want to. Can we test that approach and document as an option?

dmitryax: That's fine. We need to test and document this as an option asking user to install the helm chart 2 times with appropriate nodeSelectors/tolerations

CHANGELOG.md
jvoravong: Done

docs/advanced-configuration.md
jvoravong: Done

helm-charts/splunk-otel-collector/templates/secret.yaml
jvoravong: I agree. Done.

docs/advanced-configuration.md
jvoravong: We have proposed 3 ways to handle etcd secrets here.
1. Have the user supply the etcd cert values to the helm chart in values.yaml or with -set. Then have the helm chart handle creating the secret and using the secret.
2. Have the users supply a supply a secret name that contains the etcd certs. Then have helm use the supplied secret.
3. Have the user deploy two instances of the helm chart; one for master nodes and one for worker nodes. The helm chart uses the etcd certs that are present on master nodes and doesn't use any kubernetes secrets.

All 3 of these methods sound acceptable to me. Let me know if you think otherwise.

dmitryax: Sounds good to me. Thanks

helm-charts/splunk-otel-collector/values.yaml
jvoravong: This change has been made.

docs/advanced-configuration.md
jvoravong: Update:
Adding the 3rd method mentioned above is a lot more work than I anticipated. To keep the scope of this ticket smaller, I am only going to support the 1st and 2nd methods mentioned above. We can add the 3rd method in another ticket at a later time.

jvoravong:
Since we have a splunk and etcd secret now, I refactored (secret->secret-splunk) the original secret file names.

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -289,7 +289,55 @@
...
+            "coredns": {
+              "enabled": {
+                "type": "boolean"
+              }
+            },
+            "etcd": {
+              "enabled": {
+                "type": "boolean"
+              },
+              "type": "object",
```
dmitryax, line 317: `additionalProperties` and secret related keys are missed.

Also it would be great if you can add validation secret.create=true or secret.name!=""

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -211,9 +211,48 @@ extraAttributes:
...
+    etcd:
+      # Specifies whether to collect etcd metrics.
+      # For set up etcd metrics details see: docs/advanced-configuration.md#setting-up-etcd-metrics
+      enabled: false
+      secret:
+        # If name is empty the default name is used.
+        name: ""
+        # Option for creating a new secret or using an existing one.
+        create: false
+        # Used when secret.create=true creates a se. The PEM-format CA certificate for the etcd client.
```
dmitryax, line 235:
```suggestion
        # Used when secret.create=false. The PEM-format CA certificate for the etcd client.
```

and the same for others

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -211,9 +211,48 @@ extraAttributes:
...
+        # NOTE: The content of the certificate itself should be used here, not the
+        #       file path. The certificate will be stored as a secret in kubernetes.
+        clientCert: ""
+        # Used when secret.create=true. The private key for the etcd client.
+        # NOTE: The content of the key itself should be used here, not the file path.
+        #       The key will be stored as a secret in kubernetes.
+        clientKey: ""
+        # Optional. Used when secret.create=true. The PEM-format CA certificate file.
```
dmitryax, line 243:
```suggestion
        # Optional. Used when secret.create=true and skipVerify=false. The PEM-format CA certificate file.
```

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -146,6 +146,17 @@ Define name for the Secret
 {{- end -}}
 {{- end -}}
 
+{{/*
+Define name for the etcd Secret
+*/}}
+{{- define "splunk-otel-collector.secret_etcd" -}}
```
dmitryax, line 152: Nit: just for consistency with the established naming convention in helm

```suggestion
{{- define "splunk-otel-collector.etcdSecret" -}}
```

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -289,7 +289,55 @@
           "type": "boolean"
         },
         "controlPlaneEnabled": {
-          "type": "boolean"
+          "type": "boolean",
+          "deprecated": true
+        },
+        "controlPlaneMetrics": {
+          "type": "object",
```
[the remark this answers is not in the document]
jvoravong: Good call, done.

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -289,7 +289,55 @@
...
+          "properties:": {
+            "apiserver": {
```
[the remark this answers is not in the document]
jvoravong: Done

helm-charts/splunk-otel-collector/values.yaml
jvoravong: Done

helm-charts/splunk-otel-collector/templates/_helpers.tpl
jvoravong: Done

helm-charts/splunk-otel-collector/values.schema.json
jvoravong: Updated so all the properties are present. Added in some validations.

helm-charts/splunk-otel-collector/values.yaml
jvoravong: I think we want to keep this "secret.create=true", but I did get rid of everything else. I also added more comments for context. Let me know if you want this to still change.

dmitryax: Makes sense. I just saw create=false in the values.yaml before

dmitryax (approved):
LGTM

PR #388, opened 2022-02-24
set profilingEnabled to default false

emaderer:
Until we add filtering for profiling in logs pipeline


dmitryax:
Please add a change log entry marked as a breaking change :)

dmitryax:
also run `make render`

dmitryax (approved):

PR #390, opened 2022-02-28
Add splunk-otel-collector secret to cluster receiver

tongpu:
Prior to this fix the cluster receiver wouldn't start because it was looking for the CA, but as it was not mounted it failed to start the process.

Would there be any other places where this might need to be added?

### DIFF
--- a/helm-charts/splunk-otel-collector/Chart.yaml
+++ b/helm-charts/splunk-otel-collector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: splunk-otel-collector
-version: 0.43.4
+version: 0.43.5
 appVersion: 0.43.0
 description: Splunk OpenTelemetry Collector for Kubernetes
 icon: https://github.com/signalfx/splunk-otel-collector-chart/tree/main/splunk.png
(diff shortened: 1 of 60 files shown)


helm-charts/splunk-otel-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: splunk-otel-collector
-version: 0.43.4
+version: 0.43.5
```
dmitryax, line 3: If you want to release it right away, please add a section to CHANGELOG and run `make render`

dmitryax (commented):
Thanks for fixing it!

dmitryax:
> Would there be any other places where this might need to be added?

I think it's also missed in gateway-deployment.yaml

helm-charts/splunk-otel-collector/Chart.yaml
dmitryax: Feel free to keep the current version. we will release it after

tongpu:
> > Would there be any other places where this might need to be added?
> 
> I think it's also missed in gateway-deployment.yaml

I've added it to gateway-deployment.yaml as well and ran `make render` to render the static manifests. Should I updated the changelog as well?

dmitryax (approved):

PR #391, opened 2022-02-28
Adding a maxUnavailable variable for daemonSet.RollingUpdate

legoGoat:
We've recently realised that rolling updates are very time-consuming (if applicable, can also cause timeouts) when the number of pods grows significantly and [the default maxUnavailable is set to 1](https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/#performing-a-rolling-update).
This change provides an easy way to adjust the number if needed.


dmitryax (approved):

dmitryax:
@legoGoat thanks for the contribution!

PR #394, opened 2022-03-02
Release 0.44.0

jvoravong:
Release 0.44.0


dmitryax:
As a rule of thumb it's usually better to not include a version upgrade in the same PR with a feature added on top of something else that is not released yet.

emaderer (approved):

PR #410, opened 2022-03-14
Add support for otelcol feature gates to the agent, clusterReceiver, and gateway

jvoravong:
Description:
We have a bug fix coming out in v0.47.0 that uses a feature gate. Now seems like a good time to add feature gate support to the collector chart.

Testing:
Tested these changes manually, I couldn't come up with a good way to unit or e2e test these changes.

### DIFF
--- a/docs/advanced-configuration.md
+++ b/docs/advanced-configuration.md
@@ -524,11 +524,10 @@ details see the
 
 Helm Install Example:
 ```bash
-# Due to syntax limitations, when using "," characters between feature gate identifiers you must escape them with "\\".
-helm install {name} --set agent.featureGates=+feature1 --set=clusterReceiver.featureGates=feature1\\,-feature2 {other_flags}
+helm install {name} --set agent.featureGates=+feature1 --set clusterReceiver.featureGates=feature2 --set gateway.featureGates=-feature2 {other_flags}
 ```
-Would result in the agent having feature1 enabled, the clusterReceiver having feature1 enabled, and the clusterReceiver
-having feature2 disabled.
+Would result in the agent having feature1 enabled, the clusterReceiver having feature2 enabled, and the gateway having
+feature2 disabled.
 
 ### Highlighted feature gates
 - receiver.k8sclusterreceiver.reportCpuMetricsAsDouble
--- a/docs/advanced-configuration.md
+++ b/docs/advanced-configuration.md
@@ -515,6 +515,37 @@ autodetect:
   prometheus: true
 ```
 
+## Using feature gates
+Enable or disable features of the otel-collector agent, clusterReceiver, and gateway (respectively) using feature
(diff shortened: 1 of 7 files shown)


docs/advanced-configuration.md
```diff
@@ -515,6 +515,37 @@ autodetect:
...
+## Using feature gates
+Enable or disable features of the otel-collector agent, clusterReceiver, and gateway (respectively) using feature
+gates. Use the agent.featureGates, clusterReceiver.featureGates, and gateway.featureGates configs to enable or disable
+features, these configs will be used to populate the otelcol binary startup argument "--feature-gates". For more
+details see the
+[feature gate documentation](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/featuregate/README.md).
+
+Helm Install Example:
+```bash
+helm install {name} --set agent.featureGates=feature1,clusterReceiver.featureGates=+feature2,gateway.featureGates=-feature3 {other_flags}
```
dmitryax, line 527: Hmm.. the feature gates are also separated by commas, so maybe split it to several --set args like

```suggestion
helm install {name} --set agent.featureGates=+feature1,-feature2 --set=clusterReceiver.featureGates=+feature1,-feature2 {other_flags}
```

jvoravong: Good point. We do have to escape commas (special character), so added that too.

docs/advanced-configuration.md
```diff
@@ -515,6 +515,38 @@ autodetect:
...
+Enable or disable features of the otel-collector agent, clusterReceiver, and gateway (respectively) using feature
+gates. Use the agent.featureGates, clusterReceiver.featureGates, and gateway.featureGates configs to enable or disable
+features, these configs will be used to populate the otelcol binary startup argument "--feature-gates". For more
+details see the
+[feature gate documentation](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/featuregate/README.md).
+
+Helm Install Example:
+```bash
+# Due to syntax limitations, when using "," characters between feature gate identifiers you must escape them with "\\".
+helm install {name} --set agent.featureGates=+feature1 --set=clusterReceiver.featureGates=feature1\\,-feature2 {other_flags}
```
dmitryax, line 528: Does `--set="clusterReceiver.featureGates=feature1,-feature2"` work instead of escaping?

jvoravong: If we quote the whole value we still have to use a "\" to escape any "," (it's a helm syntax compilation thing). I'm seeing what else we can do here. We might be able to use a helper function here that accepts strings or lists (which might be cleaner).

dmitryax: Ok, it's very unlikely that users need more than one feature gate at a time. Let's simplify it to one feature gate per each --set. User can always define them in yaml if needed

jvoravong: Done

dmitryax (approved):

PR #416, opened 2022-03-22
Clarify ports condition in service template

SimonAlling:
This PR clarifies the condition in two ways:

  * It replaces `eq true …` with `or …`, which should be easier to
    understand for most readers.

  * It moves the four disjuncts to their own lines, making it easier for
    a human to parse the condition as a whole and its subexpressions.

💡 `git show --color-words` is useful.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/service.yaml
+++ b/helm-charts/splunk-otel-collector/templates/service.yaml
@@ -24,7 +24,7 @@ spec:
   {{ $tracesEnabled    := and (eq (include "splunk-otel-collector.tracesEnabled"    $) "true") (has "traces"    $port.enabled_for) }}
   {{ $logsEnabled      := and (eq (include "splunk-otel-collector.logsEnabled"      $) "true") (has "logs"      $port.enabled_for) }}
   {{ $profilingEnabled := and (eq (include "splunk-otel-collector.profilingEnabled" $) "true") (has "profiling" $port.enabled_for) }}
-  {{- if or $metricsEnabled $tracesEnabled $logsEnabled $profilingEnabled /* Helm 3.5 (due to Go 1.15) doesn't support newlines in actions. 👉 https://github.com/signalfx/splunk-otel-collector-chart/pull/416 */ }}
+  {{- if or $metricsEnabled $tracesEnabled $logsEnabled $profilingEnabled /* Helm 3.5 (due to Go 1.15) doesn't support newlines in actions. 👉 https://github.com/signalfx/splunk-otel-collector-chart/pull/416#issuecomment-1076160057 */ }}
   - name: {{ $key }}
     port: {{ $port.containerPort }}
     targetPort: {{ $key }}
--- a/helm-charts/splunk-otel-collector/templates/service.yaml
+++ b/helm-charts/splunk-otel-collector/templates/service.yaml
@@ -20,13 +20,11 @@ spec:
   type: {{ .Values.service.type }}
   ports:
   {{- range $key, $port := $gateway.ports }}
-  {{- if
-    or
-      (and (eq (include "splunk-otel-collector.metricsEnabled" $) "true") (has "metrics" $port.enabled_for))
-      (and (eq (include "splunk-otel-collector.tracesEnabled" $) "true") (has "traces" $port.enabled_for))
-      (and (eq (include "splunk-otel-collector.logsEnabled" $) "true") (has "logs" $port.enabled_for))
-      (and (eq (include "splunk-otel-collector.profilingEnabled" $) "true") (has "profiling" $port.enabled_for))
-  }}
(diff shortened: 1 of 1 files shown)


SimonAlling:
I ran some basic tests in [sprig-repl](https://github.com/n0madic/sprig-repl) to verify that the `eq true …` → `or …` change doesn't affect the semantics:

```
> eq (or false false false) (eq true false false false)
true
> eq (or false false true) (eq true false false true)
true
> eq (or false true false) (eq true false true false)
true
> eq (or false true true) (eq true false true true)
true
> eq (or true false false) (eq true true false false)
true
> eq (or true false true) (eq true true false true)
true
> eq (or true true false) (eq true true true false)
true
> eq (or true true true) (eq true true true true)
true
```

dmitryax:
Thanks @SimonAlling. I like how readability is improved here. 

Rendering is failing with the following message:
```
Error: parse error at (splunk-otel-collector/templates/service.yaml:23): unexpected unclosed action in if
```
PTAL

SimonAlling:
> Rendering is failing with the following message:
> 
> ```
> Error: parse error at (splunk-otel-collector/templates/service.yaml:23): unexpected unclosed action in if
> ```

Yup, noticed. Spontaneous guess: line breaks are not supported or require special treatment.
 
> PTAL

Will do!

SimonAlling:
Interestingly, templating works fine with Helm 3.6+, but not with 3.5.4 (which is used in CI). Specifically:

| `helm version` output                                            | `helm template` outcome |
|------------------------------------------------------------------|-------------------------|
| `version.BuildInfo{Version:"v3.5.4", […] GoVersion:"go1.15.11"}` | ❌                      |
| `version.BuildInfo{Version:"v3.6.3", […] GoVersion:"go1.16.5"}`  | ✅                      |
| `version.BuildInfo{Version:"v3.7.0", […] GoVersion:"go1.16.8"}`  | ✅                      |

Note the Go version in each case. A quick glance at [the Go 1.16 release notes](https://go.dev/doc/go1.16#text/template) shines light on the matter:

> Newlines characters are now allowed inside action delimiters, permitting actions to span multiple lines.

So if upgrading to Helm 3.6+ in CI is OK, then we could just do that before merging this PR. Note that [v3.6.3 is already used in one of the CI workflows](https://github.com/signalfx/splunk-otel-collector-chart/blob/b673ed480ec10d6041a5c9b710c7df331883adcb/.github/workflows/lint-test.yaml#L17). Let me know what you think.

dmitryax:
> Otherwise, there is always this Helm 3.5-compatible solution:

It looks good to me!

SimonAlling:
> It looks good to me!

Do you prefer the Helm 3.5-compatible solution over upgrading to Helm 3.6+ in CI (as is already the case in one of the workflows)?

SimonAlling:
> I would prefer `Helm 3.5-compatible solution` for now

I have pushed it, @dmitryax! :slightly_smiling_face:

dmitryax:
Looks like CI is failing because of the comments

helm-charts/splunk-otel-collector/templates/service.yaml
```diff
@@ -20,7 +20,11 @@ spec:
...
+  {{ $metricsEnabled   := and (eq (include "splunk-otel-collector.metricsEnabled"   $) "true") (has "metrics"   $port.enabled_for) }}
+  {{ $tracesEnabled    := and (eq (include "splunk-otel-collector.tracesEnabled"    $) "true") (has "traces"    $port.enabled_for) }}
+  {{ $logsEnabled      := and (eq (include "splunk-otel-collector.logsEnabled"      $) "true") (has "logs"      $port.enabled_for) }}
+  {{ $profilingEnabled := and (eq (include "splunk-otel-collector.profilingEnabled" $) "true") (has "profiling" $port.enabled_for) }}
+  {{- if or $metricsEnabled $tracesEnabled $logsEnabled $profilingEnabled /* Helm 3.5 (due to Go 1.15) doesn't support newlines in actions. 👉 https://github.com/signalfx/splunk-otel-collector-chart/pull/416#issuecomment-1076160057 */ }}
```
dmitryax, line 27:
```suggestion
  {{- if or $metricsEnabled $tracesEnabled $logsEnabled $profilingEnabled }}
```

SimonAlling:
> Looks like CI is failing because of the comments

Ah, lol. I have pushed a commit that brings back the comment in a way that should pass CI. WDYT?

dmitryax:
> Ah, lol. I have pushed a commit that brings back the comment in a way that should pass CI. WDYT?

I think it's already in a pretty good shape, template variables are being used in many places in this helm chart, so I would leave it without any comments.

SimonAlling:
> I think it's already in a pretty good shape, template variables are being used in many places in this helm chart, so I would leave it without any comments.

OK! I have reverted the commit that brought back the comment.

dmitryax (approved):
@SimonAlling thanks for the contribution!

PR #421, opened 2022-03-23
reorder custom operators

rockb1017:
when user specify both multiline config and provide custom operators, custom operators gets ignored because after multiline config, logs are directly passed to `clean-up-log-record` operator.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -7,6 +7,7 @@ processes](https://github.com/signalfx/splunk-otel-collector-chart/pull/421)
 
 If you are referencing fields in log body (ie, `$$.pod_name`), change them
 corresponding attributes/resource (ie, `$$resource["k8s.pod.name"]`).
+If you are referencing log message itself by `$$.log`, you need to change it to `$$`.
 
 ## 0.43.1 to 0.43.2
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -5,8 +5,9 @@
 [#421 extra operators for container logs are inserted after all default data
 processes](https://github.com/signalfx/splunk-otel-collector-chart/pull/421)
 
-If you are referencing fields in log body (ie, `$$.pod_name`), change them
-corresponding attributes/resource (ie, `$$resource["k8s.pod.name"]`).
+If you have any custom log processing operators defined in `logsCollection.containers.extraOperators`
+configuration, change any references to log body fields (ie, `$$.pod_name`) in the operators to
+corresponding resource attributes (ie, `$$resource["k8s.pod.name"]`).
 If you are referencing log message itself by `$$.log`, you need to change it to `$$`.
 
 ## 0.43.1 to 0.43.2
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,13 @@
 # Upgrade guidelines
 
+## 0.46.0 to 0.46.1
+
+[#421 extra operators for container logs are inserted after all default data
+processes](https://github.com/signalfx/splunk-otel-collector-chart/pull/421)
+
+If you are referencing fields in log body (ie, `$$.pod_name`), change them
+corresponding attributes/resource (ie, `$$resource["k8s.pod.name"]`).
+
 ## 0.43.1 to 0.43.2
 
 [#375 Resource detection processor is configured to override all host and cloud
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -358,9 +358,6 @@ receivers:
         is_first_entry: '($$.log) matches {{ .firstEntryRegex | quote }}'
       {{- end }}
       {{- end }}
-      {{- with .Values.logsCollection.containers.extraOperators }}
-      {{ . | toYaml | nindent 6 }}
-      {{- end }}
       # Clean up log record
       - type: restructure
         id: clean-up-log-record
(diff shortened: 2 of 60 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -342,6 +342,9 @@ receivers:
           com.splunk.sourcetype: 'EXPR("kube:container:"+$$.container_name)'
         attributes:
           log.iostream: 'EXPR($$.stream)'
+      {{- with .Values.logsCollection.containers.extraOperators }}
+      {{ . | toYaml | nindent 6 }}
+      {{- end }}
```
dmitryax, line 347: Why don't put it after `clean-up-log-record`? I think it's better to apply custom operators on a record that user gets at the end of the default transformation. We don't lose any data after `clean-up-log-record`, right?

rockb1017: that makes sense. updated.

dmitryax: Thanks. I believe this is a breaking change, right? User's `extraOperators` configs have to be changed to adopt to the new log structure that we have at this point or they will work as before?

rockb1017: Yes. 
Since we put it after clean up log record, it is breaking change if they were using `$$.pod_name` instead of `$$resource["k8s.pod.name"]`

it is not a breaking change if we just put it before multiline config as my original commit.

dmitryax: I think it's still much cleaner to provide a way to manipulate records after all the default transformation is done on the whole recombined logs instead of on each broken line. If we keep it before multiline, user can break the further recombine operator.

Can you please mark this as a breaking change in Chandelog and add the steps on how to upgrade in https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md?

UPGRADING.md
```diff
@@ -1,5 +1,13 @@
 # Upgrade guidelines
 
+## 0.46.0 to 0.46.1
+
+[#421 extra operators for container logs are inserted after all default data
+processes](https://github.com/signalfx/splunk-otel-collector-chart/pull/421)
+
+If you are referencing fields in log body (ie, `$$.pod_name`), change them
+corresponding attributes/resource (ie, `$$resource["k8s.pod.name"]`).
```
dmitryax, line 9: Let's mention `logsCollection.containers.extraOperators` config
```suggestion
If you have any custom log processing operators defined in `logsCollection.containers.extraOperators`
configuration, change any references to log body fields (ie, `$$.pod_name`) in the operators to
corresponding resource attributes (ie, `$$resource["k8s.pod.name"]`).
```

Another question, do body fields `$$.pod_name` really need to be replaces with resource attributes `$$resource["k8s.pod.name"]`, not log attributes `$$attributes["k8s.pod.name"]`?

rockb1017:
```
      # Move out attributes to Attributes
      - type: metadata
        resource:
          k8s.pod.uid: 'EXPR($$.uid)'
          k8s.container.restart_count: 'EXPR($$.restart_count)'
          k8s.container.name: 'EXPR($$.container_name)'
          k8s.namespace.name: 'EXPR($$.namespace)'
          k8s.pod.name: 'EXPR($$.pod_name)'
          com.splunk.sourcetype: 'EXPR("kube:container:"+$$.container_name)'
        attributes:
          log.iostream: 'EXPR($$.stream)'
```

This is what we have. so it is added to resource.

rockb1017: we can put them all under attributes. it will be the same. I don't think there will be any improvements? can i ask why you want to use log attributes?

dmitryax: It's not about what we _want_ to use. We have this operator `clean-up-log-record` that translates log representation from one format to another. We charge expected input for `logsCollection.containers.extraOperators` and the custom operators will work with different format. We need to provide users explicit guidance on what to change in their `extraOperators` config to keep the same behavior.

You are saying that they need to replace `$$.pod_name` to `$$resource["k8s.pod.name"]`, but I don't think it's correct. As far as I see `clean-up-log-record` moves everything from `log` attribute to the root of the entry so I believe instructions must be different something like replacing `$$.log` with `$$`. Please correct me if I'm wrong.

rockb1017: $$.pod_name to $$resource["k8s.pod.name"] 
^ this was just one example. 
all parsed metadata is added as either resource/attributes. so users should grep it from resource/attributes instead of trying to get it from log body like `$$.pod_name`. 
I agree that I need to add another line for `$$.log`.

UPGRADING.md
```diff
@@ -1,5 +1,15 @@
...
 
+## 0.46.0 to 0.46.1
+
+[#421 extra operators for container logs are inserted after all default data
+processes](https://github.com/signalfx/splunk-otel-collector-chart/pull/421)
+
+If you have any custom log processing operators defined in `logsCollection.containers.extraOperators`
+configuration, change any references to log body fields (ie, `$$.pod_name`) in the operators to
+corresponding resource attributes (ie, `$$resource["k8s.pod.name"]`).
+If you are referencing log message itself by `$$.log`, you need to change it to `$$`.
```
dmitryax, line 11: I still don't think the first part is correct recommendation for this change

```suggestion
referencing log message by `$$.log`, you need to change it to `$$`.
```

UPGRADING.md
rockb1017: @dmitryax Could you help me understand what seems incorrect to you?

dmitryax: > $$.pod_name to $$resource["k8s.pod.name"]
> ^ this was just one example.
> all parsed metadata is added as either resource/attributes. so users should grep it from resource/attributes instead of trying to get it from log body like $$.pod_name.

This is correct, but user has to do this even before this change. `$$.pod_name` gives nothing within `extraOperators` before and after this change, right? If so, why should it be added to the upgrade guidelines?

rockb1017: before the upgrade, it would certainly yield pod name. since it comes before `clean-up-log-record`, body would have that field.

atoulme:
Closed as inactive. Feel free to reopen if this PR is still being worked on.

PR #435, opened 2022-04-13
Use add/move operators instead of metadata/restructure

harshit-splunk:
metadata and restructure operators are removed from opentelemetry-log-collection v0.28.0 (https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/CHANGELOG.md#breaking-changes-1)


jvoravong (approved):
From my understanding and with these changes, our filelog receiver will work with opentelemetry-log-collection version 0.27.2. This is good first step, we still have to make other updates in other PRs to have the filelog receiver support opentelemetry-log-collection versions 0.28.0 and 0.29.0.

These changes look good to me.

PR #444, opened 2022-05-02
Make sure that logs or profiling data is sent only when it's enabled

dmitryax:
This change helps to avoid situations when profiling or logs data can be sent to the backend even if it's disabled

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -187,6 +187,7 @@ splunk_hec/platform_logs:
   timeout: {{ .Values.splunkPlatform.timeout }}
   splunk_app_name: {{ .Chart.Name }}
   splunk_app_version: {{ .Chart.Version }}
+  profiling_data_enabled: false
   tls:
     insecure_skip_verify: {{ .Values.splunkPlatform.insecureSkipVerify }}
     {{- if .Values.splunkPlatform.clientCert }}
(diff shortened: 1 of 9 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -567,6 +567,8 @@ exporters:
   splunk_hec/o11y:
     endpoint: {{ include "splunk-otel-collector.o11yIngestUrl" . }}/v1/log
```
rmfitzpatrick, line 568: Unrelated, but do you know why we don't use a hec receiver in the gateway? Is it because there's no access token passthrough?

rmfitzpatrick (approved):

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -187,6 +187,7 @@ splunk_hec/platform_logs:
   timeout: {{ .Values.splunkPlatform.timeout }}
   splunk_app_name: {{ .Chart.Name }}
   splunk_app_version: {{ .Chart.Version }}
+  profiling_data_enabled: false
```
rmfitzpatrick, line 190: This would prevent users from falling back to an older collector version if needed for some reason. May be worth planning a ~feature set helper in the future to make this easier.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: Do you mean using hec between agent and gateway? We use OTLP for performance reasons

helm-charts/splunk-otel-collector/templates/config/_common.tpl
dmitryax: We can add a condition based `{{ .Chart.AppVersion }}` We haven't been doing that so far, but maybe makes sense here

dmitryax: Hmm. actually I don't think `.Chart.AppVersion` can be set by the end user

dmitryax: Probably we can parse `image.otelcol.tag`, but not sure it's worth the added complication. I'd rather keep it as is and add such backward compatibility if there is an ask for that

PR #454, opened 2022-05-27
Make Daemonset's upgrade strategy configurable

ikarlashov:
To avoid a corner-case when new pods can't be scheduled on worker node due to the lack of resources, we're using **OnDelete** updateStrategy. 

This PR makes splunk-otel-collector Daemonset's updateStrategy configurable with keeping default value as "RollingUpdate".


helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -1,1410 +1,1557 @@
...
-                  ]
-                },
-                "memory": {
-                  "type": "string"
+                },
+                "securityContext": {
+                    "type": "object"
+                },
+                "updateStrategy": {
+                    "type": "string"
```
rmfitzpatrick, line 316:
```suggestion
                    "type": "string",
                    "enum": [
                        "OnDelete",
                        "RollingUpdate"
                    ]
```

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -1,1410 +1,1557 @@
 {
-  "$schema": "http://json-schema.org/schema#",
```
rmfitzpatrick, line 2: Do you mind providing only the minimum required changes here? It looks like some descriptions are getting removed.

dmitryax (approved):

dmitryax:
@ikarlashov thanks for the contribution!

PR #460, opened 2022-06-03
Add retry_on_failure/sending_queue config to values.yaml

harshit-splunk:
Expose the batch, retry_on_failure and sending_queue configurations so they are configurable within the values.yaml. 
Users would still use default values, but this way users have full control.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Add retry_on_failure/sending_queue config to values.yaml (#460)
+- Add `splunkPlatform.retryOnFailure` and `splunkPlatform.sendingQueue` config options to values.yaml (#460)
 
 ## [0.52.0] - 2022-06-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- Add retry_on_failure/sending_queue/batch config to values.yaml (#460)
+- Add retry_on_failure/sending_queue config to values.yaml (#460)
 
 ## [0.51.0] - 2022-05-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Added
+
+- Add retry_on_failure/sending_queue/batch config to values.yaml (#460)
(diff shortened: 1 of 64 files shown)


harshit-splunk:
As of now, I have added retry_on_failure and sending_queue to splunkPlatform only.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -161,6 +182,22 @@ autodetect:
   prometheus: false
   istio: false
 
+################################################################################
+# Optional: batch processor config
+################################################################################
+
+# Refer https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor
+# for detailed examples
+batchConfig:
```
rmfitzpatrick, line 191: I'm not sure I see the benefit for having a shared value for agent, gateway, and cluster receiver processor use. These can already be set w/ `<agent|gateway|clusterReceiver>.config.processors.batch.<field>` right?

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -226,6 +230,10 @@ splunk_hec/platform_metrics:
     {{- if .Values.splunkPlatform.caFile }}
     ca_file: /otel/etc/splunk_platform_hec_ca_file
     {{- end }}
+  retry_on_failure:
+    {{- toYaml .Values.splunkPlatform.retry_on_failure | nindent 4 }}
```
rmfitzpatrick, line 234: the previous convention appears to be camel case so I'd suggest staying with that. We may want to treat all fields individually like is done w/ tls above to avoid a direct map injection method?

helm-charts/splunk-otel-collector/values.yaml
harshit-splunk: > These can already be set w/ <agent|gateway|clusterReceiver>.config.processors.batch.<field> right?

Yes, I had missed that. I'll remove it.

helm-charts/splunk-otel-collector/templates/config/_common.tpl
harshit-splunk: Makes sense.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Added
+
+- Add retry_on_failure/sending_queue config to values.yaml (#460)
```
rmfitzpatrick, line 9:
```suggestion
- Add `splunkPlatform.retryOnFailure` and `splunkPlatform.sendingQueue` config options to values.yaml (#460)
```

rmfitzpatrick (approved):
lgtm w/ doc updates

PR #461, opened 2022-06-03
Specify host mountinfo to agent self to prevent scraping errors

rmfitzpatrick:
Per https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/5879 https://github.com/shirou/gopsutil/pull/1133 the current agent configuration* almost guarantees failed disk usage metric generation by using unavailable host paths (which may be compounded by current SA receiver functionality: https://github.com/signalfx/splunk-otel-collector/pull/1615).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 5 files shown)


mpepping:
@rmfitzpatrick this still seems feasible to prevent the host path errors. Whilst it is a workaround, can we stage the PR?

dmitryax (approved):

PR #482, opened 2022-07-11
Ensure fluentd config path is respected

bambro:
This PR fixes an issue where `fluentd.path` is incorrectly used instead of `fluentd.config.path`, causing the default value `/var/log/containers/*.log` to be used.


jvoravong (approved):

jvoravong:
Thanks for finding and fixing this @bambro.

PR #486, opened 2022-07-13
Removed unnecessary change of group ownership in chmod initContainer

harshit-splunk:
~~Instead of modifying permissions of the entire pod/container directory, modify only logs file.~~

A file access control list (ACL) can provide permissions to a specific user or group without modifying the user/group owner of the file.
So, there is no need to change the group owner here:
https://github.com/signalfx/splunk-otel-collector-chart/blob/f5f6df821655f34a2c4e9c126edc5283be7a79d4/helm-charts/splunk-otel-collector/templates/daemonset.yaml#L147-L149

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Removed unnecessary change of group ownership in chmod initContainer (#486)
+
 ## [0.55.0] - 2022-07-19
 
 ### Changed
--- a/helm-charts/splunk-otel-collector/templates/daemonset.yaml
+++ b/helm-charts/splunk-otel-collector/templates/daemonset.yaml
@@ -144,21 +144,21 @@ spec:
           {{ if .Values.logsCollection.containers.enabled -}}
           if [ -d "/var/lib/docker/containers" ];
           then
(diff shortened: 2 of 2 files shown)


dmitryax (commented):
This is an init container which applies chown one time only, any new pods created after that won't get the right owner flags.

harshit-splunk:
> This is an init container which applies chown one time only, any new pods created after that won't get the right owner flags.

right! I'll close the PR

harshit-splunk:
@dmitryax 
https://github.com/signalfx/splunk-otel-collector-chart/blob/d104bb85d9633200473aa487538ad7fa13c7bffd/helm-charts/splunk-otel-collector/templates/daemonset.yaml#L147-L149

Is there any specific reason we need to change the group of all files inside the container directory?
I think we can avoid that and use `setfacl` without changing the original group.

harshit-splunk:
> @harshit-splunk does it break this functionality openshift?

it doesn't work with the current implementation either for openshift. It only gives read access to existing pod files. Agent will not have read access to newly created pods.

dmitryax:
Sounds good. @harshit-splunk can you please add make the statement that it doesn't work in cri-o more clear [here](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/advanced-configuration.md#running-the-container-in-non-root-user-mode), not just didn't work during testing

dmitryax (approved):

PR #489, opened 2022-07-16
Fix k8s events collection with enabled gateway

dmitryax:
Fix a bug: if gateway enabled, otlp exporter is used for k8s events in the pipeline, but it's not defined.


dmitryax:
Creted https://github.com/signalfx/splunk-otel-collector-chart/pull/491 instead

Issue #455, opened 2022-05-27
The batch processor is placed before filter


lindhe:
I recently read through [the docs for batch processor](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor) and it said:

> The batch processor should be defined in the pipeline after [...] any data drops such as sampling.

I noticed that this chart places batch _before_ filter. If that is intentional, I think it should be documented. Otherwise, this sounds like something that could be an easy performance improvement to make.

Here's some places where it's placed before filter:

* https://github.com/signalfx/splunk-otel-collector-chart/blob/4d78c50b85cb00be101260c82dcd68b70e1bd763/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl#L623-L626
* https://github.com/signalfx/splunk-otel-collector-chart/blob/4d78c50b85cb00be101260c82dcd68b70e1bd763/helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl#L225-L226

PR #498, opened 2022-07-22
Example for filtering containers

billg-splunk:
Adding an example to filter containers by name.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -204,9 +204,9 @@ fluentd:
           firstline: /\d{4}-\d{1,2}-\d{1,2}/
 ```
 
-## Filter out specific pods
+## Filter out specific containers
 
-This example shows how you can filter out specific pods from metrics
+This example shows how you can filter out specific containers from metrics
 pipelines. This could be adapted for other metadata or pipelines.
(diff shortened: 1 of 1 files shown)


billg-splunk:
One note - this example should conform to the actual order of processors. (It is based on the current order, but depending on the result of #455 (batch should be after filtering) it should be updated with that order.

examples/README.md
```diff
@@ -204,6 +204,53 @@ fluentd:
           firstline: /\d{4}-\d{1,2}-\d{1,2}/
 ```
 
+## Filter out specific pods
```
rmfitzpatrick, line 207: nit, this is for specific containers in a pod since `k8s.pod.name` would be the attribute to use otherwise: https://github.com/open-telemetry/opentelemetry-specification/blob/a25d5f03ab58ecf88c09f635df97d2328b5ba237/semantic_conventions/resource/k8s.yaml#L50

billg-splunk: Agreed, will fix

rmfitzpatrick (approved):
Thanks!

billg-splunk:
> Thanks!

Thank YOU! 🤜 🤛

Issue #520, opened 2022-08-26
Removing metrics/agent from Pipelines


srawale:
We want only traces as part of the helm chart computed values. For this e have set metricsEnabled: false everywhere in values.yaml. However computed values still show -
      pipelines:
        metrics/agent:
          exporters:
          - signalfx
          processors:
          - memory_limiter
          - batch
          - resource/add_agent_k8s
          - resourcedetection
          - resource
          receivers:
          - prometheus/agent

As we do not have any license for metrics we do not want this section to occur in computed values for helm chart as deploying this is giving us warning messages. We would like to know what all other things are needed to disable metric totally.

jvoravong:
That looks like a bug.
Created: https://github.com/signalfx/splunk-otel-collector-chart/pull/521

dmitryax:
This is not a bug, this is done intentionally. Self reporting metrics are important for collector's observability. Users need to see how collector behaves. And the metrics are free of charge on Splunk Observability side with purchased APM product.

srawale:
<img width="1791" alt="Screenshot 2022-09-01 at 3 26 05 PM" src="https://user-images.githubusercontent.com/34511394/187888399-d73cbe7f-71d8-4741-8ec9-f715535abb4c.png">
@dmitryax screenshot attached.

MyneniMahesh:
@dmitryax We have only subscribed for the APM module. Since the new collector is publishing the collector metrics we are getting the above warning messages

atoulme:
I am following up with the team. Please feel free to file a support ticket referencing this issue in the mean time. Thank you!

MyneniMahesh:
@atoulme , I have created the support please find the below ticket. Could you please let us know once the issue is resolved.
We are waiting for this issue to be resolved to upgrade our signafx smart agent to splunk OpenTelemetry Collector. 
Thanks for looking in to it. 

https://splunkcommunities.force.com/customer/s/case/5005a00002ESlbrAAD/we-have-only-subscribed-for-the-apm-module-since-the-new-collector-is-publishing-the-collector-metrics-by-default-we-are-getting-warning-messages-in-signalfx-sas-platform

atoulme:
The support case is under review and will be updated by our team - it is best to use it going forward to track the issue. Specifically, the team will communicate with you resolution over the support case. Closing this ticket.

PR #521, opened 2022-08-26
Fix a bug where collector internal metrics (pipeline: metrics/agent) are always exported even if metricsEnabled=false

jvoravong:
We always include the internal collector metrics pipeline even if metricsEnabled=false.
Addresses https://github.com/signalfx/splunk-otel-collector-chart/issues/520

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Fixed
+
+Fix a bug where collector internal metrics (pipeline: metrics/agent) are always exported even if metricsEnabled=false (#521)
```
dmitryax, line 9: This is not a bug. Let's hold off until the discussion in the issue is resolved

PR #528, opened 2022-09-02
Add options to create pod security policies during chart installation

mr-miles:
When installing the 0.58.0 chart on EKS, nothing worked until I added a psp.  This was an option in the splunk-connect-for-kubernetes chart, so this PR adds the same ability to the splunk-otel-collector-chart.

Adding parameters to match the splunk-connect-for-kubernetes chart, and the policy that worked for my setup.


dmitryax:
PSP was removed from this repo some time ago as its deprecated and will be dropped in Kubernetes 1.25 https://kubernetes.io/docs/concepts/security/pod-security-policy/

jheiselman:
Not everyone can upgrade their k8s versions as quickly as they are released. We are currently on 1.21 still and will be until probably next year. I would suggest adopting a policy of not removing features that are still supported, even if they are deprecated as it makes adoption much harder for those with very large (>500 nodes), sometimes slower moving infrastructure.

dmitryax:
@mr-miles @jheiselman I understand the concern, we don't want to add a config option that is broken on the latest kubernetes version. I added a document covering how to manually install the PSP https://github.com/signalfx/splunk-otel-collector-chart/pull/580

jheiselman:
Excellent. That works for me. Thank you.

Issue #506, opened 2022-08-09
Add image option in Values Yaml file init container


kumachop2:
We are  not enabled internet access our cluster and need to point to local registry. Please refer below link and add redhat image option in values.yaml file.

https://github.com/signalfx/splunk-otel-collector-chart/blob/153db1f22d62480c10b7a6cc50d754e1a07a4597/helm-charts/splunk-otel-collector/templates/daemonset.yaml#L139

atoulme:
Closing this issue as the PR #534 offers this possibility. Please comment if more work is needed.

PR #534, opened 2022-09-09
Add image option in Values file

pratik141:
Add image option in Values Yaml file init container #506

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/_helpers.tpl
+++ b/helm-charts/splunk-otel-collector/templates/_helpers.tpl
@@ -210,6 +210,13 @@ Create the opentelemetry collector image name.
 {{- end -}}
 
 {{/*
+Create the Universal Base image name.
+*/}}
+{{- define "splunk-otel-collector.image.ubi" -}}
+{{- printf "%s:%s" .Values.image.UniversalBaseImage.repository (.Values.image.UniversalBaseImage.tag | default "latest") -}}
+{{- end -}}
+
+{{/*
 Convert memory value from resources.limit to numeric value in MiB to be used by otel memory_limiter processor.
 */}}
 {{- define "splunk-otel-collector.convertMemToMib" -}}
--- a/helm-charts/splunk-otel-collector/templates/_helpers.tpl
+++ b/helm-charts/splunk-otel-collector/templates/_helpers.tpl
@@ -212,7 +212,7 @@ Create the opentelemetry collector image name.
 {{/*
 Create the chown image name.
 */}}
-{{- define "splunk-otel-collector.image.ubi" -}}
+{{- define "splunk-otel-collector.image.chown" -}}
 {{- printf "%s:%s" .Values.image.chown.repository (.Values.image.chown.tag | default "latest") -}}
 {{- end -}}
 
--- a/helm-charts/splunk-otel-collector/templates/_helpers.tpl
+++ b/helm-charts/splunk-otel-collector/templates/_helpers.tpl
@@ -210,10 +210,10 @@ Create the opentelemetry collector image name.
 {{- end -}}
 
 {{/*
-Create the Universal Base image name.
+Create the chown image name.
 */}}
 {{- define "splunk-otel-collector.image.ubi" -}}
-{{- printf "%s:%s" .Values.image.UniversalBaseImage.repository (.Values.image.UniversalBaseImage.tag | default "latest") -}}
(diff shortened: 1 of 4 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -833,6 +833,15 @@ image:
     # The policy that specifies when the user wants the opentelemetry collector images to be pulled
     pullPolicy: IfNotPresent
 
+  UniversalBaseImage:
+    # The registry and name of the Universal Base Image 8 image to pull
+    repository: registry.access.redhat.com/ubi8/ubi
+    # The tag of the Splunk OTel Collector image, default value is the chart appVersion
```
atoulme, line 839: this should be latest

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -833,6 +833,15 @@ image:
...
+    tag: ""
+    # The policy that specifies when the user wants the opentelemetry collector images to be pulled
```
atoulme, line 841: this comment needs to be updated as well

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -833,6 +833,14 @@ image:
     # The policy that specifies when the user wants the opentelemetry collector images to be pulled
     pullPolicy: IfNotPresent
 
+  UniversalBaseImage:
```
atoulme, line 836:
```suggestion
  universalBaseImage:
```

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -833,6 +833,14 @@ image:
     # The policy that specifies when the user wants the opentelemetry collector images to be pulled
     pullPolicy: IfNotPresent
 
+  universalBaseImage:
```
dmitryax, line 836: It doesn't have to UBI, can be any other linux image. So let's call it same as the container name for consistency with options in this section.
```suggestion
  # Image to be used by init container that patches log directories on the host, so the collector can read from them as a non-root user.
  # Effective only if `agent.securityContext.runAsUser` and `agent.securityContext.runAsGroup` are set to non-zero values.
  chown:
```

pratik141: updated

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -209,6 +209,13 @@ Create the opentelemetry collector image name.
 {{- printf "%s:%s" .Values.image.otelcol.repository (.Values.image.otelcol.tag | default .Chart.AppVersion) -}}
 {{- end -}}
 
+{{/*
+Create the chown image name.
+*/}}
+{{- define "splunk-otel-collector.image.ubi" -}}
```
dmitryax, line 215:
```suggestion
{{- define "splunk-otel-collector.image.chown" -}}
```

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -136,7 +136,7 @@ spec:
         {{- end }}
         {{- if and $agent.securityContext.runAsUser $agent.securityContext.runAsGroup }}
         - name: chown
-          image: registry.access.redhat.com/ubi8/ubi
+          image: {{ template "splunk-otel-collector.image.ubi" . }}
```
dmitryax, line 139:
```suggestion
          image: {{ template "splunk-otel-collector.image.chown" . }}
```

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -833,6 +833,16 @@ image:
     # The policy that specifies when the user wants the opentelemetry collector images to be pulled
     pullPolicy: IfNotPresent
 
+    # Image to be used by init container that patches log directories on the host, so the collector can read from them as a non-root user.
+    # Effective only if `agent.securityContext.runAsUser` and `agent.securityContext.runAsGroup` are set to non-zero values.
```
dmitryax, line 837:
```suggestion
  # Image to be used by init container that patches log directories on the host, so the collector can read from them as a non-root user.
  # Effective only if `agent.securityContext.runAsUser` and `agent.securityContext.runAsGroup` are set to non-zero values.
```

dmitryax (approved):

PR #537, opened 2022-09-21
Internal metrics, only if metrics are enabled

atoulme:
Do not create a pipeline to send internal metrics if no metrics are enabled.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
@@ -239,7 +239,7 @@ service:
         {{- end }}
     {{- end }}
 
-    {{- if and (or (eq (include "splunk-otel-collector.splunkO11yEnabled" $) "true") (eq (include "splunk-otel-collector.platformMetricsEnabled" $) "true")) (or (eq (include "splunk-otel-collector.o11yMetricsEnabled" .) "true") (eq (include "splunk-otel-collector.platformMetricsEnabled" $) "true")) }}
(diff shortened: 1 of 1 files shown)


dloucasfx (commented):
So the idea was to keep internal metrics intact, even if `o11yMetricsEnabled` is true
This was discussed in https://github.com/signalfx/splunk-otel-collector-chart/issues/520 and internally.
The fix was to treat internal metrics in the backend as bundled/free.
I don't think we want to remove the whole pipeline

dloucasfx (approved):

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
```diff
@@ -239,7 +239,7 @@ service:
         {{- end }}
     {{- end }}
 
-    {{- if or (eq (include "splunk-otel-collector.splunkO11yEnabled" $) "true") (eq (include "splunk-otel-collector.platformMetricsEnabled" $) "true") }}
+    {{- if and (or (eq (include "splunk-otel-collector.splunkO11yEnabled" $) "true") (eq (include "splunk-otel-collector.platformMetricsEnabled" $) "true")) (or (eq (include "splunk-otel-collector.splunkO11yEnabled" .) "true") (eq (include "splunk-otel-collector.platformMetricsEnabled" $) "true")) }}
```
dmitryax, line 242: Why is this change needed? looks like both and argument are the same, no?

atoulme: Oh they are the same now. I missed that. I will fix.

atoulme: see https://github.com/signalfx/splunk-otel-collector-chart/pull/544

PR #550, opened 2022-10-03
Add more documentation about k8s pod labels extraction

dmitryax:


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -202,15 +202,21 @@ extraAttributes:
...
   #  If you want to change the default attribute name `k8s.pod.labels.<label_name>`, you could do that using a `tag_name` field:
   #  fromLabels:
   #    - key: my_key
   #      tag_name: my_tag
   #      from: pod
+  #
+  #  `key_regex` field can be used to get a specific set of labels that match a regex.
+  #  If `key_regex` used is used, the `key` field accepts regexp matching groups.
+  #  The following example will fetch all the pod labels and propagate them to the attributes as is,
+  #  without "k8s.pod.labels." prefix. "$" from the matching group must be escaped as "$$".
```
atoulme, line 215: Does the `k8s.pod.labels` need to be specified or is it removed by default?
Should this example live under https://github.com/signalfx/splunk-otel-collector-chart/tree/main/examples ?

dmitryax: `k8s.pod.labels` is removed by overriding `tag_name`. This doc is just to cover more functionality of this config option rather than just an example, so I would keep it here

atoulme: OK.

atoulme (approved):

PR #554, opened 2022-10-05
Increase number of queue consumers in gateway default configuration

dmitryax:
Given that collector is supposed to handle higher data volume in gateway mode, we want to increase default numer of queue consumers.

Related issue: https://github.com/signalfx/splunk-otel-collector/issues/2082


atoulme (approved):
LGTM.

PR #568, opened 2022-10-20
add allowedCapabilities to the openshift scc

dloucasfx:
The current chart is resulting of this error when deployed on Azure openshift

```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(SecurityContextConstraints): missing required field "allowedCapabilities" in com.github.openshift.api.security.v1.SecurityContextConstraints
```
Adding the below should fix it

```
allowedCapabilities: []
```

Note: I still need to validate the fix on Axure Openshift
Signed-off-by: Dani Louca <dlouca@splunk.com>

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/securityContextConstraints.yaml
+++ b/helm-charts/splunk-otel-collector/templates/securityContextConstraints.yaml
@@ -33,6 +33,7 @@ seLinuxContext:
     type: "spc_t"
     level: "s0"
 allowedFlexVolumes: []
+allowedCapabilities: []
 defaultAddCapabilities: []
 fsGroup:
   type: MustRunAs


dloucasfx:
FYI, I don't have an openshift cluster to repro, but based on the api docs, openshift 3.11 requires `allowedCapabilities` https://docs.openshift.com/container-platform/3.11/rest_api/security_openshift_io/securitycontextconstraints-security-openshift-io-v1.html
and 
but 4.x does not require it https://docs.openshift.com/container-platform/4.4/rest_api/security_apis/securitycontextconstraints-security-openshift-io-v1.html

jvoravong (approved):

PR #574, opened 2022-11-02
prepare release 0.62.1

atoulme:


dmitryax (approved):

atoulme:
The checksums and most of the work is done by calling “make render” which updates the final yaml files. The only manual change is the update of the version of the chart.

PR #578, opened 2022-11-09
Add network-explorer

dustin-bingham:


atoulme:
Nice work! I think it would be good to have a rendered folder for just the network explorer. If you look at the render task in the Makefile, you should be able to add a rendered folder for just the network explorer.

dustin-bingham:
Will do!

helm-charts/splunk-otel-collector/templates/configmap-gateway.yaml
```diff
@@ -1,5 +1,6 @@
 {{ $gateway := fromYaml (include "splunk-otel-collector.gateway" .) }}
-{{ if $gateway.enabled }}
+{{ $gatewayEnabled := eq (include "splunk-otel-collector.gatewayEnabled" .) "true" }}
+{{ if $gatewayEnabled }}
```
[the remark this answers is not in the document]
dustin-bingham: It collapses the logic of checking whether networkExplorer is enabled and if gateway under networkExplorer is enabled.  It appeared more readable to me this way?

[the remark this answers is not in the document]
atoulme: I understand now. Sure that works!

helm-charts/splunk-otel-collector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: splunk-otel-collector
-version: 0.62.1
+version: 0.62.2
```
atoulme, line 3: if you don't mind - can we revert that change here and in rendered files to keep the diff a bit smaller?

dustin-bingham:
Sure no problem!

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1022,3 +1022,269 @@ service:
...
+      ## Ref: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#seccomp
+      ## Ref: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#sysctl
+      ##
+      # seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
+      # seccomp.security.alpha.kubernetes.io/defaultProfileName: 'docker/default'
+      # apparmor.security.beta.kubernetes.io/defaultProfileName: 'runtime/default'
+
+  # otel collector in gateway configuration for network explorer.
+  # these values override those specified above.
+  gateway:
```
dmitryax, line 1279: Why do we need another section for the gateway? Can't we reuse existing `gateway`?

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -324,16 +324,46 @@ compatibility with the old config group name: "otelAgent".
 {{- end }}
 {{- end -}}
 
+{{/*
+The name of the gateway service.
+*/}}
+{{- define "splunk-otel-collector.gatewayServiceName" -}}
+{{  (include "splunk-otel-collector.fullname" . ) | trunc 63 | trimSuffix "-" }}
+{{- end -}}
```
dmitryax, line 332: Is this helper being used?

dustin-bingham: Yes, it's being used in the reducer-deployment.yml - it's the destination for the reducer's output.

helm-charts/splunk-otel-collector/values.yaml
dustin-bingham: These are used to supplement / overwrite the existing gateway settings.  In the _helpers template you can see the logic for it (the gateway define)  I'm not at all sold on doing it this way, however.  Basically the important things we want to do is set the replicas down to 1, ensure that the gateway is enabled, and especially set the limits lower (based on feedback we've received)  So if you think we ought to decompose this to the individual settings or something, that's fine too.

helm-charts/splunk-otel-collector/templates/_helpers.tpl
dustin-bingham: Actually - I intended to use it there, but I ended up just using fullname.  Let me fix that up real quick.

helm-charts/splunk-otel-collector/values.yaml
dmitryax: The problem is that we expose the same user configuration interface in two places which makes it confusing and harder to maintain I believe. We can still use one config interface even if different default values applied based on how the gateway is enabled. But it would be better to find some middle ground and have a reasonable default for both use cases.

> set the replicas down to 1
Why is this required? It's not the best choice for HA

> ensure that the gateway is enabled
You're already doing it with the gatewayEnabled helper, right?

> especially set the limits lower (based on feedback we've received)
The gateway is recommended to be enabled on big clusters, that's why we have such high limits. We can discuss something in between if you agree.

helm-charts/splunk-otel-collector/templates/pdb-gateway.yaml
```diff
@@ -1,5 +1,6 @@
 {{ $gateway := fromYaml (include "splunk-otel-collector.gateway" .) }}
-{{- if and $gateway.enabled .Values.podDisruptionBudget }}
+{{ $gatewayEnabled := eq (include "splunk-otel-collector.gatewayEnabled" .) "true" }}
+{{- if and $gatewayEnabled .Values.podDisruptionBudget }}
```
dmitryax, line 3: This is a breaking change. `splunk-otel-collector.gatewayEnabled` should use `splunk-otel-collector.gatewayEnabled` helper instead of `gateway.enabled` to keep the same behavior

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -324,6 +324,20 @@ compatibility with the old config group name: "otelAgent".
...
+
+{{/*
+Whether the gateway is enabled, either through network explorer, or through its own flag.
+*/}}
+{{- define "splunk-otel-collector.gatewayEnabled" -}}
+{{- or .Values.gateway.enabled .Values.networkExplorer.enabled }}
```
dmitryax, line 338: Please use `splunk-otel-collector.gateway` helper instead `.Values.gateway` to preserve current behavior in all the places where `splunk-otel-collector.gatewayEnabled` is used

helm-charts/splunk-otel-collector/templates/network-explorer/kernel-collector-podsecuritypolicy.yaml
```diff
@@ -0,0 +1,36 @@
+{{- if and .Values.networkExplorer.enabled .Values.networkExplorer.podSecurityPolicy.enabled }}
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
```
dmitryax, line 3: PSPs are removed from kubernetes in 1.25, see https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future. Can we do not include it?

tigrannajaryan:
@dustin-bingham I think this adds significant complexity to the Helm chart that warrants having a system diagram that describes the Helm chart itself.

Can you put together such a diagram with some description of how things work, including the components that already exist in the Helm chart (the Collector daemonset and the gateway)?

helm-charts/splunk-otel-collector/templates/network-explorer/k8s-collector-deployment.yaml
```diff
@@ -0,0 +1,162 @@
+{{- if and .Values.networkExplorer.enabled .Values.networkExplorer.k8sCollector.enabled }}
+{{ $gateway := fromYaml (include "splunk-otel-collector.gateway" .) }}
+# The k8s-collector consists of two services:
+# 1) k8s-watcher: talks to the Kubernetes API server to determine the current state of
+#    the cluster; sets up watches to be notified of subsequent changes to pods, services
+#    and other resources.
+# 2) k8s-relay: relays the information collected by k8s-watcher to the reducer.
```
dmitryax, line 7: Looks like this functionality supported by [collector k8sattributes processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor). Not asking for any changes in scope of this PR, just thoughts for future integration: I believe we can adjust the k8s attribute processor for this use case and use it instead of this component given that we are sending data through the collector anyway.

dmitryax (approved):
LGTM. Do we also need to mention the network explorer in the chart README.md and maybe add installation instructions in https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/advanced-configuration.md?

dmitryax:
Also please add an entry in the changelog. Can be done in a separate PR

PR #588, opened 2022-11-21
add k8s object receiver

omrozowicz-splunk:
Replacing `k8seventsreceiver` with `k8objectsreceiver`.

As using k8s objects receiver requires creating new rules in `clusterRole.yaml`, the PR contains `splunk-otel-collector.clusterRole.rules` which is the template function that aggregates and renders all the apiGroups rules with "get", "list", "watch" verbs, so there is no duplications.

It requires splunk-otel-collector with https://github.com/signalfx/splunk-otel-collector/pull/2270 to be released.

There are slight indent differences in rendered templates because previously it was not entirely consistent (sometimes with whitespaces after apiGroups, sometimes not). With the change from thsi PR it's handled by the function:

```
- apiGroups:
  - {{ $groupName }}
```

Also, `eventsEnabled` name was changed to `objectsEnabled` so the name reflects better what actually we pull.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,38 +1,65 @@
 # Upgrade guidelines
 
-## 0.64.0 to 0.66.0
-Before this change, [k8s events receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8seventsreceiver) was used to collect Kubernetes Events.
-Now we utilize [k8s objects receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver) that can pull or watch any object from Kubernetes API server.
-Therefore, `clusterReceiver.eventsEnabled` is now deprecated, and to maintain the same behavior you should set
-`clusterReceiver.objectsEnabled` to `true` and configure `clusterReceiver.k8sObjects` to watch event objects:
+## $CURRENT_VERSION to $NEXT_VERSION
+
+There is a new receiver: [Kubernetes Objects Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver) that can pull or watch any object from Kubernetes API server.
+It will replace the [Kubernetes Events Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8seventsreceiver) in the future.
+
+To migrate from Kubernetes Events Receiver to Kubernetes Object Receiver, configure `clusterReceiver` values.yaml section with:
+
 ```yaml
-  objectsEnabled: true
-  k8sObjects:
-    - mode: watch
-      name: events
+k8sObjects:
+  - mode: watch
+    name: events
 ```
-You can monitor more kubernetes objects configuring `clusterReceiver.k8sObjects` according to the instruction from
-[k8s objects receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver).
 
-Remember to define `rbac.customRules` when needed.
+There are differences in the log record formatting between the previous `k8s_events` receiver and the now adopted `k8sobjects` receiver results.
+The `k8s_events` receiver stores event messages their log body, with the following fields added as attributes:
 
-There is a difference in the event formatting between `k8s_events` receiver and `k8sobjects` receiver results.
-`k8s_events` receiver stores event message as a log body, with the following fields added as a resource/attribute:
-* k8s.object.kind
-* k8s.object.name
-* k8s.object.uid
-* k8s.object.fieldpath
-* k8s.object.api_version
-* k8s.object.resource_version
-* k8s.event.reason
-* k8s.event.action
-* k8s.event.start_time
-* k8s.event.name
-* k8s.event.uid
-* k8s.namespace.name
+* `k8s.object.kind`
+* `k8s.object.name`
+* `k8s.object.uid`
+* `k8s.object.fieldpath`
+* `k8s.object.api_version`
+* `k8s.object.resource_version`
+* `k8s.event.reason`
+* `k8s.event.action`
+* `k8s.event.start_time`
+* `k8s.event.name`
+* `k8s.event.uid`
+* `k8s.namespace.name`
 
-In case of `k8sobjects`, the whole payload is stored in the log body and `object.message` refers to the event message.
+Now with the `k8sobjects` receiver, the whole payload is stored in the log body and `object.message` refers to the event message.
 
+
+You can monitor more Kubernetes objects configuring by `clusterReceiver.k8sObjects` according to the instructions from the
+[Kubernetes Objects Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver) documentation.
+
+Remember to define `rbac.customRules` when needed. For example, when configuring:
+
+```yaml
+objectsEnabled: true
+k8sObjects:
+  - name: events
+    mode: watch
+    group: events.k8s.io
+    namespaces: [default]
+```
+
+You should add `events.k8s.io` API group to the `rbac.customRules`:
+
+```yaml
+rbac:
+  customRules:
+    - apiGroups:
+      - "events.k8s.io"
+      resources:
+      - events
+      verbs:
+      - get
+      - list
+      - watch
+```
 
 ## 0.58.0 to 0.59.0
 [receiver/filelogreceiver] Datatype for `force_flush_period` and `poll_interval` were changed from map to string.
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,39 @@
 # Upgrade guidelines
 
+## 0.64.0 to 0.66.0
+Before this change, [k8s events receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8seventsreceiver) was used to collect Kubernetes Events.
+Now we utilize [k8s objects receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver) that can pull or watch any object from Kubernetes API server.
+Therefore, `clusterReceiver.eventsEnabled` is now deprecated, and to maintain the same behavior you should set
+`clusterReceiver.objectsEnabled` to `true` and configure `clusterReceiver.k8sObjects` to watch event objects:
+```yaml
+  objectsEnabled: true
+  k8sObjects:
+    - mode: watch
+      name: events
+```
+You can monitor more kubernetes objects configuring `clusterReceiver.k8sObjects` according to the instruction from
+[k8s objects receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver).
+
+Remember to define `rbac.customRules` when needed.
+
+There is a difference in the event formatting between `k8s_events` receiver and `k8sobjects` receiver results.
+`k8s_events` receiver stores event message as a log body, with the following fields added as a resource/attribute:
+* k8s.object.kind
+* k8s.object.name
+* k8s.object.uid
+* k8s.object.fieldpath
+* k8s.object.api_version
+* k8s.object.resource_version
+* k8s.event.reason
+* k8s.event.action
+* k8s.event.start_time
+* k8s.event.name
+* k8s.event.uid
+* k8s.namespace.name
+
+In case of `k8sobjects`, the whole payload is stored in the log body and `object.message` refers to the event message.
+
+
 ## 0.58.0 to 0.59.0
 [receiver/filelogreceiver] Datatype for `force_flush_period` and `poll_interval` were changed from map to string.
 
--- a/helm-charts/splunk-otel-collector/templates/_helpers.tpl
+++ b/helm-charts/splunk-otel-collector/templates/_helpers.tpl
@@ -395,3 +395,12 @@ compatibility with the old config group name: "otelK8sClusterReceiver".
 {{- $clusterReceiver.k8sEventsEnabled }}
 {{- end }}
 {{- end -}}
+
+{{/*
+Whether clusterReceiver should be deployed
+*/}}
+{{- define "splunk-otel-collector.clusterReceiverDeploy" -}}
+{{- $clusterReceiver := fromYaml (include "splunk-otel-collector.clusterReceiver" .) }}
+{{- and $clusterReceiver.enabled (or (eq (include "splunk-otel-collector.metricsEnabled" .) "true") $clusterReceiver.objectsEnabled ) -}}
+{{- end -}}
+
--- a/helm-charts/splunk-otel-collector/templates/_helpers.tpl
+++ b/helm-charts/splunk-otel-collector/templates/_helpers.tpl
@@ -396,11 +396,32 @@ compatibility with the old config group name: "otelK8sClusterReceiver".
 {{- end }}
 {{- end -}}
 
+
 {{/*
-Whether clusterReceiver should be deployed
+Whether object collection by k8s object receiver is enabled
 */}}
-{{- define "splunk-otel-collector.clusterReceiverDeploy" -}}
+{{- define "splunk-otel-collector.objectsEnabled" -}}
 {{- $clusterReceiver := fromYaml (include "splunk-otel-collector.clusterReceiver" .) }}
-{{- and $clusterReceiver.enabled (or (eq (include "splunk-otel-collector.metricsEnabled" .) "true") $clusterReceiver.objectsEnabled ) -}}
+{{- if gt (len $clusterReceiver.k8sObjects) 0 }}
+{{- printf "true" }}
+{{- else }}
+{{- printf "false" }}
+{{- end -}}
 {{- end -}}
 
+{{/*
+Whether object collection by k8s object receiver or/and event collection by k8s event receiver is enabled
+*/}}
+{{- define "splunk-otel-collector.objectsOrEventsEnabled" -}}
+{{- $clusterReceiver := fromYaml (include "splunk-otel-collector.clusterReceiver" .) }}
+{{- or $clusterReceiver.eventsEnabled ( eq (include "splunk-otel-collector.objectsEnabled" .) "true") -}}
+{{- end -}}
+
+
+{{/*
+Whether clusterReceiver should be enabled
+*/}}
+{{- define "splunk-otel-collector.clusterReceiverEnabled" -}}
+{{- $clusterReceiver := fromYaml (include "splunk-otel-collector.clusterReceiver" .) }}
+{{- and $clusterReceiver.enabled (or (eq (include "splunk-otel-collector.metricsEnabled" .) "true") (eq (include "splunk-otel-collector.objectsOrEventsEnabled" .) "true")) -}}
+{{- end -}}
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
@@ -35,9 +35,10 @@ receivers:
     {{- if eq (include "splunk-otel-collector.distribution" .) "openshift" }}
     distribution: openshift
     {{- end }}
-  {{- if and $clusterReceiver.eventsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
-  k8s_events:
+  {{- if and $clusterReceiver.objectsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
+  k8sobjects:
     auth_type: serviceAccount
+    objects: {{ .Values.clusterReceiver.k8sObjects | toJson }}
   {{- end }}
   {{- if eq (include "splunk-otel-collector.o11yInfraMonEventsEnabled" .) "true" }}
   smartagent/kubernetes-events:
@@ -89,7 +90,7 @@ processors:
         value: {{ .Values.clusterName }}
   {{- end }}
 
-  {{- if and $clusterReceiver.eventsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
+  {{- if and $clusterReceiver.objectsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
   # Drop high cardinality k8s event attributes
   attributes/drop_event_attrs:
     actions:
@@ -132,7 +133,7 @@ processors:
         value: {{ .value }}
       {{- end }}
 
-  {{- if and $clusterReceiver.eventsEnabled .Values.environment }}
+  {{- if and $clusterReceiver.objectsEnabled .Values.environment }}
   resource/add_environment:
     attributes:
       - action: insert
@@ -158,7 +159,7 @@ exporters:
     timeout: 10s
   {{- end }}
 
-  {{- if and (eq (include "splunk-otel-collector.o11yLogsEnabled" .) "true") $clusterReceiver.eventsEnabled }}
+  {{- if and (eq (include "splunk-otel-collector.o11yLogsEnabled" .) "true") $clusterReceiver.objectsEnabled }}
   splunk_hec/o11y:
     endpoint: {{ include "splunk-otel-collector.o11yIngestUrl" . }}/v1/log
     token: "${SPLUNK_OBSERVABILITY_ACCESS_TOKEN}"
@@ -172,7 +173,7 @@ exporters:
   {{- include "splunk-otel-collector.splunkPlatformMetricsExporter" . | nindent 2 }}
   {{- end }}
 
-  {{- if and (eq (include "splunk-otel-collector.platformLogsEnabled" .) "true") $clusterReceiver.eventsEnabled }}
+  {{- if and (eq (include "splunk-otel-collector.platformLogsEnabled" .) "true") $clusterReceiver.objectsEnabled }}
   {{- include "splunk-otel-collector.splunkPlatformLogsExporter" . | nindent 2 }}
     sourcetype: kube:events
   {{- end }}
@@ -231,10 +232,10 @@ service:
         {{- end }}
     {{- end }}
 
-    {{- if and $clusterReceiver.eventsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
+    {{- if and $clusterReceiver.objectsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
     logs:
       receivers:
-        - k8s_events
+        - k8sobjects
       processors:
         - memory_limiter
         - batch
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
@@ -107,12 +107,11 @@ processors:
   {{- end }}
 
   {{- if and (eq (include "splunk-otel-collector.objectsEnabled" .) "true") (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
-  # TODO: After updating to 0.66.0, change syntax according to
-  # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor
   transform/add_sourcetype:
-    logs:
-      statements:
-        - set(resource.attributes["com.splunk.sourcetype"], Concat(["kube:object:", attributes["event.name"]], ""))
+    log_statements:
+      - context: log
+        statements:
+          - set(resource.attributes["com.splunk.sourcetype"], Concat(["kube:object:", attributes["event.name"]], ""))
   {{- end }}
 
   # Resource attributes specific to the collector itself.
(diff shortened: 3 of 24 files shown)


helm-charts/splunk-otel-collector/templates/clusterRole.yaml
```diff
@@ -10,76 +10,7 @@ metadata:
...
-  - watch
-- apiGroups:
-    - autoscaling
-  resources:
-    - horizontalpodautoscalers
-  verbs:
-    - get
-    - list
-    - watch
+{{- include "splunk-otel-collector.clusterRole.rules" . }}
```
rmfitzpatrick, line 13: I think for readability and maintenance it's preferable to still have the rules in this file instead of migrating them to the values file which is already a little unwieldy.

```
Template files should have the extension .yaml if they produce YAML output.
The extension .tpl may be used for template files that produce no formatted content.
```
- https://helm.sh/docs/chart_best_practices/templates/

rmfitzpatrick:
Do you have any guidance and examples on the differences in reported telemetry between the two receivers? Presumably this is going to affect any user systems that depend on the events receiver.

rendered/manifests/eks-fargate/clusterRole.yaml
```diff
@@ -15,28 +15,6 @@ metadata:
     release: default
     heritage: Helm
 rules:
-- apiGroups:
```
rmfitzpatrick, line 18: Ideally only functional content changes would alter how the manifests are rendered.

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -390,8 +390,213 @@ compatibility with the old config group name: "otelK8sClusterReceiver".
 {{- define "splunk-otel-collector.o11yInfraMonEventsEnabled" -}}
 {{- $clusterReceiver := fromYaml (include "splunk-otel-collector.clusterReceiver" .) }}
 {{- if eq (toString $clusterReceiver.k8sEventsEnabled) "<nil>" }}
-{{- .Values.splunkObservability.infrastructureMonitoringEventsEnabled }}
+{{- .Values.splunkObservability.infrastructureMonitoringObjectsEnabled }}
```
rmfitzpatrick, line 393: "objects enabled" is more ambiguous than "events enabled" so how about something like "k8s object reporting enabled" for all applicable vars?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -123,11 +123,11 @@ splunkObservability:
...
   logsEnabled: false
 
-  # Option to send Kubernetes events to Splunk Observability Infrastructure Monitoring as data events:
+  # Option to send Kubernetes objects to Splunk Observability Infrastructure Monitoring as data events:
   # https://docs.splunk.com/Observability/alerts-detectors-notifications/view-data-events.html
-  # To send Kubernetes events to Splunk Observability Log Observer, set both clusterReceiver.eventsEnabled and
+  # To send Kubernetes events to Splunk Observability Log Observer, set both clusterReceiver.objectsEnabled and
   # splunkObservability.logsEnabled to true.
-  infrastructureMonitoringEventsEnabled: false
+  infrastructureMonitoringObjectsEnabled: false
```
dmitryax, line 130: Please don't rename this config option, `infrastructureMonitoringEvents` is the name of some functionality in Splunk Observability IM, not `infrastructureMonitoringObjects`.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -423,13 +423,65 @@ clusterReceiver:
...
+  objects:
+    v1:
+      - name: pods
+        mode: pull
+      - name: namespaces
+        mode: pull
+      - name: nodes
+        mode: pull
+      - name: events
+        mode: watch
```
dmitryax, line 484: Why do we need this config interface causing all the complicated translations? Why can't we use the same config interface provided by https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver?

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -35,9 +35,11 @@ receivers:
     {{- if eq (include "splunk-otel-collector.distribution" .) "openshift" }}
     distribution: openshift
     {{- end }}
-  {{- if and $clusterReceiver.eventsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
-  k8s_events:
+  {{- if and $clusterReceiver.objectsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
+  k8sobjects:
     auth_type: serviceAccount
+    objects:
+{{ include "splunk-otel-collector.objectsEnabled" . | nindent 6 }}
```
dmitryax, line 42: Shouldn't it be `splunk-otel-collector.clusterRole.clusterReceiver.objects`?

helm-charts/splunk-otel-collector/values.yaml
harshit-splunk: It is required to add rules for all objects in the ClusterRole manifest.
k8sobjectsreceiver automatically detects the resource apiGroups by looking up the resource name from the kube-apiserver. But I don't think that the helm chart provides such functionality.

harshit-splunk:
> Do you have any guidance and examples on the differences in reported telemetry between the two receivers? Presumably this is going to affect any user systems that depend on the events receiver.

k8sobject stores payload into the log body, but k8s_events only stores event.Message field as body and only adds following fields as resource/attribue
- k8s.object.kind
- k8s.object.name
- k8s.object.uid
- k8s.object.fieldpath
- k8s.object.api_version
- k8s.object.resource_version  
- k8s.event.reason
- k8s.event.action
- k8s.event.start_time
- k8s.event.name
- k8s.event.uid
- k8s.namespace.name

k8s_event payload
image

k8sobjects payload
image

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
omrozowicz-splunk: Right! actually, I'd even remove `clusterRole` element. To make it `splunk-otel-collector.clusterReceiver.objects`

helm-charts/splunk-otel-collector/templates/_helpers.tpl
omrozowicz-splunk: Below @dmitryax wrote `infrastructureMonitoringEventsEnabled  is the name of some functionality in Splunk Observability IM`, so I think I shouldn't have changed it at all

rendered/manifests/eks-fargate/clusterRole.yaml
omrozowicz-splunk: Ok, I brought the old way back and it detects no differences now

helm-charts/splunk-otel-collector/values.yaml
dmitryax: Do we really need the autogeneration of `ClusterRole` manifest based on this config? Introducing another config interface layer that is different from the receiver's one just to build `ClusterRole` with all the helm magic that needs to be maintained seems like an overcomplication to me.

What if we keep `eventsEnabled` and `objectsEnabled` that generate some default config for `k8sobjectsreceiver` with a generic list of the most common objects? And, if users need different behavior, they can reconfigure `clusterReceiver.config.receivers.k8sobjects.objects` and potentially `rbac.customRules`?

dmitryax:
> k8sobject stores payload into the log body, but k8s_events only stores event.Message field as body and only adds following fields as resource/attribute

This needs to be well documented in https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md

helm-charts/splunk-otel-collector/values.yaml
harshit-splunk: > Do we really need the autogeneration of ClusterRole manifest based on this config? Introducing another config interface layer that is different from the receiver's one just to build ClusterRole with all the helm magic that needs to be maintained seems like an overcomplication to me.

Makes sense. 

> What if we keep eventsEnabled and objectsEnabled that generate some default config for k8sobjectsreceiver with a generic list of the most common objects? And, if users need different behavior, they can reconfigure clusterReceiver.config.receivers.k8sobjects.objects and potentially rbac.customRules?

Agreed. It will allow us to follow the same config interface used by k8sobjects. 

Do we need both `eventsEnabled` and `objectsEnabled`? k8sobjects can already collect events but with different structures.

dmitryax: I agree that we should migrate to `k8sobjects` receiver for events. We can keep the same `eventsEnabled` option that renders part of `k8sobjects` receiver config instead of `k8sevents` receiver. The only concern I have is that this is a breaking change for users, we need to document it and provide a config snippet on how to configure `k8sevents` receiver via `clusterReceiver.config` for users that want to keep the existing behavior.

harshit-splunk: Maybe we can allow both k8sobject and k8s_event, and document that k8s_event will be removed in future and how to migrate.

dmitryax: Maybe we can deprecate `eventsEnabled` option in favor of just one `k8sObjects` option that will expose the k8sobjects receiver interface as is. `k8sObjects` default value will be `[]` resulting in the `k8sobjects` receiver not being set up. Deprecating `eventsEnabled`, we will ask the user to setup 

```yaml
  k8sObjects:
    - mode: watch
      name: events
```

instead. It should collect the same data just in a different format. Users can also specify any other objects additionally in `k8sObjects` section, but they need to make sure that `rbac.customRules` is defined when needed. WDYT?

harshit-splunk: Agreed

helm-charts/splunk-otel-collector/templates/clusterRole.yaml
omrozowicz-splunk: Thanks for the advice! Firstly I adjusted this, but then I deleted it completely, according to the conversation below. Now user needs to set it by himself using `rbac.customRules`

omrozowicz-splunk:
One more thing, now sourcetype of events generated by k8sobjectsreceiver is `kube:events`. Do we want to keep this convention? We could generate the sourcetype based on the resource we pull/watch, following `kube:object:<resource>, for ex. `kube:object:pods`. What do you think?

dmitryax:
We could generate the sourcetype based on the resource we pull/watch, following kube:object:<resource>, for ex. kube:object:pods. What do you think?

Sorry, I'm not an expert when it comes to Splunk Platform specific fields. What are the expected values for `sourcetype`? Do we have any conventions for that? 

And you want to achieve that? I think we can use the transform processor to compose sourcetype from event payload

harshit-splunk:
> 





> Sorry, I'm not an expert when it comes to Splunk Platform specific fields. What are the expected values for sourcetype? Do we have any conventions for that?

_The [source type](https://docs.splunk.com/Splexicon:Sourcetype) is one of the [default fields](https://docs.splunk.com/Splexicon:Defaultfield) that the Splunk platform assigns to all incoming data. It tells the platform what kind of data you have, so that it can format the data intelligently during indexing. Source types also let you categorize your data for easier searching._

In SCK, we are following aslo kube:object:<resource-name> convention.

> And how do you want to achieve that? I think we can use the transform processor to compose sourcetype from event payload

K8sobjects adds the resource name as an `event.name` attribute. I was thinking we can use [logstransformprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/logstransformprocessor)
 and use stanza's [add operator](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/add.md) to add `com.splunk.sourcetype` attribute.

There was a bug in k8sobject receiver. `event.name` attribute is empty when watch mode is used. I have created https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/16543 to fix that. Can you please review that as well?

dmitryax:
@harshit-splunk thanks for explaining

> In SCK, we are also following kube:object:<resourcename> convention.

Sounds good to me. Do we have to stick with what was adopted in SCK? Using `k8s` in `k8s:object:<resourcename>` I think is less confusing than `kube`.

> K8sobjects adds the resource name as an event.name attribute. I was thinking we can use [logstransformprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/logstransformprocessor)
and use stanza's [add operator](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/add.md) to add com.splunk.sourcetype attribute.

Let's better use transform processor since that is what will be the standard in OTel collector going forward

harshit-splunk:
> Do we have to stick with what was adopted in SCK? Using k8s in `k8s:object:<resourcename>` I think is less confusing than `kube`.

Agrred, but we are already using `kube` prefix for container logs (i.e. `kube:container:<container name>`. So, we should stick with same prefix.

> Let's better use [transform processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor) since that is what will be the standard in OTel collector going forward

Okay, makes sense.

dmitryax:
> Agrred, but we are already using kube prefix for container logs (i.e. kube:container:<container name>. So, we should stick with same prefix.

Ok let's keep `kube`

> We need to wait until this https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/16543 is merged

Reviewing it 

> Another thing is that in transform processor was a syntax change between 0.64 and 0.65, so the current implementation works for 0.64 (notation with logs instead of logs_statement), but it won't work for 0.66

It's just a configuration option, right? We will just need to update here it with the upgrade to 0.66. Please add a note comment

omrozowicz-splunk:
> > Another thing is that in transform processor was a syntax change between 0.64 and 0.65, so the current implementation works for 0.64 (notation with logs instead of logs_statement), but it won't work for 0.66
> 
> It's just a configuration option, right? We will just need to update here it with the upgrade to 0.66. Please add a note comment

Done

UPGRADING.md
```diff
@@ -1,5 +1,39 @@
 # Upgrade guidelines
 
+## 0.64.0 to 0.66.0
+Before this change, [k8s events receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8seventsreceiver) was used to collect Kubernetes Events.
+Now we utilize [k8s objects receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver) that can pull or watch any object from Kubernetes API server.
+Therefore, `clusterReceiver.eventsEnabled` is now deprecated, and to maintain the same behavior you should set
```
rmfitzpatrick, line 6: https://stackoverflow.com/a/9208164

It appears these changes remove `clusterReceiver.eventsEnabled` functionality in total, making it obsolete. Is this correct?

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -395,3 +395,11 @@ compatibility with the old config group name: "otelK8sClusterReceiver".
 {{- $clusterReceiver.k8sEventsEnabled }}
 {{- end }}
 {{- end -}}
+
+{{/*
+Whether clusterReceiver should be deployed
+*/}}
+{{- define "splunk-otel-collector.clusterReceiverDeploy" -}}
```
rmfitzpatrick, line 402: grammatically clearer and follows the existing convention

```suggestion
{{- define "splunk-otel-collector.clusterReceiverEnabled" -}}
```

UPGRADING.md
omrozowicz-splunk: Yes, right, k8s object receiver is not providing exactly the same functionality as k8s events receiver. Updated.

rmfitzpatrick: So https://github.com/signalfx/splunk-otel-collector-chart/pull/588#discussion_r1036766723 could also be expanded so that we provide an actual deprecation path by using the objects receiver w/ just the watched events for this value instead of obsoleting in one pass. The warning check also needs to be updated with a description of the behavior.

imo users should never have their config become "broken" during an upgrade and we should provide at least a release of warnings and ideally remediation logic in helpers (cc @dmitryax)

omrozowicz-splunk: Ok, so I updated it in the way that when `eventsEnabled` is set to true we return warning message and event objects are watched, so behaviour is similar (but not the same, as previous event formatting was different, and the sourcetype was `kube:events` and now it is `kube:object<resource_id>`. It is described in details in the upgrading doc.

test/.pytest.expect
```diff
@@ -0,0 +1,3 @@
+pytest-expect file v1
```
omrozowicz-splunk, line 1: This expect fail is here because of https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/16543 which is not yet merged

UPGRADING.md
```diff
@@ -1,5 +1,53 @@
 # Upgrade guidelines
 
+## 0.64.0 to 0.66.0
```
rmfitzpatrick, line 3: Would be 0.67.0 if part but maybe a version placeholder makes more sense.

UPGRADING.md
```diff
@@ -1,5 +1,53 @@
...
+Before this change, [k8s events receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8seventsreceiver) was used to collect Kubernetes Events.
+Now we utilize [k8s objects receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver) that can pull or watch any object from Kubernetes API server.
```
rmfitzpatrick, line 5:
```suggestion
Now we utilize the [Kubernetes Objects Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver) that can pull or watch any object from Kubernetes API server.
```

UPGRADING.md
```diff
@@ -1,5 +1,53 @@
...
+* k8s.event.start_time
+* k8s.event.name
+* k8s.event.uid
+* k8s.namespace.name
+
+In case of `k8sobjects`, the whole payload is stored in the log body and `object.message` refers to the event message.
+
+
+You can monitor more kubernetes objects configuring `clusterReceiver.k8sObjects` according to the instruction from
+[k8s objects receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver).
```
rmfitzpatrick, line 46:
```suggestion
You can monitor more Kubernetes objects configuring by `clusterReceiver.k8sObjects` according to the instructions from the
[Kubernetes Objects Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver) documentation.
```

UPGRADING.md
```diff
@@ -1,5 +1,53 @@
...
+* k8s.object.api_version
+* k8s.object.resource_version
+* k8s.event.reason
+* k8s.event.action
+* k8s.event.start_time
+* k8s.event.name
+* k8s.event.uid
+* k8s.namespace.name
+
+In case of `k8sobjects`, the whole payload is stored in the log body and `object.message` refers to the event message.
```
rmfitzpatrick, line 42:
```suggestion
Now with the `k8sobjects` receiver, the whole payload is stored in the log body and `object.message` refers to the event message.
```

Mind showing an example message?

UPGRADING.md
```diff
@@ -1,5 +1,53 @@
...
+* k8s.object.uid
+* k8s.object.fieldpath
+* k8s.object.api_version
+* k8s.object.resource_version
+* k8s.event.reason
+* k8s.event.action
+* k8s.event.start_time
+* k8s.event.name
+* k8s.event.uid
+* k8s.namespace.name
```
rmfitzpatrick, line 40:
```suggestion
There are differences in the log record formatting of events between the previous `k8s_events` receiver and the now adopted `k8sobjects` receiver. The `k8s_events` receiver stores event messages their log body, with the following fields added as attributes:

* `k8s.object.kind`
* `k8s.object.name`
* `k8s.object.uid`
* `k8s.object.fieldpath`
* `k8s.object.api_version`
* `k8s.object.resource_version`
* `k8s.event.reason`
* `k8s.event.action`
* `k8s.event.start_time`
* `k8s.event.name`
* `k8s.event.uid`
* `k8s.namespace.name`
```

rmfitzpatrick:
> k8s_event payload image
> 
> k8sobjects payload image

Looking at this more, as a user I would find this migration less than desirable and potentially fully breaking. The k8sobjects receiver could benefit from following the otel semantic conventions and maybe contribute new ones: https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/k8s.yaml.

Given the substance of the changes could it make sense to not alter the events receiver usage at all but have an opt-in for the new objects receiver?

UPGRADING.md
omrozowicz-splunk: Changed it to `$CURRENT_VERSION to $NEXT_VERSION`. Same in NOTES.txt

dmitryax:
> Given the substance of the changes could it make sense to not alter the events receiver usage at all but have an opt-in for the new objects receiver?

We can leave it with opt-in for now, but I believe we eventually need to deprecate the k8sevents receiver in favor of the k8sobjects receiver. The reason is that I don't think we should parse all the k8s event fields and compose attributes from it that can go with unnecessary high cardinality. Sending events in the format defined by the emitting system actually fits better the new events API that is currently being defined in the OTel spec. We should move this discussion to an issue in contrib. We will work on that with @harshit-splunk

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -35,6 +35,11 @@ receivers:
     {{- if eq (include "splunk-otel-collector.distribution" .) "openshift" }}
     distribution: openshift
     {{- end }}
+  {{- if and $clusterReceiver.objectsEnabled "true" (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
+  k8sobjects:
+    auth_type: serviceAccount
+    objects: {{ .Values.clusterReceiver.k8sObjects | toJson }}
```
rmfitzpatrick, line 41: Should this use `toYaml`? It would probably be worth adding a new manifest run w/ values that exercise this to help vet the output.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -431,6 +431,57 @@ clusterReceiver:
   # Once the receiver is stabilized, it'll be enabled by default in this helm chart
   eventsEnabled: false
 
+  # This flag enables Kubernetes objects collection using OpenTelemetry Kubernetes Object Receiver
+  # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver
+  # This option requires `logsEnabled` to be set to `true` for either `splunkObservability` or `splunkPlatform`
+  # depending on where you want to send the events. Otherwise, this option will not have any effect.
+  # The receiver currently is in alpha state which means that events format might change over time.
+  # Once the receiver is stabilized, it'll be enabled by default in this helm chart
+  objectsEnabled: false
```
dmitryax, line 440: Do we need this option? Can we use `k8sObjects` just instead given that it's empty by default?

omrozowicz-splunk: Ok, I deleted `objectsEnabled`

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
rmfitzpatrick: I'm not sure so would defer to whatever makes sense to your use case. thanks for adding the rendered manifests.

test/.pytest.expect
dmitryax: Is this PR functional to be merged with the current `sourcetype`? If so, can we add `test_k8s_objects_sourcetype` later instead?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -123,9 +123,9 @@ splunkObservability:
   tracesEnabled: true
   logsEnabled: false
 
-  # Option to send Kubernetes events to Splunk Observability Infrastructure Monitoring as data events:
+  # Option to send Kubernetes objects to Splunk Observability Infrastructure Monitoring as data events:
   # https://docs.splunk.com/Observability/alerts-detectors-notifications/view-data-events.html
-  # To send Kubernetes events to Splunk Observability Log Observer, set both clusterReceiver.eventsEnabled and
+  # To send Kubernetes events to Splunk Observability Log Observer, set both clusterReceiver.objectsEnabled and
   # splunkObservability.logsEnabled to true.
```
dmitryax, line 129:
```suggestion
  # To send Kubernetes events to Splunk Observability Log Observer, configure clusterReceiver.k8sObjects
  # and set splunkObservability.logsEnabled to true.
```

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -123,9 +123,9 @@ splunkObservability:
   tracesEnabled: true
   logsEnabled: false
 
-  # Option to send Kubernetes events to Splunk Observability Infrastructure Monitoring as data events:
+  # Option to send Kubernetes objects to Splunk Observability Infrastructure Monitoring as data events:
```
dmitryax, line 126:
```suggestion
  # Option to send Kubernetes events to Splunk Observability Infrastructure Monitoring as data events:
```

dmitryax (commented):
Overall, looks good to me. just couple nits

helm-charts/splunk-otel-collector/values.yaml
omrozowicz-splunk: Done

helm-charts/splunk-otel-collector/values.yaml
omrozowicz-splunk: Done

test/.pytest.expect
omrozowicz-splunk: Currently sourcetypes render correctly except `watch` case. So the only test case which is failing now is the one checking 
```
    - name: events
      mode: watch
```
Yes, we can add it later. I deleted pytest expect and `("sourcetype", "kube:object:event", 1),` from `test_k8s_objects_sourcetype`

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -395,3 +395,33 @@ compatibility with the old config group name: "otelK8sClusterReceiver".
...
+{{- printf "false" }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Whether object collection by k8s object receiver or/and event collection by k8s event receiver is enabled
+*/}}
+{{- define "splunk-otel-collector.objectsOrEventsEnabled" -}}
+{{- $clusterReceiver := fromYaml (include "splunk-otel-collector.clusterReceiver" .) }}
+{{- or $clusterReceiver.eventsEnabled ( eq (include "splunk-otel-collector.objectsEnabled" .) "true") -}}
```
dmitryax, line 417:
```suggestion
{{- or $clusterReceiver.eventsEnabled (eq (include "splunk-otel-collector.objectsEnabled" .) "true") -}}
```

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
```diff
@@ -1,5 +1,5 @@
 {{ $clusterReceiver := fromYaml (include "splunk-otel-collector.clusterReceiver" .) }}
-{{- if and $clusterReceiver.enabled (eq (include "splunk-otel-collector.metricsEnabled" .) "true") }}
+{{ if eq (include "splunk-otel-collector.clusterReceiverEnabled" . ) "true" }}
```
dmitryax, line 2:
```suggestion
{{ if eq (include "splunk-otel-collector.clusterReceiverEnabled" .) "true" }}
```

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -395,3 +395,33 @@ compatibility with the old config group name: "otelK8sClusterReceiver".
...
+{{/*
+Whether object collection by k8s object receiver is enabled
+*/}}
+{{- define "splunk-otel-collector.objectsEnabled" -}}
+{{- $clusterReceiver := fromYaml (include "splunk-otel-collector.clusterReceiver" .) }}
+{{- if gt (len $clusterReceiver.k8sObjects) 0 }}
+{{- printf "true" }}
+{{- else }}
+{{- printf "false" }}
+{{- end -}}
```
dmitryax, line 409: I believe this can be simplified to 
```suggestion
{{- gt (len $clusterReceiver.k8sObjects) 0 }}
```

dmitryax (approved):
LGTM. Just a couple more nits, and I think we can merge it. Thanks for all the work!

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -231,7 +247,7 @@ service:
         {{- end }}
     {{- end }}
 
-    {{- if and $clusterReceiver.eventsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
+    {{- if and $clusterReceiver.eventsEnabled  (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
```
rmfitzpatrick, line 250: nit, no change needed

rmfitzpatrick (approved):

PR #596, opened 2022-12-08
Fix/fix security issues

wojtekzyla:
This PR ensures that:

- containers don't disable default seccomp profile, 
- containers mount root directory as read only,
- restrict container from acquiring additional privileges.


wojtekzyla:
It is not working in the non-root mode

dmitryax:
The container has to write to the host filesystem for checkpointing. There is an option to do it as non-root user, but it'll spin up another init container to give the user write permissions to the checkpointing directories. This option can be used for that https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/values.yaml#L561

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -419,6 +424,10 @@ clusterReceiver:
   terminationGracePeriodSeconds: 600
   priorityClassName: ""
 
+  # Container configurations
+  container:
```
harshit-splunk, line 428: No need to add container-level securityContext. All the container by default inherits pod-level securityContext.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -988,6 +997,10 @@ gateway:
...
```
harshit-splunk, line 1001: Same as above, No need to add container-level securityContext.

wojtekzyla: I noticed that some things (I think it was readOnlyRootFilesystem and allowPrivilegeEscalation) can't be set at the pod level

harshit-splunk: No need to set `readOnlyRootFilesystem` as the cluster receiver and the gateway doesn't need to access the root directory.

harshit-splunk: Alternatively, you can use this securityContext at container level instead of pod level. I don't think we should maintain 2 different security context for same thing.

wojtekzyla: Ok, I deleted container securityContext configuration and left only the original pod securityContext configuration.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -350,13 +350,18 @@ agent:
       memory: 500Mi
 
   # To collect container logs and journald logs, it will run the agent as a root user.
-  # To run it as non root user, uncomment below `securityContext` options.
+  # To run it as non root user, change runAsUser and runAsGroup to non-zero value (e.g. 20000) and change runAsNonRoot to true.
```
harshit-splunk, line 353: Also mention here that running as non-root user doesn't work in openshift cluster

PR #599, opened 2022-12-08
fix network explorer imagePullSecret validation

dustin-bingham:


rmfitzpatrick:
Would it make sense to maintain the value type and use a similar template as this? https://github.com/signalfx/splunk-otel-collector-chart/blob/b096518ddde2deeecc3c942da5bc7d4262da977a/helm-charts/splunk-otel-collector/templates/serviceAccount.yaml#L6

Exercising this in a rendered manifest could be helpful as well depending on how readily applied you'd like them to be.

dustin-bingham:
Yes, I think it would. I tried to leave the network-explorer templates
close to the originals for easier comparison and to make the transition
smooth for our existing customers (but as time goes on, that will have less
value).  So if you think it's worth doing now, we can do that - or we can
just fix the immediate bug and I can "collapse" the imagePullSecrets in a
future MR.  What do you think?

rmfitzpatrick:
If I'm following correctly, accepting an array of strings is a preferable interface over requiring an array of the reference objects and is already the expected one despite not having a description or example in https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/advanced-configuration.md.

Maybe you could define a new `networkExplorerImagePullSecrets` helper that requires minimal changes to the existing templates?

dustin-bingham:
I went ahead with your first suggestion after all :) Let me know if this is
closer to what you are thinking.  Thanks!

rmfitzpatrick (approved):

PR #607, opened 2022-12-14
Fix/read permission journal log

wojtekzyla:
This PR will enable collecting container logs and journald logs when agent is run as a non-root user.


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -355,8 +355,8 @@ agent:
   # permissions of container logs directories on the host filesystem to make logs readable by this non root user.
 
   securityContext: {}
-  #   runAsUser: 20000
-  #   runAsGroup: 20000
+    # runAsUser: 20000
+    # runAsGroup: 20000
```
atoulme, line 359: probably can revert that change

wojtekzyla: Sure, I reverted that

PR #615, opened 2022-12-22
Add splunkPlatform.sendingQueue.storage config options to values.yaml

aryznar-splunk:


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -526,6 +528,7 @@ logsCollection:
     index: ""
 
   checkpointPath: "/var/addon/splunk/otel_pos"
+  persistentStorageDirectory: "persistent_storage"
```
dmitryax, line 531: I don't think we should put it in checkpointing subdirectory. Maybe `persistentQueueStoragePath: /var/addon/splunk/persistent_queue` ?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -95,6 +95,8 @@ splunkPlatform:
     #   num_seconds is the number of seconds to buffer in case of a backend outage
     #   requests_per_second is the average number of requests per seconds.
     queueSize: 5000
+    # When set, enables persistence and uses the component specified as a storage extension for the persistent queue
+    storage: true
```
dmitryax, line 99: Given that we've been reproducing the exporter configuration layout here, I think we shouldn't override the same configuration with a different value type to avoid confusion. I mean that `storage` is a string pointing to a storage extension in the exporter interface, not a boolean flag. 

Maybe we add this interface instead:

```
splunkPlatform:
  persistentQueueEnabled:
    logs: false
    metrics: false
```

WDYT?

helm-charts/splunk-otel-collector/values.yaml
aryznar-splunk: It was also considered.I will implement this feedback in the next year :). Marry Christmas and Happy New Year!!!

helm-charts/splunk-otel-collector/values.yaml
aryznar-splunk: good idea will do.

helm-charts/splunk-otel-collector/values.yaml
dmitryax: Merry Christmas and Happy New Year! :)

harshit-splunk:
Closing the PR as @aryznar-splunk will not be able to work on this. I'll create a new PR to add this feature.

PR #617, opened 2023-01-03
Fix/seccurity issues fix

wojtekzyla:
This PR will enable running fluentd and otel containers with the following securityContext settings:

```
securityContext:
    readOnlyRootFilesystem: true
    allowPrivilegeEscalation: false
    seccompProfile:
      type: RuntimeDefault
    runAsUser: 20000
    runAsGroup: 20000
    runAsNonRoot: true
```


harshit-splunk (changes requested):

crobert-1:
Can a little more context for this PR be provided? Pardon my ignorance, but I don't have any context other than what's included here. What do the new settings allow that wasn't allowed before?

harshit-splunk:
@crobert-1, one of our customers has detected compliance alerts from their security scans.
The aim of this PR is to address some of those alerts that can be resolved.


**Compliance alerts that can be resolved**
| Finding                                                 | Container               | Reference                                                                                                 |
|---------------------------------------------------------|-------------------------|-----------------------------------------------------------------------------------------------------------|
| Container is running as root                            | fluentd                 | https://www.tenable.com/audits/items/CIS_Kubernetes_v1.3.0_Level_2.audit:65cc48471d64b35904d4742905011dbe |
| Mount container's root filesystem as read only          | fluentd                 | https://www.tenable.com/audits/items/CIS_Docker_1.11.0_v1.0.0_L1.audit:6752d9423f1e5b59b5db7e01a30ae764   |
| Restrict container from acquiring additional privileges | fluentd, otel-collector | https://www.tenable.com/audits/items/CIS_Docker_1.13.0_L1_v1.0.0.audit:ad29f1459362152272392154b60e43bf   |
| Do not disable default seccomp profile                  | fluentd, otel-collector | https://www.tenable.com/audits/items/CIS_Docker_1.13.0_L1_v1.0.0.audit:50147be1f26fcf1120989f6c006f3cc0   |

**Compliance alerts that cannot be resolved**
| Finding                                                      | Container      | Reference                                                                                               | Reason                                                                                                                               |
|--------------------------------------------------------------|----------------|---------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------|
| Do not mount sensitive host system directories on containers | otel-collector | https://www.tenable.com/audits/items/CIS_Docker_1.11.0_v1.0.0_L1.audit:abec2c62679e1ddd15b272dd6b8aa26d | Required to collect host metrics. Mounted as readonly                                                                                |
| Do not share the host's network namespace                    | otel-collector | https://www.tenable.com/audits/items/CIS_Docker_1.13.0_L1_v1.0.0.audit:0a426209222f96d306a4758cc8e2376a | It is required to collect the underlying node's networking metrics (https://github.com/signalfx/splunk-otel-collector-chart/pull/37) |
| Do not share the host's UTS namespace                        | otel-collector | https://www.tenable.com/audits/items/CIS_Docker_1.13.0_L1_v1.0.0.audit:4011a2f1faca2ba2a8e2a61d3463ceeb | It is required to collect the underlying node's networking metrics (https://github.com/signalfx/splunk-otel-collector-chart/pull/37) |

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -350,13 +350,19 @@ agent:
       memory: 500Mi
 
   # To collect container logs and journald logs, it will run the agent as a root user.
-  # To run it as non root user, uncomment below `securityContext` options.
+  # To run it as non root user, change runAsUser and runAsGroup to non-zero value (e.g. 20000) and change runAsNonRoot to true.
   # Setting runAsUser and runAsGroup to a non root user enables an init container that patches group
   # permissions of container logs directories on the host filesystem to make logs readable by this non root user.
+  # NOTE: during internal tests running this container as a non-root user with cri-o runntime didn't work.
```
crobert-1, line 356: How impactful is this? Do many users run this with cri-o runtime? If this is impactful is there an issue filed to follow up and resolve?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -557,8 +563,20 @@ fluentd:
       cpu: 100m
       memory: 200Mi
 
+# To run fluentd container as a non root user and enable other seccurity features,
```
crobert-1, line 566: Nit: typo

s/seccurity/security

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -350,13 +350,19 @@ agent:
       memory: 500Mi
 
   # To collect container logs and journald logs, it will run the agent as a root user.
-  # To run it as non root user, uncomment below `securityContext` options.
+  # To run it as non root user, change runAsUser and runAsGroup to non-zero value (e.g. 20000) and change runAsNonRoot to true.
```
crobert-1, line 353: So this security fix is to allow fluentd to run as non-root, but default is still root? Is that correct?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -557,8 +563,20 @@ fluentd:
       cpu: 100m
       memory: 200Mi
 
+# To run fluentd container as a non root user and enable other seccurity features,
+# delete runAsUser: 0 and uncomment other securityContext settings
+# NOTE: during internal tests running this container as a non-root user with cri-o runntime didn't work.
```
crobert-1, line 568: What kind of testing has been done running fluentd as non-root? I'm glad at least some kind of error popped up in testing, anytime something is changed to non-root permissions I'd be pretty concerned about things breaking.

crobert-1 (commented):
Thanks @harshit-splunk for the explanation, this makes a lot more sense now!

helm-charts/splunk-otel-collector/values.yaml
harshit-splunk: Yes, Openshift by default uses CRI-O runtime. So, it is widely used.

helm-charts/splunk-otel-collector/values.yaml
harshit-splunk: > So this security fix is to allow fluentd to run as non-root, but default is still to run as root? Is that correct?

Yes. The default value should run on any environment. If we change it to run as non-root by default, it will be broken for CRI-O runtime.

> Also, whether or not it's run as root, the default path to write checkpoints is no longer under /var/log, right?

Correct

helm-charts/splunk-otel-collector/values.yaml
harshit-splunk: We have tested on Minikube with different runtimes (i.e. docker, containerd and cri-o). Also, tested it on [rosa](https://aws.amazon.com/rosa/).

We have followed these steps to test running as a non-root user on each runtime:
1.  Deploy. the collector
2. Deploy datagen pod
3. Verify datagen logs are collected in splunk

Since we have changed the default checkpoint path for fluentd, we have tested the following scenarios:
- fluentd(old version) -> fluentd(current branch)
- fluentd(current branch) -> otel(current branch)
- fluentd(old version) -> otel(current branch)

For the above scenarios, we have validated that it reads the correct checkpoints and continues reading the data from where it left

helm-charts/splunk-otel-collector/values.yaml
crobert-1: So for now, most users shouldn't use the functionality of this PR, right? The belief is also that the problem is in CRI-O, and nothing to do with our side, correct? Just trying to make sure I understand properly.

If these things are the case it should be made pretty clear to customers. We don't want anyone trying to actually use this if we know it's broken.

harshit-splunk: > So for now, most users shouldn't use the functionality of this PR, right? 

Yes, we can make it more clear who should not enable this feature

>  The belief is also that the problem is in CRI-O, and nothing to do with our side.

Correct. I can brief you about what is expected and what is going wrong.

ACL(Access control lists) can add read, write or execute permission to specified users or groups.
When the patch-log-directory container is executed, it applied default ACLs to /var/log directory. So, when any new file/directory is created inside /var/log directory, it will inherit those ACLs. It is working as expected in docker/containerd runtime. But, not working in cri-o runtime.

```bash
bash-4.4# getfacl /var/log/pods/default_sck-splunk-otel-collector-agent-nspsk_ca2e984d-8338-4561-8337-d371d832f6af/otel-collector
getfacl: Removing leading '/' from absolute path names
# file: var/log/pods/default_sck-splunk-otel-collector-agent-nspsk_ca2e984d-8338-4561-8337-d371d832f6af/otel-collector
# owner: root
# group: root
user::rwx
group::r-x
group:20000:r-x
mask::r-x
other::r-x
default:user::rwx
default:group::r-x
default:group:20000:r-x
default:mask::r-x
default:other::r-x

bash-4.4# getfacl /var/log/pods/default_sck-splunk-otel-collector-agent-nspsk_ca2e984d-8338-4561-8337-d371d832f6af/otel-collector/0.log 
getfacl: Removing leading '/' from absolute path names
# file: var/log/pods/default_sck-splunk-otel-collector-agent-nspsk_ca2e984d-8338-4561-8337-d371d832f6af/otel-collector/0.log
# owner: root
# group: root
user::rw-
group::r-x                      #effective:---
group:20000:r-x                 #effective:---
mask::---
other::---
```

If I manually create a file there, it inherits corrects ACLs

```sh
bash-4.4# touch sample.txt
bash-4.4# getfacl /var/log/pods/default_sck-splunk-otel-collector-agent-nspsk_ca2e984d-8338-4561-8337-d371d832f6af/otel-collector/sample.txt 
getfacl: Removing leading '/' from absolute path names
# file: var/log/pods/default_sck-splunk-otel-collector-agent-nspsk_ca2e984d-8338-4561-8337-d371d832f6af/otel-collector/sample.txt
# owner: root
# group: root
user::rw-
group::r-x                      #effective:r--
group:20000:r-x                 #effective:r--
mask::r--
other::r--
```

crobert-1: That makes sense! Thanks for filing an issue against CRI-O as well!

I'll include the issue here as well just for future reference: https://github.com/cri-o/cri-o/issues/6519

crobert-1 (changes requested):
It looks like the CI/CD `pre-commit` failures are related to changes made in this PR, so those need to be fixed.

harshit-splunk:
> It looks like the CI/CD pre-commit failures are related to changes made in this PR, so those need to be fixed.

When I do `make render`, it splits `set(resource.attributes["com.splunk.sourcetype"], Concat(["kube:object:", attributes["event.name"]], ""))` into 2 different lines. Weird.

harshit-splunk:
pre-commit failure is related to https://github.com/signalfx/splunk-otel-collector-chart/pull/634

crobert-1 (approved):

helm-charts/splunk-otel-collector/templates/revert-patch-log-dirs-hook.yaml
```diff
@@ -0,0 +1,82 @@
...
+    fi;
+    if [ -d "/var/log/pods" ];
+    then
+        setfacl --recursive --remove-all  /var/log/pods;
+    fi;
+    {{- end }}
+    {{- if .Values.logsCollection.journald.enabled }}
+    if [ -d "{{ .Values.logsCollection.journald.directory }}" ];
+    then
+        setfacl --recursive --remove-all d:m::rx,m::rx,d:g:{{ .Values.agent.securityContext.runAsGroup | default 20000 }}:rx,g:{{ .Values.agent.securityContext.runAsGroup | default 20000 }}:rx {{ .Values.logsCollection.journald.directory }};
```
rmfitzpatrick, line 41: Why is the default 20000 and not the splunk-otel-collector uid? https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/Dockerfile#L14. This would prevent non-native Smart Agent receiver functionality.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -662,7 +683,7 @@ fluentd:
     # Prefix for pos_file tail source parameter
     # Can be used if you want to run multiple instances of fluentd on the same host
     # https://docs.fluentd.org/input/tail#pos_file-highly-recommended
-    posFilePrefix: /var/log/splunk-fluentd
+    posFilePrefix: /var/addon/splunk/fluent_pos/splunk-fluentd
```
rmfitzpatrick, line 686: Is `/var/run` a more conventional place for uses like this?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -350,13 +350,19 @@ agent:
...
-  #   runAsUser: 20000
-  #   runAsGroup: 20000
+  # NOTE: Running this container as a non-root user doesn't work with openshift cluser or cri-o runntime.
+  # See https://github.com/cri-o/cri-o/issues/6519 for more details.
+  securityContext:
+    readOnlyRootFilesystem: true
+    allowPrivilegeEscalation: false
+    seccompProfile:
+      type: RuntimeDefault
+    runAsUser: 0
```
rmfitzpatrick, line 363: Sorry I'm not following why we are now explicitly running as root?

rendered/manifests/agent-only/daemonset.yaml
```diff
@@ -79,6 +79,14 @@ spec:
           protocol: TCP
         image: quay.io/signalfx/splunk-otel-collector:0.67.0
         imagePullPolicy: IfNotPresent
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
```
rmfitzpatrick, line 84: Do you know offhand if this will this prevent writing to any non-volume directory?

helm-charts/splunk-otel-collector/values.yaml
wojtekzyla: By default this should run on any environment. Running as a non-root with CRI-O runtime causes crash.

helm-charts/splunk-otel-collector/templates/revert-patch-log-dirs-hook.yaml
wojtekzyla: Here we are deleting ACLs set by the init container. Currently 20000 is the default for user in the init container. This can be changed if it's preventing any functionalities.

helm-charts/splunk-otel-collector/values.yaml
wojtekzyla: Otel checkpoint is already stored in `/var/addon/splunk/` directory so the same convention was used here.

rendered/manifests/agent-only/daemonset.yaml
wojtekzyla: Yes, testing showed that it prevents write operation also to non-volume directories

harshit-splunk (approved):

rendered/manifests/agent-only/daemonset.yaml
rmfitzpatrick: This would be a breaking change for the smart agent receiver and collectd-based monitors

PR #638, opened 2023-01-19
Revert "Fix/security issues fix (#617)"

rmfitzpatrick:
a9f24c089c57be8195f24ecc9ced59ad68e27abc introduces at least one breaking change that will prevent collectd-based Smart Agent receiver instances from working and will cause the collector to error out:

```
Error: cannot start pipelines: failed to create collectd config dir at /usr/lib/splunk-otel-collector/agent-bundle/run/collectd/global/managed_config: mkdir /usr/lib/splunk-otel-collector/agent-bundle/run/collectd/global: read-only file system
2023/01/19 16:02:08 main.go:111: application run finished with error: cannot start pipelines: failed to create collectd config dir at /usr/lib/splunk-otel-collector/agent-bundle/run/collectd/global/managed_config: mkdir /usr/lib/splunk-otel-collector/agent-bundle/run/collectd/global: read-only file system
```

There are also some other concerns about its effects and design that I think would be best discussed in a GH issue. Any immediate concerns about the fluentd init containers or agent security context should already be configurable w/ values as is.


rmfitzpatrick:
cc @wojtekzyla

atoulme (approved):

Issue #651, opened 2023-01-31
OpenShift Install helm chart SCC problem - missing required field "allowHostDirVolumePlugin" in io.openshift.security.v1.SecurityContextConstraints


adamjk-dev:
Recipient,

Using chart version 0.68.1, having issues doing an install on a RedHat OpenShift cluster.
OCP version: 4.10.26

**Helm install command:**
```
ns='splunk'
ver_dir='v0681'
ver='0.68.1'
helm install splunk-otel-collector charts/${ver_dir}/ -n=${ns} --version ${ver} -f charts/${ver_dir}/overrides.yaml
```

**overrides.yaml:**
```
---
distribution: "openshift"

clusterName: "my_cluster"

logsEngine: otel

splunkPlatform:
  # Required for Splunk Enterprise/Cloud. URL to a Splunk instance to send data
  # to. e.g. "http://X.X.X.X:8088/services/collector". Setting this parameter
  # enables Splunk Platform as a destination.
  endpoint: "<<REDACTED>>"
  # Required for Splunk Enterprise/Cloud (if `endpoint` is specified). Splunk
  # HTTP Event Collector token.
  token: "<<REDACTED>>"

  # Name of the Splunk event type index targeted. Required when ingesting logs to Splunk Platform.
  index: "index_name"
  # Name of the Splunk metric type index targeted. Required when ingesting metrics to Splunk Platform.
  metricsIndex: ""
  # Optional. Default value for `source` field.
  source: "kubernetes"
  # Optional. Default value for `sourcetype` field. For container logs, it will
  # be container name.
  sourcetype: ""
  # Maximum HTTP connections to use simultaneously when sending data.
  maxConnections: 200
  # Whether to disable gzip compression over HTTP. Defaults to true.
  disableCompression: true
  # HTTP timeout when sending data. Defaults to 10s.
  timeout: 10s
  # Whether to skip checking the certificate of the HEC endpoint when sending
  # data over HTTPS.
  insecureSkipVerify: true
  # The PEM-format CA certificate for this client.
  # NOTE: The content of the certificate itself should be used here, not the
  #       file path. The certificate will be stored as a secret in kubernetes.
  clientCert: ""
  # The private key for this client.
  # NOTE: The content of the key itself should be used here, not the file path.
  #       The key will be stored as a secret in kubernetes.
  clientKey: ""
  # The PEM-format CA certificate file.
  # NOTE: The content of the file itself should be used here, not the file path.
[...]

PR #643, opened 2023-01-23
Allow to overwrite default SCC with values.yaml

harshit-splunk:
When using `openshift` distribution, helm chart creates SecurityContextConstraints object, but it's content cannot be modified. With this PR, user can for example add `seccompProfiles` to the SCC with following config:

```yaml
securityContextConstraintsOverwrite:
  seccompProfiles:
  - runtime/default
```

PR #652, opened 2023-01-31
Fix SCC render issue

harshit-splunk:
Fixes https://github.com/signalfx/splunk-otel-collector-chart/issues/651

It fixes the bug introduced in https://github.com/signalfx/splunk-otel-collector-chart/pull/643


rmfitzpatrick (approved):
```
 users:
-- system:serviceaccount:default:default-splunk-otel-collectorallowHostDirVolumePlugin: true
+- system:serviceaccount:default:default-splunk-otel-collector
+allowHostDirVolumePlugin: true
```

PR #657, opened 2023-02-07
re-order k8sattributes processor to precede batch processor

dloucasfx:
based on this https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor#traces , `k8sattributes` should precede `batch` processor, otherwise the `from: connection` will be missing the IP info

Signed-off-by: Dani Louca <dlouca@splunk.com>


dmitryax (approved):

dmitryax:
Thanks for fixing it

PR #660, opened 2023-02-09
Fix naming convention issue in helm chart

vihas-splunk:
While setting `renameFieldsSck` to true, the `resource/logs` sets the fields as per the SCK convention and copies their values from otel fields. But there are some otel fields which are not set in the first place. 
So, move the `resource` above `resource/logs` as it sets all the required fields and the latter can then copy without any issues.  
Also, add `label_app` field, if `renameFieldsSck` is true

The SCK's field `container_image` is split into two parts as per Otel convention i.e. image_name and image_tag, so combine them if user has enabled `renameFieldsSck`


dmitryax:
This solves the problem only for a specific label fetched from the pods, fetching of that label appeared to be set by default configuration, but it might change in the future. Instead of hardcoding the rules, maybe we can just ask users to add additional configuration which would be?

```yaml
  fromLabels:
    - key: app
      tag_name: label_app
```

vihas-splunk:
> This solves the problem only for a specific label fetched from the pods, fetching of that label appeared to be set by default configuration, but it might change in the future. Instead of hardcoding the rules, maybe we can just ask users that want this label in SCK format to add additional configuration which would be?
> 
> ```yaml
>   fromLabels:
>     - key: app
>       tag_name: label_app
> ```

Yes, we can ask users but the problem is that the SCK-Otel wouldn't provide all the SCK labels as mentioned in https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/migration-from-sck.md#extracted-fields-for-logs

There are a couple of fields missing, so I guess it's better to add those fields in our processor with existing fields.

dmitryax:
Ok. Feel free to keep it, but please make sure it's only applicable if `Values.splunkPlatform.fieldNameConvention.keepOtelConvention` is set to true. `container_image` should not be added by default

vihas-splunk:
> Ok. Feel free to keep it, but please make sure it's only applicable if `Values.splunkPlatform.fieldNameConvention.keepOtelConvention` is set to true. `container_image` should not be added by default

Took care of that.

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -151,12 +154,20 @@ resource/logs:
...
       action: delete
     {{- end }}
     {{- end }}
     {{- end }}
+  
+transform/sck_fields:
+  log_statements:
+    - context: log
+      statements:
+        - set(resource.attributes["container_image"], Concat([resource.attributes["container.image.name"], resource.attributes["container.image.tag"]], ":"))
```
dmitryax, line 170: This should not be added if  `.Values.splunkPlatform.fieldNameConvention.renameFieldsSck` is false

vihas-splunk: Sorry about that, I earlier added that part in _otel_agent.tpl. I have updated the PR ;)

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -151,12 +154,22 @@ resource/logs:
...
       action: delete
     {{- end }}
     {{- end }}
     {{- end }}
+
+transform/sck_fields:
+  log_statements:
+    - context: log
+      statements:
+        {{- if .Values.splunkPlatform.fieldNameConvention.renameFieldsSck }}
```
dmitryax, line 170: You need to apply the condition to the whole receiver and its reverence in the pipeline

vihas-splunk: I agree, but the reason why added it here is that if someone wants to add anything to the transform processor in the future, it would be easier to do. 
I have also made transform/logs in a similar fashion to resource/logs.

dmitryax: If users want to add another processor and configure it, it's easy to define in the `agent.config` section. We should not add components with an empty body just to make them visible, it's a lot of different components available

vihas-splunk: Yes, they can, but many customers will be migrating from SCK to SCK-Otel as SCK will be deprecated soon [https://github.com/splunk/splunk-connect-for-kubernetes#end-of-support].
That's why I'm adding this new processor. The only reason for adding a new processor is that to make SCK-Otel backward compatible with SCK. For other stuff, users can make use of `agent.config` as you suggested.

dmitryax: Please don't add a new processor in the real production config if it's not being used by default. Just mention how to configure it in the migration doc

examples/add-receiver-creator/rendered_manifests/configmap-agent.yaml
```diff
@@ -135,6 +135,10 @@ data:
         - system
         override: true
         timeout: 10s
+      transform/logs:
+        log_statements:
+        - context: log
+          statements: null
```
dmitryax, line 141: To clarify my previous comment, this should not be rendered by default. Config should always have components that are used in the pipelines, otherwise it brings confusion for users and maintainers

vihas-splunk: agree with your point.
I have updated the PR and the processor wouldn't be rendered by default.

dmitryax (approved):

dmitryax:
@vihas-splunk please run `make render`

vihas-splunk:
> @vihas-splunk please run `make render`

Done, can you approve the workflow? I can go ahead and merge after that
I've verified that pre-commit checks are passing on my local machine.

vihas-splunk:
@dmitryax PR looks good to merge

PR #663, opened 2023-02-09
Add examples for supported Kubernetes distributions and Kubernetes clusters with windows nodes

jvoravong:
Adding more examples for usability. 

I'm not going to include the rendered manifests for the examples until after I get the base example content approved so reviewers are not blinded by a +6k change set. So I will have one more commit after I get approvals.

### DIFF
--- a/docs/migration-from-sck.md
+++ b/docs/migration-from-sck.md
@@ -22,8 +22,6 @@ Splunk OpenTelemetry Collector for Kubernetes has the following components and a
 2. Splunk OpenTelemetry Collector Cluster Receiver (`clusterReceiver`) to fetch metrics from a Kubernetes API (deployed as a Kubernetes 1-replica Deployment)
 3. Optional Splunk OpenTelemetry Collector Gateway (`gateway`) to forward data through it to reduce load on Kubernetes API and apply additional processing (deployed as a Kubernetes Deployment)
 
-There is no application available for fetching Kubernetes objects metadata from a Kubernetes cluster.
-
 ### Changes in logs in Splunk OpenTelemetry Collector for Kubernetes
 
 * Red Hat Universal Base Image (UBI) Docker images for our applications are no longer available, as we now use scratch images.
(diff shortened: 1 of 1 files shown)


examples/distribution-vanilla/README.md
```diff
@@ -0,0 +1,6 @@
+# Example of chart configuration
+
+## How to install the collector on a Vanilla (unmodified) Kubernetes distribution
+
+In this example we will show how to install the collector on this Kubernetes
+distribution.
```
dmitryax, line 6: Maybe something like this?

```suggestion
This example works on any cloud providers and distributions but won't collect related metadata. This should be installed on self-hosted Kubernetes clusters.
```

examples/distribution-vanilla/distribution-vanilla-values.yaml
```diff
@@ -0,0 +1,7 @@
+clusterName: CHANGEME
+splunkObservability:
+  realm: CHANGEME
+  accessToken: CHANGEME
+
+distribution: ""
+cloudProvider: ""
```
dmitryax, line 7: Should be removed if we rename it to default values

examples/distribution-vanilla/README.md
dmitryax: did you miss pushing your changes?

examples/distribution-gke/README.md
```diff
@@ -0,0 +1,6 @@
+# Example of chart configuration
+
+## How to install the collector on GKE
+
+In this example we will show how to install the collector on this Kubernetes
+distribution.
```
dmitryax, line 6: Maybe something like that to avoid copy-pasting?

`distribution-gke-values.yaml` file should be used to install the collector in Google Kubernetes Engine (GKE).

atoulme (approved):

examples/distribution-gke/README.md
dmitryax: Ok sounds good

dmitryax (approved):

PR #665, opened 2023-02-09
Change networkExplorer.kernelCollector.useDockerMetadata default value to false.

jimwsplk:


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1126,8 +1126,8 @@ networkExplorer:
     # uncomment the line below to disable automatic kernel headers fetching
     # fetchKernelHeaders: false
 
-    # assumes docker is being used for kubernetes containers
-    useDockerMetadata: true
+    # uncomment to enable enrichment using Docker metadata
+    # useDockerMetadata: true
```
jimwsplk, line 1130: Simply changed the default value for NetworkExplorer useDockerMetadata.  (It defaults to false here: https://github.com/jimwsplk/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/network-explorer/kernel-collector-daemonset.yaml#L46)

Tested and validated deploying with this change in my dev environment.

PR #668, opened 2023-02-16
Fix/persistent queue

wojtekzyla:
Things included in this PR:

- enable persistent queues configuration in the helm chart
- while creating ACLs in the initContainer in the daemonset, change the default user and group of the otel image to 999
- add post delete hook which deletes ACLs set for directories used by agent, gateway and clusterReceiver while those pods run as non-root users

PR #675, opened 2023-02-27
Add persistent queues

omrozowicz-splunk:
Things included in this PR:

1. enable persistent queues configuration in the helm chart
2. post delete hook which deletes ACLs set for directories used by agent, gateway and clusterReceiver while those pods run as non-root users

Continuation of this PR: https://github.com/signalfx/splunk-otel-collector-chart/pull/668


dmitryax:
@omrozowicz-splunk can you please split the PR separately for each problem?

PR #567, opened 2022-10-20
Make sure filelog receiver uses file_storage for checkpointing

harshit-splunk:
Fixes https://github.com/signalfx/splunk-otel-collector-chart/issues/564

`storage` parameter wasn't configured. So, it wasn't reading the checkpoint at all.

PR #680, opened 2023-03-03
Make journald receiver cursors persistent

vinzent:
Currently migrated fluentd checkpoint files are not used due to missing storage param to the journald

According docs cursors/positions/checkpoints are only persisted on disk, if storage param is given. See https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/journaldreceiver/README.md

Equivalent to filelog change in #567


dmitryax (approved):

dmitryax:
@vinzent, thanks for the fix!

PR #687, opened 2023-03-05
Fix acl permissions

wojtekzyla:
Things included in this PR:

- while creating ACLs in the initContainer in the daemonset, change the default user and group of the otel image to 999 (this is because splunk-otel-collector has uid of 999)

- add post delete hook which deletes ACLs set for directories used by agent, gateway and clusterReceiver while those pods run as non-root users

This is a separate PR for some changes made in #668 . Those changes were originally included in #617 however, that PR was reverted.


dmitryax:
Please link an issue that this PR resolves. This needs clarification why 2000 -> 999 user change is required. Also, how do we handle upgrades for users that already run agents with 2000?

wojtekzyla:
> Please link an issue that this PR resolves. This needs clarification why 2000 -> 999 user change is required. Also, how do we handle upgrades for users that already run agents with 2000?

I tested this migration 2000 -> 999  and pods didn't crash, both logs and metrics were still sent properly and there was no data loss.

dmitryax (approved):
LGTM

kumachop2:
@wojtekzyla Thanks for making acl changes for init containers. We are using helm v0.72.0 deployed in AKS env, But Azure defender reported running container runAsUser as root should be avoided. Please correct me if any thing missing here.

Later made changes to ds template runAsUser value to 999 under initi container (
https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/daemonset.yaml#L108  & https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/daemonset.yaml#L165 ).

Post deployment Splunk agents wasn't initialized and here is error:
**Error from server (BadRequest): container "otel-collector" in pod "*****-splunk-otel-collector-agent-2rnnr" is waiting to start: PodInitializing**

Since we are not using fluentd I have made changes to acl (https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/daemonset.yaml#L165) like below to resolve container initialization issues with Splunk Agents pods.
              setfacl -n -Rm d:g:{{ $agent.securityContext.runAsGroup | default 999 }}:rx,g:{{ $agent.securityContext.runAsGroup | default 999 }}:rx /var/log/pods; 
**( discarded d:m::rx,m::rx,)**

I believe setacl command was changed part of helm chart v0.68.0. 

With Above changes Azure defender stopped reporting running container runAsUser as root should be avoided.

Please do review and assist me with refrain from making template changes or permanent fix for defender error.

wojtekzyla:
Hi @kumachop2 , init-containers must run as root user in order to set the ACLs on the host machine.  Also. this dm:rx,m::rx, that you mentioned was added, because without it those ACLs weren't set correctly while using some container runtimes.

kumachop2:
@wojtekzyla Thank you for prompt response. I would like to know the fix/steps to address Azure defender error on  init-containers  running as root user.

PR #701, opened 2023-03-21
OTL-1409 Add support for Operator based Java auto-instrumentation

jvoravong:
Description:
- With these changes I am proposing to add Operator based Java auto-instrumentation to this Helm chart as an experimental feature.
- Most of the rendered manifests can be ignored, I suggest only looking at the rendered manifests under the operator subdirectory. They are the only ones containing original content that is being added.
- The example and documentation being added only shows how to configure auto-instrumentation for a single and simple happy path use case. There are other ways we could setup auto-instrumentation that we should document in the future.
- More follow up PRs are to come.

New Documentation Links:
- https://github.com/jvoravong/splunk-otel-collector-chart/blob/OTL-1409/docs/auto-instrumentation-introduction.md
- https://github.com/jvoravong/splunk-otel-collector-chart/blob/OTL-1409/docs/auto-instrumentation-install.md
- https://github.com/jvoravong/splunk-otel-collector-chart/blob/OTL-1409/examples/enable-operator-and-auto-instrumentation/README.md

Testing:
- Tested locally and manually with the following specs.
  - minikube v1.29.0 on Darwin 13.2.1
  - hyperkit driver
  - Resources CPUs=4, Memory=8192MB, Disk=20000MB


examples/enable-operator-and-auto-instrumentation/README.md
```diff
@@ -0,0 +1,189 @@
+# Example of chart configuration
+
+## How to deploy the OpenTelemetry Operator and Java auto-instrumentation
+> :Notice: Operator related features should be considered to have an alpha
+> maturity level and be experimental. There may be breaking changes or
+> Operator features may be replaced entirely with a better alternative
+> solution in the future.
+
+In the following example we will show how to instrument
+[spring-petclinic-microservices](https://github.com/spring-petclinic/spring-petclinic-microservices).
```
rmfitzpatrick, line 10: Do you mind adding more detail about what this example will do and why?

examples/enable-operator-and-auto-instrumentation/README.md
```diff
@@ -0,0 +1,189 @@
...
+# cert-manager-6b466d8fb4-njt5n              1/1     Running   0          5h26m```
+```
+
+If it does need to be installed, run this.
+
+```
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.10.2/cert-manager.yaml
+```
+
+The cert-manager adds certificates and certificate issuers as resource types in Kubernetes clusters, and simplifies the process of obtaining, renewing and using those certificates.
```
rmfitzpatrick, line 30: I'd move this to be the first paragraph in this section and ideally explain why it's needed for the example.

examples/enable-operator-and-auto-instrumentation/README.md
```diff
@@ -0,0 +1,189 @@
...
+#       OTEL_RESOURCE_ATTRIBUTES_NODE_NAME:   (v1:spec.nodeName)
+#       OTEL_PROPAGATORS:                    tracecontext,baggage,b3
+#       OTEL_RESOURCE_ATTRIBUTES:            k8s.container.name=app,k8s.deployment.name=spring-petclinic,k8s.namespace.name=spring-petclinic,k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME),k8s.replicaset.name=spring-petclinic-9d5bc5fff
+#     Mounts:
+#       /otel-auto-instrumentation from opentelemetry-auto-instrumentation (rw)
+```
+
+</details>
+
+### 7. Check out the results at [Splunk Observability APM](tps://app.signalfx.com/#/apm)
```
rmfitzpatrick, line 176: The realm will be needed for this (and I think us1 is a safer bet if linking to one only).

```suggestion
### 7. Check out the results at [Splunk Observability APM](https://app.signalfx.com/#/apm)
```

jvoravong: Good to know, done

examples/enable-operator-and-auto-instrumentation/README.md
jvoravong: Done

jvoravong:
Thanks for the feedback @rmfitzpatrick. I have met with several others I work with and have received excellent feedback. More changes coming soon.

ci_scripts/setup_cert_manager.sh
dmitryax, line 1: Is this used in CI?

docs/auto-instrumentation-install.md
```diff
@@ -0,0 +1,278 @@
...
+## Getting started with auto-instrumentation
+
+### 1. If the cert-manager available in the cluster, deploy it
+
+The cert-manager adds certificates and certificate issuers as resource types in Kubernetes clusters, and simplifies the
+process of obtaining, renewing and using those certificates. You can use the following make commands to deploy the
+cert-manager.
+
+```
+make cert-manager
```
dmitryax, line 33: This implies that the user clones the repo first. Not sure this make target simplifies the user journey

docs/auto-instrumentation-install.md
```diff
@@ -0,0 +1,278 @@
...
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.10.0/cert-manager.yaml
+```
+
+### 2. Deploy the Helm Chart with the Operator enabled
+
+You can pass `--set operator.enabled=true` when deploying the chart to enable the operator. You can also use commands
+like in the following example.
+
+```
+VALUES_FILE=examples/enable-operator-and-auto-instrumentation/enable-operator-and-auto-instrumentation-values.yaml
```
dmitryax, line 46: Same concern here. We shouldn't expect users to clone the repo. They should not use `enable-operator-and-auto-instrumentation-values.yaml` as is, but use it as an example to create their own values.yaml

docs/auto-instrumentation-install.md
```diff
@@ -0,0 +1,278 @@
...
+    env:
+      - name: SPLUNK_OTEL_AGENT
+        valueFrom:
+          fieldRef:
+            apiVersion: v1
+            fieldPath: status.hostIP
+      - name: OTEL_EXPORTER_OTLP_ENDPOINT
+        value: http://$(SPLUNK_OTEL_AGENT):4317
+      - name: OTEL_TRACES_EXPORTER
+        value: otlp
```
dmitryax, line 85: Is there a way to install it automatically from our chart?

docs/auto-instrumentation-install.md
```diff
@@ -0,0 +1,278 @@
...
+metadata:
+  name: splunk-instrumentation
+spec:
+  exporter:
+    endpoint: http://$(SPLUNK_OTEL_AGENT):4317
+  propagators:
+    - tracecontext
+    - baggage
+    - b3
+  dotnet:
```
dmitryax, line 75: Shouldn't this be java?

docs/auto-instrumentation-install.md
```diff
@@ -0,0 +1,278 @@
...
+kubectl get otelinst -n spring-petclinic
+# NAME                          AGE   ENDPOINT
+# splunk-instrumentation        3m   http://$(SPLUNK_OTEL_AGENT):4317
+```
+
+</details>
+
+### 5. Instrument application by setting an annotation
+
+An _inject instrumentation_ annotation can be added to the following.
```
dmitryax, line 125: Is the annotation name `inject instrumentation`? Isn't it `instrumentation.opentelemetry.io/inject-java`?

docs/auto-instrumentation-install.md
```diff
@@ -0,0 +1,278 @@
...
+- Namespace: All pods within that namespace will be instrumented.
+- Pod Spec Objects: PodSpec objects that are available as part of Deployment,
+  Statefulset, or other resources can be annotated.
+  following values.
+- "true" - inject and Instrumentation resource from the namespace.
+- "my-instrumentation" - name of Instrumentation CR instance in the current
+  namespace.
+- "my-other-namespace/my-instrumentation" - name and namespace of.
+  Instrumentation CR instance in another namespace.
+- "false" - do not inject.
```
dmitryax, line 135: These parts are pretty unclear to me as a user, with the header saying: "An _inject instrumentation_ annotation can be added to the following."

examples/enable-operator-and-auto-instrumentation/README.md
```diff
@@ -0,0 +1,138 @@
+# Example of chart configuration
+
+## How to deploy the OpenTelemetry Operator and Java auto-instrumentation
+
+In the following example we will show how to instrument a project based on
+[spring-petclinic-microservices](https://github.com/spring-petclinic/spring-petclinic-microservices).
+
+## Getting started
```
dmitryax, line 8: Do we really need 3 "getting started" pages? :)

docs/auto-instrumentation-install.md
jvoravong: I've thought about this one, it is tricky. I included the necessary raw kubectl commands for this install as well to get around this for that reason.

We can't run a helm install with operator.enabmed=true unless the cert-manager is already up and running.

One solution I thought of but haven't had time to test is including the cert-manager as a subchart, I'm not sure what the order of operations would for all the subcharts installing thought. I'll try this solution today to see how plausible it is.

docs/auto-instrumentation-install.md
jvoravong: I meant for this step to be easily modifiable for the user to use their own values file. I can improve this with your input in mind, I think I should also update some related customer docs because they use similar setup logic.

docs/auto-instrumentation-install.md
jvoravong: I can be, but is does not need to be. In this case the instrumentation library used would be the native otel dotnet library. I'll add more details for clarification, we can add splunk and native otel instrumentation examples which should clarify things.

docs/auto-instrumentation-install.md
jvoravong: I spent a large chunk of time this week trying to solve this exact issue, I decided to put this code out this way at the start for now. Their are technical limitations we have to work around, I also think their may be a bug or undocumented functionality (still confirming, reported by other Splunkers as well) that relates to https://github.com/open-telemetry/opentelemetry-operator/issues/886 that has automating this more difficult.

I'm not sure if we should consider this a blocker for initial release. I think we should highly prioritize addressing this, but it is likely more appropriate to do in a follow up PR. Any thoughts?

docs/auto-instrumentation-install.md
jvoravong: This is a direct copy of the upstream documentation, this has been a pain point for new users. I do think these docs could be improved. I need to search the Operator code in depth to actually add any more details about this, I'd likely add these new docs to the upstream project as well.

I'll see what I can do here.

docs/auto-instrumentation-install.md
jvoravong: Other annotation names include:
instrumentation.opentelemetry.io/inject-python
instrumentation.opentelemetry.io/inject-nodejs
instrumentation.opentelemetry.io/inject-dotnet
more...

I'll add more specs and supported annotation names to this areas.

examples/enable-operator-and-auto-instrumentation/README.md
jvoravong: This me being an engineer who writes docs, willl update.

docs/auto-instrumentation-install.md
dmitryax: A follow up sounds good to me 👍

docs/auto-instrumentation-install.md
jvoravong: Done

docs/auto-instrumentation-install.md
jvoravong: Done

docs/auto-instrumentation-install.md
jvoravong: Done

docs/auto-instrumentation-install.md
jvoravong: Done

examples/enable-operator-and-auto-instrumentation/README.md
jvoravong: Done

jvoravong:
I decided to not use alias names for the subcharts because the cert-manager chart has naming requirements the alias name would have broken.

docs/auto-instrumentation-install.md
```diff
@@ -0,0 +1,282 @@
...
+        valueFrom:
+          fieldRef:
+            apiVersion: v1
+            fieldPath: status.hostIP
+      - name: OTEL_EXPORTER_OTLP_ENDPOINT
+        value: https://$(SPLUNK_OTEL_AGENT):4318
+      - name: OTEL_TRACES_EXPORTER
+        value: otlp
+  dotnet:
+    image: your-customized-auto-instrumentation-image:dotnet
```
dmitryax, line 78: It's confusing to me, as a user, what is `your-customized-auto-instrumentation-image:dotnet`. Why this option needs to be highlighted?

docs/auto-instrumentation-install.md
```diff
@@ -0,0 +1,282 @@
...
+    env:
+      - name: OTEL_EXPORTER_OTLP_ENDPOINT
+        value: http://{TARGET}:4317
+  # nodjs:
+  # python:
```
dmitryax, line 83: I'd remove it until we have proper support

docs/auto-instrumentation-install.md
```diff
@@ -0,0 +1,282 @@
...
+```
+
+</details>
+
+```
+# Install
+kubectl apply -f splunk-instrumentation.yaml
+# Check the current deployed values
+kubectl get otelint -o yaml
```
dmitryax, line 92:
```suggestion
kubectl get otelinst -o yaml
```

docs/auto-instrumentation-install.md
```diff
@@ -0,0 +1,282 @@
...
+# monitoring    splunk-otel-collector-opentelemetry-operator-56c4ddb4db-zcjgh   2/2     Running
+
+kubectl get mutatingwebhookconfiguration.admissionregistration.k8s.io -n monitoring
+# NAME                                      WEBHOOKS   AGE
+# splunk-otel-collector-cert-manager-webhook              1          14m
+# splunk-otel-collector-opentelemetry-operator-mutation   3          14m
+
+kubectl get otelinst -n spring-petclinic
+# NAME                          AGE   ENDPOINT
+# splunk-instrumentation        3m   http://$(SPLUNK_OTEL_AGENT):4317
```
dmitryax, line 118: This seems unexpected

helm-charts/splunk-otel-collector/crds/cert-manager.crds.yaml
dmitryax, line 1: How is this file being used?

docs/auto-instrumentation-install.md
jvoravong: I'll remove the other libraries that are not java for now. We just tested dotnet and I think notes like this will be good for developers but not for customers. I'll add some dev notes else where in a follow up PR.

docs/auto-instrumentation-install.md
jvoravong: - Instrumentation is a namespaced object we have to store some where. Ideally we would store one Instrumentation instance in the namespace the collector is installed in. Then auto-instrumentation happening in other namespaces should reference our one default Instrumentation instance .
- The `instrumentation.opentelemetry.io/inject-{library}: "namespace-name/instrumentation-name"` annotation  is supposed to allow us to use an Instrumentation instances from different namespaces. Last week I was digging into this and I did find a bug that relates to this. The cross namespace annotation is only respected in a subset of cases it should be respected. Until this bug is addressed, the work around is we will have to create an Instrumentation object in every namespace that will be using auto-instrumentation.

helm-charts/splunk-otel-collector/crds/cert-manager.crds.yaml
jvoravong: This is the simplest solution I could find to address errors from the cert-manager and operator being installed at the same time. The cert-manager chart will install the crds that the operator chart requires these crds for install, resulting in order of operations errors (helm install & upgrade). To fix the order of operations issues I tried using a wait initContainer and [chart hooks](https://helm.sh/docs/topics/charts_hooks/#the-available-hooks), these solutions did not work out. 

Helm will install any crds added to the crd directory to make available for Helm install and upgrade actions. By adding the cert-manager crds in this manner, we avoid install order of operations errors when installing the cert-manager and operator. I'll add a note to the code about this.

Any thoughts?

docs/auto-instrumentation-install.md
dmitryax: Should this be `spring-petclinic` namespace? This instruction doesn't mention spring-petclinic prior to this

helm-charts/splunk-otel-collector/charts/cert-manager-v1.11.0.tgz
dmitryax, line 1: Why are these binaries checked in? I don't think they need to be here

helm-charts/splunk-otel-collector/crds/cert-manager.crds.yaml
dmitryax: The problem is that now the crd will always be installed with the chart even if the operator is not enabled, right?

docs/auto-instrumentation-install.md
jvoravong: Yeah it should be spring-petclinic for now. We want the collector and operator to reside in a namespace our target applications are not located in. I added more notes to the install instructions. 

Hopefully in the future we should be able to support better cross namespace access of Instrumentation objects.

dmitryax: Not sure I understand why this namespace is used here. This document instructs users to install collector in `monitoring` and instrumentation in the default namespace (should they be the same?), but no mentions of `spring-petclinic` up to this point

helm-charts/splunk-otel-collector/crds/cert-manager.crds.yaml
jvoravong: Uploaded the hook solution.

docs/auto-instrumentation-install.md
jvoravong: Updated docs

helm-charts/splunk-otel-collector/charts/cert-manager-v1.11.0.tgz
jvoravong: I think you are right. Removed them. Looks like that broke some CI/CD steps. Fixing.

docs/auto-instrumentation-install.md
dmitryax: How did you update the docs?

We still ask user to deploy instrumentation in the [default namespace](https://github.com/signalfx/splunk-otel-collector-chart/pull/701/files/5dddb1579d4a929077a9ed241228cf9373071e42#diff-831285ac5240fe6b18737811f1bb2387e71261c895f92c090a8878e10d7e22d7R90):

```
kubectl apply -f splunk-instrumentation.yaml
```

but here look for it in `spring-petclinic`

```
kubectl get otelinst -n spring-petclinic
```

dmitryax (approved):
Couple nits, otherwise LGTM

docs/auto-instrumentation-install.md
jvoravong: I did miss adding the "-n spring-petclinic" to the apply command. Updated.

.gitignore
```diff
@@ -1,2 +1,5 @@
 .idea
 *.iml
+
+# Helm
+**/charts/*.tgz
```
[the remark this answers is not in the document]
jvoravong: The subcharts still get locally downloaded and used in local dev and CI/CD workflows (or else we run into "can't render subcharts" errors). I'm going to leave this here for this PR, will remove it in the future if possible.

docs/auto-instrumentation-install.md
dmitryax: Still unclear why do we install it in `spring-petclinic` while the collector is installed in `monitoring`?

jvoravong: - I want this example to demonstrate how to auto-instrument applications that do not reside within the same namespace as the operator.
- There is a bug or undocumented feature that prevents application pods from being auto-instrumented using an Instrumentation object that does not reside within the same namespace. I've been digging into.
  - Observed Behaviors
    - The application and operator pods exist in namespace X as well as the instrumentation object exist in namespace Y -> Auto-instrumentation succeeds
    - The application pods and instrumentation object exist in namespace X as well as the operator pods exist in namespace Y -> Auto-instrumentation succeeds
    - The application pods exist in namespace X as well as the operator pods and instrumentation object exist in namespace Y -> Auto-instrumentation fails (application cannot find Instrumentation)

dmitryax: Ok, thanks for clarifying

Issue #697, opened 2023-03-16
Errors in agent logs


KalebHawkins:
Hello, I am trying to use the collector on an Openshift 4.11 cluster to forward logs to Splunk.  I am using chart version 0.70. Using a pretty basic configuration file.

```yaml
splunk:
  clusterName: "<REDACTED>"
  splunkPlatform:
    endpoint: "<REDACTED>"
    token: "<REDACTED>"
    index: "<REDACTED>"
    metricsIndex: "<REDACTED>"
    tracesIndex: "<REDACTED>"
    logsEnabled: true
    metricsEnabled: true
    tracesEnabled: false
  logsEngine: otel
  distribution: "openshift"
  environment: <REDACTED>

  agent:
    enabled: true
    resources:
      limits:
        #cpu: 1000m
        memory: 4Gi
        
  clusterReceiver:
    enabled: true
    resources:
      limits:
        #cpu: 1000m
        memory: 8Gi

  logsCollection:
    containers:
      enabled: true
      containerRuntime: "cri-o"
```

The agents get the following error messages in their logging. I am wondering if this is an Openshift thing or if I have something misconfigured.

```
2023-03-16T15:53:22.447Z error prometheusexporter/prometheus.go:139 Could not get prometheus metrics {"kind": "receiver", "name": "receiver_creator", "pipeline": "metrics", "name": "smartagent/kubernetes-scheduler/receiver_creator{endpoint=\"xxx.xxx.xxx.xxx\"}/k8s_observer/ebe4c92e-8b5d-4be3-9945-cd439c5825c8", "monitorID": "smartagentkubernetesschedulerreceiver_creatorendpoint2071306682k8s_observerebe4c92e8b5d4be39945cd439c5825c8", "error": "Get \"http://xxx.xxx.xxx.xxx:10251/metrics\": dial tcp xxx.xxx.xxx.xxx:10251: connect: connection refused", "monitorType": "kubernetes-scheduler"}
```

```
[...]

PR #711, opened 2023-03-28
Update the port used to monitor the Kubernetes scheduler (10251->10259)

jvoravong:
Related Issue: https://github.com/signalfx/splunk-otel-collector-chart/issues/697

The previous port was deprecated in Kubernetes 1.19, it is time to update the out of date port. The new port was tested with Kubernetes 1.25 and Openshift 4.12.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Changed
+
+- Update the port used to monitor the Kubernetes scheduler (10251->10259) [#711](https://github.com/signalfx/splunk-otel-collector-chart/pull/711)
+
 ### Added
 
 - Add experimental support for deploying OpenTelemetry Operator as a subchart [#691](https://github.com/signalfx/splunk-otel-collector-chart/pull/691)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Update the port used to monitor the Kubernetes scheduler (10251->10259) [#711](https://github.com/signalfx/splunk-otel-collector-chart/pull/711)
+- Update the Kubernetes scheduler monitor to stop using insecure port 10251 and start using secure port 10259 with authentication [#711](https://github.com/signalfx/splunk-otel-collector-chart/pull/711)
 
 ### Added
 
--- a/examples/add-receiver-creator/rendered_manifests/configmap-agent.yaml
+++ b/examples/add-receiver-creator/rendered_manifests/configmap-agent.yaml
@@ -222,8 +222,11 @@ data:
             config:
               extraDimensions:
                 metric_source: kubernetes-scheduler
-              port: 10251
+              port: 10259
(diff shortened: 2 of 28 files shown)


examples/add-receiver-creator/rendered_manifests/configmap-agent.yaml
```diff
@@ -222,8 +222,11 @@ data:
             config:
               extraDimensions:
                 metric_source: kubernetes-scheduler
-              port: 10251
+              port: 10259
+              skipVerify: true
               type: kubernetes-scheduler
+              useHTTPS: true
+              useServiceAccount: true
```
crobert-1, line 229: Can you explain why we're also adding the `skipVerify`, `useHTTPS`, and `useServiceAccount` key-value pairs in this change?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -184,8 +184,11 @@ receivers:
         config:
           extraDimensions:
             metric_source: kubernetes-scheduler
-          port: 10251
+          skipVerify: true
+          port: 10259
           type: kubernetes-scheduler
+          useHTTPS: true
+          useServiceAccount: true
```
jinja2, line 191: Looks like [this change](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md#changes-by-kind-17) is made default in k8s version 1.23. We might still have customers on older k8s version who will need a workaround to use the old config. How about we add a check to add the new config iff k8s version is >= 1.23?

examples/add-receiver-creator/rendered_manifests/configmap-agent.yaml
jinja2: They have [secured the metrics endpoint of kube-scheduler](https://github.com/kubernetes/kubernetes/pull/96345), so we need to authenticate with k8s service account when scraping `/metrics` from scheduler.

jvoravong: If I recall properly, w need [skipVerify](https://github.com/signalfx/signalfx-agent/blob/a963c180611efabc6cb8f8e3abb3f3584b7d59a6/docs/monitors/kubernetes-scheduler.md?plain=1#L38)=true because the certificates for authentication are usually self hosted in a Kubernetes cluster.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
jvoravong: We don't lookup the Kubernetes version and use it in the Helm templates. I want to let this age out and delete this later to keep the scope small.

I want to refactor all of our control plane monitoring in the future and only maintenance it for now.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Changed
+
+- Update the port used to monitor the Kubernetes scheduler (10251->10259) [#711](https://github.com/signalfx/splunk-otel-collector-chart/pull/711)
```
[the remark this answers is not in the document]
jvoravong: done

jinja2 (approved):

PR #712, opened 2023-03-30
add max_batch_size to recombine instructions

atoulme:
Sets values for max_batch_size (default is 1000) for crio to 8, containerd and docker to 16 respectively.

### DIFF
--- a/examples/only-logs-otel/rendered_manifests/configmap-agent.yaml
+++ b/examples/only-logs-otel/rendered_manifests/configmap-agent.yaml
@@ -163,7 +163,7 @@ data:
           combine_with: ""
           id: crio-recombine
           is_last_entry: attributes.logtag == 'F'
-          max_batch_size: 16
+          max_batch_size: 8
           output: handle_empty_log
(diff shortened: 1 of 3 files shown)


matthewmodestino:
Suggest perhaps we go with `max_log_size` first to see if we can just set the size to 1MB vs the different settings for docker and cri-o/containerd. 


https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/recombine.md

If we do go with `max_batch_size` docker splits at 16384 and cri-o/containerd is 8192, so I would flip your numbers. 8 for docker, 16 for cri-o/containerd.

Lets be sure to capture the behaviour of 20MB+ logs, i assume we would flush then recombine the next batch from the big log together. Will help us identify them in Splunk.

atoulme:
OK, I'll rework this.

jvoravong (approved):

atoulme:
Let's go first with #713 as indicated by @matthewmodestino

atoulme:
Closing in favor of #713.

PR #728, opened 2023-04-12
Expose idle_conn_timeout setting

atoulme:
Expose idle_conn_timeout on the splunk HEC exporters created as part of the helm chart.

### DIFF
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -50,6 +50,8 @@ splunkPlatform:
   disableCompression: true
   # HTTP timeout when sending data. Defaults to 10s.
   timeout: 10s
+  # Idle connection timeout. defaults to 10s
+  idleConnTimeout: 10s
   # Whether to skip checking the certificate of the HEC endpoint when sending
   # data over HTTPS.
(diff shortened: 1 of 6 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -50,6 +50,8 @@ splunkPlatform:
   disableCompression: true
   # HTTP timeout when sending data. Defaults to 10s.
   timeout: 10s
+  # Idle connection timeout. defaults to 10s
+  idleConnTimeout: 10s
```
jinja2, line 54: Is the current default 10s? It seems a little aggressive especially if using connection pooling

atoulme: the default will be 10s going forward. 30s is the old default. The value has changed based on the defaults and best practices of Splunk Enterprise (set by `busyKeepAliveIdleTimeout` and defaults to 12 seconds).
PR on contrib: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/20653
Discussion: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/20543

jinja2 (approved):

PR #732, opened 2023-04-13
Example showing the collector installed as both an agent and separately a gateway

### LINKED ISSUE #522: Include an Agent+Gateway rendered manifest example
Currently, the example manifests are either "gateway-only" or "agent-only".

There doesn't appear (to me) to be an example which shows how you may typically deploy this in a k8s cluster with both the agent + gateway working together (in particular, the k8sattributes receiver config which does a `passthrough` on the agent).

This would probably be a good example to add.

atoulme:
Fixes #522

### DIFF
--- a/examples/collector-agent-and-gateway/README.md
+++ b/examples/collector-agent-and-gateway/README.md
@@ -1,5 +1,0 @@
-# Example of chart configuration
-
-## Deploy the OTel Collector in agent mode and separately as gateway mode
-
-This configuration will install the collector as an agent deployment, and separately as a gateway deployment.
--- a/examples/collector-agent-and-gateway/README.md
+++ b/examples/collector-agent-and-gateway/README.md
@@ -1,0 +1,5 @@
+# Example of chart configuration
+
+## Deploy the OTel Collector in agent mode and separately as gateway mode
+
+This configuration will install the collector as an agent deployment, and separately as a gateway deployment.
--- a/examples/collector-agent-and-gateway/collector-agent-and-gateway-values.yaml
+++ b/examples/collector-agent-and-gateway/collector-agent-and-gateway-values.yaml
@@ -1,0 +1,11 @@
+clusterName: CHANGEME
+splunkObservability:
+  realm: CHANGEME
+  accessToken: CHANGEME
+
+agent:
+  enabled: true
+gateway:
+  enabled: true
+clusterReceiver:
+  enabled: false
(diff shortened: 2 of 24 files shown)


examples/collector-agent-and-gateway/rendered_manifests/clusterRoleBinding.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: default-splunk-otel-collector
+  labels:
+    app.kubernetes.io/name: splunk-otel-collector
+    helm.sh/chart: splunk-otel-collector-0.72.0
+    app.kubernetes.io/managed-by: Helm
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/version: "0.72.0"
```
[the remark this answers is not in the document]
atoulme: this is generated from our values so while it shows a lot, we just set it in values.yaml and `make render` does the rest.

examples/collector-agent-and-gateway/collector-agent-and-gateway-values.yaml
```diff
@@ -0,0 +1,11 @@
...
+splunkObservability:
+  realm: CHANGEME
+  accessToken: CHANGEME
+
+agent:
+  enabled: true
+gateway:
+  enabled: true
+clusterReceiver:
+  enabled: false
```
dmitryax, line 11: Why `clusterReceiver` disabled? It's probably better to include it to get the all the o11y IM k8s experience

dmitryax:
I think we should remove one of the gateway-only examples now. They are very similar https://github.com/signalfx/splunk-otel-collector-chart/pull/735

examples/collector-agent-and-gateway/README.md
```diff
@@ -0,0 +1,5 @@
+# Example of chart configuration
+
+## Deploy the OTel Collector in agent mode and separately as gateway mode
+
+This configuration will install the collector as an agent deployment, and separately as a gateway deployment.
```
jvoravong, line 5:
```suggestion
This configuration will install the collector as an agent daemonset, and separately as a gateway deployment.
```

examples/collector-agent-and-gateway/collector-agent-and-gateway-values.yaml
atoulme: ok I'll do that instead.

PR #739, opened 2023-04-18
Otl 1409 Add java spring-petclinc demo to example

jvoravong:
Description:
- Adding a pure Java application demo to demonstrate how to setup auto-instrumentation. This demo is self contained, will be minimal maintenance, and is easy to spin up.
- This demo was created by covering the https://github.com/spring-petclinic/spring-petclinic-microservices/blob/master/docker-compose.yml file to Kubernetes manifests using [kompose.](https://kompose.io/).
- I had to update the port we send traces to from 4317 to 4318. 4318 is the correct port the Splunk Otel Collector receives trace telemetry data on. Upstream projects use ports 4317 and 4318 depending on the particular use case.
- These should be the last changes a user needs to setup java auto-instrumentation.

Testing:
Did end to end testing manually. We are adding automated test coverage very soon in a future ticket.


examples/enable-operator-and-auto-instrumentation/spring-petclinic/02_admin-server-deployment.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    kompose.cmd: kompose --file spring-petclinic/spring-petclinic-microservices/docker-compose.yml  convert
```
rmfitzpatrick, line 5: Is it reasonable to either include a makefile that goes through a (containerized) kompose process to generate these or to "de-kompose" them so they are standalone without a missing piece? This could even just include removing the annotations and merging the manifests into per-service files.

.gitignore
```diff
@@ -3,3 +3,4 @@
 
 # Helm
 **/charts/*.tgz
+helm-charts/splunk-otel-collector/Chart.lock
```
rmfitzpatrick, line 6: rebase?

examples/enable-operator-and-auto-instrumentation/spring-petclinic/02_admin-server-deployment.yaml
jvoravong: We could use a makefile action to generate resources like this. I declined doing so at this time for a couple of reason I could use some input on. We can easily move this demo code anywhere, I will meet with stake holder more about this later this week.

- Usually we just have to update image versions for demos like this. With the spring-clinic demo, we would likely just pull the latest image.
- We need to decide where demo projects live for all auto-instrumentations libraries, this is a great time to start to figure that out. Upstream projects like https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-demo create a whole chart for deploying demo environments. I would like us to create our own Helm chart for demo applications in the future, however, I would need to prioritize it and evaluate the acceptance criteria with the maintainer teams.

rmfitzpatrick: Longer term strategies sound good, but for these changes specifically introducing generated files without including the generator creates a missing step 0/1 for maintenance. This is especially a concern imo for customer-centric example content. If the aim is to avoid the effort to do so, merging/organizing the content in a more idiomatic way removes the requirement and can make the artifacts more intuitive.

jvoravong: Got it. Working on getting low maintenance solution for this now.

Makefile
```diff
@@ -1,13 +1,21 @@
...
 
 .PHONY: dep-build
 dep-build:
 	helm dependencies build ./helm-charts/splunk-otel-collector
+
+.PHONY: demo-update
+demo-update:
+	# Convert this into a method or script if other demos are added.
+	curl -L wget https://raw.githubusercontent.com/spring-petclinic/spring-petclinic-microservices/master/docker-compose.yml >> examples/enable-operator-and-auto-instrumentation/spring-petclinic/docker-compose.yaml
+	kompose convert --file examples/enable-operator-and-auto-instrumentation/spring-petclinic/docker-compose.yaml --out examples/enable-operator-and-auto-instrumentation/spring-petclinic/02_install_resources.yaml
```
rmfitzpatrick, line 20: This now requires kompose without providing an install option. Using a container would be my suggestion but overall I think it's better to move this to the example itself since it's not a project-wide requirement.

Makefile
```diff
@@ -1,13 +1,21 @@
...
+	helm repo update open-telemetry jetstack
 
 .PHONY: dep-build
 dep-build:
 	helm dependencies build ./helm-charts/splunk-otel-collector
+
+.PHONY: demo-update
+demo-update:
+	# Convert this into a method or script if other demos are added.
+	curl -L wget https://raw.githubusercontent.com/spring-petclinic/spring-petclinic-microservices/master/docker-compose.yml >> examples/enable-operator-and-auto-instrumentation/spring-petclinic/docker-compose.yaml
```
rmfitzpatrick, line 19:
```suggestion
	curl -L wget https://raw.githubusercontent.com/spring-petclinic/spring-petclinic-microservices/master/docker-compose.yml > examples/enable-operator-and-auto-instrumentation/spring-petclinic/docker-compose.yaml
```
if something fails in the kompose step would this duplicate the compose content on rerunning?

examples/enable-operator-and-auto-instrumentation/README.md
```diff
@@ -9,9 +9,7 @@ In the following example we will show how to instrument a project based on
 
 The Java spring-petclinic demo will create a spring-petclinic namespace and deploy the related Java applications to it.
 If you have your own Java application you want to instrument, you can still use the steps below as an example for how
-to intrument your application.
-
-TODO: Add the Kubernetes manifests for spring-petclinic to this example in a followup PR
+to instrument your application.
 
 ```
 kubectl apply -f examples/enable-operator-and-auto-instrumentation/spring-petclinic -R
```
rmfitzpatrick, line 15: I think this would install the services in the default namespace

Makefile
```diff
@@ -1,13 +1,21 @@
 .PHONY: render
 render:
+	make repo-update dep-build > /dev/null # Always make sure the required helm repos are added and up to date.
```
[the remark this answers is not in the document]
jvoravong: This was a quick poor person's solution to silencing stdout noise. Committed your change and added proper if statement checking to reduce log noise.

Makefile
jvoravong: Sgtm. Done.

examples/enable-operator-and-auto-instrumentation/README.md
jvoravong: You are right, updated.

Makefile
jvoravong: True, updated.

examples/enable-operator-and-auto-instrumentation/spring-petclinic/update-demo.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
rmfitzpatrick, line 1: I'd move this up a directory, partly so it's not attempted to be used as a manifest

jvoravong: Done

PR #740, opened 2023-04-18
Component label + local service definition

mr-miles:
Most importantly (and trivially), this PR adds a component label to the pods that are part of the daemonset.  Without this label it is impossible to target the agent pods with any selector (the other parts of the chart have component labels).

Secondly, it defines an (optional) service resource targetting the pods in the daemonset and using a "local" internal traffic policy (This tells kube-proxy to only use node local endpoints for cluster internal traffic).  This is useful because pods that have no way to be configured with the node ip can use the entry.

### DIFF
--- a/examples/add-receiver-creator/rendered_manifests/daemonset.yaml
+++ b/examples/add-receiver-creator/rendered_manifests/daemonset.yaml
@@ -31,7 +31,7 @@ spec:
         component: otel-collector-agent
         release: default
       annotations:
-        checksum/config: f216ea7912a66fe5b83ef263fa40a1c4e495cbdafcc94949770ebc5fe4e9d28d
+        checksum/config: a533fb1d9eeaa12f5cc09b2a8bee96a9faa7c0f43586cc980fcd7881e843f872
         kubectl.kubernetes.io/default-container: otel-collector
     spec:
       hostNetwork: true
--- a/examples/add-receiver-creator/rendered_manifests/daemonset.yaml
+++ b/examples/add-receiver-creator/rendered_manifests/daemonset.yaml
@@ -11,6 +11,7 @@ metadata:
     app.kubernetes.io/instance: default
     app.kubernetes.io/version: "0.83.0"
     app: splunk-otel-collector
+    component: otel-collector-agent
     chart: splunk-otel-collector-0.83.0
     release: default
     heritage: Helm
@@ -27,9 +28,10 @@ spec:
     metadata:
       labels:
         app: splunk-otel-collector
+        component: otel-collector-agent
         release: default
       annotations:
-        checksum/config: 51d5de68546dc64e3156bfb12b252f8d484339b0822eeb983d0bb7b9390c44b8
+        checksum/config: 63f04bb7bba061fa7b5e18bf6b6519b6803efe6559d30f918d9a209663eb313b
         kubectl.kubernetes.io/default-container: otel-collector
     spec:
       hostNetwork: true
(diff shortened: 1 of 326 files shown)


dmitryax:
Hi, @mr-miles. Thanks for your contribution. The change itself LGTM.

The problem is that it changes immutable fields (daemonset selector labels), which breaks the helm upgrade, and users will have to reinstall their daemonsets. 

We will need to migrate all the labels to [the recommended naming conventions](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) anyway, so I think if we can go through that only once and avoid doing it this time additionally.

What if we just add the `component: otel-agent` pod label only and keep the selector as is? The service selector should work well, right?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -378,6 +378,11 @@ agent:
   # Applicable only when updateStrategy is set to "RollingUpdate".
   # Can be an absolute number or a percentage. The default is 1.
   maxUnavailable: 1
+  
+  service:
```
jinja2, line 382: Please add new variable to the json schema here https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/values.schema.json

mr-miles:
Thanks @jinja2 @dmitryax - have done that and checks looks more healthy.  Didn't know about the daemonset selectors being immutable so learnt something too :-)

dmitryax:
Please run `make render` to generate examples

mr-miles:
Make render worked but the cri-o test is still complaining, not sure how to sort that though

dmitryax:
Looks like a flaky test. I restarted it

examples/add-receiver-creator/rendered_manifests/daemonset.yaml
```diff
@@ -22,11 +23,13 @@ spec:
   selector:
     matchLabels:
       app: splunk-otel-collector
+      # component: otel-collector-agent - needs to be added when label names are standardised
```
dmitryax, line 26: The last ask, can you please migrate it to the helm comment so it's not rendered

```suggestion
      {{/* component: otel-collector-agent - needs to be added when label names are standardised */}}
```

dmitryax (commented):
Thank you, @mr-miles. LGTM. just one nit

examples/add-receiver-creator/rendered_manifests/daemonset.yaml
dmitryax: I put it in a wrong place, sorry. Also better to use `{{/*-` to avoid empty lines

mr-miles:
Hi  @jinja2  @dmitryax - wondering if it is possible to revive this PR?  I think it just fell between the cracks.

I have just freshened it up by rebased it onto main and re-rendered the examples.

Alternatively please let me know if its no longer useful.

jinja2:
I think we can still go ahead with this PR. But I am wondering if we should make setting`hostNetwork` for the daemonset optionally `false` (default will still be true) possible with the PR. I can see users running non-hostmetrics receivers leveraging this service option and wanting to not run in the hostnetwork. @dmitryax wdyt?

jinja2:
@mr-miles can you run `make render` to regenerate examples?

mr-miles:
hmm not sure why it closed the PR ...

but I have rebased and "make render"-ed so should be good

jinja2 (approved):

PR #757, opened 2023-04-27
Add an option to use lightprometheus receiver through a feature gate

dmitryax:
Add an option to use experimental light prometheus receiver for metrics collection from discovered Prometheus endpoints.

Light Prometheus Receiver is optimized for performance and reduced memory footprint, but it does not support all Prometheus configuration options and not considered stable at this point.

### DIFF
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -1344,6 +1344,7 @@ networkExplorer:
       # seccomp.security.alpha.kubernetes.io/defaultProfileName: 'docker/default'
       # apparmor.security.beta.kubernetes.io/defaultProfileName: 'runtime/default'
 
+
 ################################################################################
 # Notice: Operator related features should be considered to have an alpha
 # maturity level and be experimental. There may be breaking changes or Operator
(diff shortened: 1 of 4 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1344,6 +1344,7 @@ networkExplorer:
       # seccomp.security.alpha.kubernetes.io/defaultProfileName: 'docker/default'
       # apparmor.security.beta.kubernetes.io/defaultProfileName: 'runtime/default'
 
+
```
jvoravong, line 1347: Nit: Extra line?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1383,3 +1384,18 @@ operator:
...
+# that are not yet ready for general availability.
+# Options in this section are not guaranteed to be stable and may change at any time.
+################################################################################
+
+featureGates:
+  # Use Light Prometheus Receiver for metrics collection from discovered Prometheus endpoints.
+  # https://github.com/signalfx/splunk-otel-collector/tree/main/internal/receiver/lightprometheusreceiver
+  # Light Prometheus Receiver is optimized for performance and reduced memory footprint.
+  # From the other hand, it does not support all Prometheus configuration options.
+  useLightPrometheusReceiver: false
```
jvoravong, line 1401: Note: I added support for using [feature gates per major component](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/advanced-configuration.md#using-feature-gates) a while ago (agent, clusterReceiver, gateway, etc). I'm fine with adding a root featureGates option, but it does seem like only the agent is using this at the moment.

jvoravong (approved):

helm-charts/splunk-otel-collector/values.yaml
dmitryax: Not sure how useful they are. I'd recommend replacing them with more generic `extraCmdAgrs`. WDYT?

dmitryax: > I'm fine with adding a root featureGates option, but it does seem like only the agent is using this at the moment.

To clarify, I'm adding _helm chart_ feature gates, not collector feature gates

jvoravong: > I'd recommend replacing them with more generic extraCmdAgrs. WDYT?

This sounds like a useful change to me, I've wanted functionality like this a couple times.

Issue #633, opened 2023-01-12
`multilineConfig` AND `extraOperators` cannot BOTH be set in values.yaml, `extraOperators` will be skipped


matthewmodestino:
Hi Team, 

Just a heads up. Currently if you try to use `multilineConfig` AND `extraOperators` your `extraOperators` will be skipped due to the nature of the rules we render from `multilineConfig`. 

As a workaround, I stole the rendered multiline rules from the configmap and commented out the `multilineConfig` in values.yaml, and put all the rules under `extraOperators` to control the flow of events.

VihasMakwana:
@matthewmodestino I'm trying to reproduce this, but I am getting both `multilineConfig` AND `extraOperators` in my configmap. Is this issue somehow fixed? If not, can you please attach the steps to reproduce?

matthewmodestino:
@VihasMakwana they will render in the configMap but the ordering and precedence can be an issue if the rules output to a subsequent operator and skips the rules.

PR #761, opened 2023-05-01
Change ordering in _otel_agent.tpl

VihasMakwana:
Fixes: #633 

Bug:
- When we use both `multilineConfig` AND `extraOperators` in values.yaml, `extraOperators` will be skipped because `multilineConfig` will output directly to the `clean-up-log-record` operator. 
- To fix this, move `extraOperators` above `multilineConfig` in _otel_agent.tpl


dmitryax:
@VihasMakwana, thanks for fixing it. Please add a changelog entry

VihasMakwana:
> @VihasMakwana, thanks for fixing it. Please add a changelog entry

@dmitryax done. please re-review

atoulme (approved):

dmitryax (approved):

PR #762, opened 2023-05-02
Add docker dependabot to java image locations

breedx-splk:
After https://github.com/signalfx/splunk-otel-java/pull/1241 gets merged and the next release publishes the updated docker image with the required dependabot tags, we can consider this.

According to the docs, dependabot [should be able](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#docker) to open PRs on docker changes. 🏆


.github/dependabot.yml
```diff
@@ -4,3 +4,11 @@ updates:
...
     schedule:
       interval: "daily"
+  - package-ecosystem: "docker"
+    directory: "helm-charts/splunk-otel-collector"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "docker"
+    directory: "helm-charts/splunk-otel-collector"
+    schedule:
+      interval: "daily"
```
atoulme, line 14: Did you add this twice? I don't see a difference.

breedx-splk: Whoops, it was supposed to be the examples dir. Fixing....

jvoravong (approved):
Awesome!

PR #765, opened 2023-05-02
Remove excessive istio attributes to avoid hitting the dimensions limit

dmitryax:
Add a new processor attributes/istio to drop excessive istio attributes and avoid running into the dimensions limit if scraping istio metrics is enabled.

(OTL-2117)

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -586,6 +586,30 @@ processors:
         new_name: container.memory.usage
   {{- end }}
 
+  {{- if or .Values.autodetect.prometheus .Values.autodetect.istio }}
+  # This processor is used to remove excessive istio attributes to avoid running into the dimensions limit.
+  # This configuration assumes single cluster istio deployment. If you run istio in multi-cluster scenarios,
+  # you may need to adjust this configuration.
+  attributes/istio:
+    include:
+      match_type: regexp
+      metric_names:
+        - istio_.*
+    actions:
+      - action: delete
+        key: source_cluster
+      - action: delete
+        key: destination_cluster
+      - action: delete
+        key: source_canonical_service
+      - action: delete
+        key: destination_canonical_service
+      - action: delete
+        key: source_canonical_revision
+      - action: delete
+        key: destination_canonical_revision
+  {{- end }}
+
 # By default only SAPM exporter enabled. It will be pointed to collector deployment if enabled,
 # Otherwise it's pointed directly to signalfx backend based on the values provided in signalfx setting.
 # These values should not be specified manually and will be set in the templates.
@@ -774,6 +798,9 @@ service:
       processors:
         - memory_limiter
         - batch
+        {{- if or .Values.autodetect.prometheus .Values.autodetect.istio }}
+        - attributes/istio
+        {{- end }}
         - resourcedetection
         - resource
         {{- if (and .Values.splunkPlatform.metricsEnabled .Values.environment) }}
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -588,7 +588,7 @@ processors:
 
   {{- if or .Values.autodetect.prometheus .Values.autodetect.istio }}
   # This processor is used to remove excessive istio attributes to avoid running into the dimensions limit.
-  # This configuration assumes single cluster istio deployment. If you run istio in multi-cluster scenarios,
+  # This configuration assumes single cluster istio deployment. If you run istio in multi-cluster scenarios or make use of the canonical service and revision labels,
   # you may need to adjust this configuration.
   attributes/istio:
     include:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   skipped when we use both `multilineConfig` AND `extraOperators` in values.yaml
 - Enable retry mechanism in filelog receiver to avoid dropping logs on backpressure from the downstream
   pipeline components [#764](https://github.com/signalfx/splunk-otel-collector-chart/pull/764)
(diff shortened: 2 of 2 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -586,6 +586,26 @@ processors:
...
+  # This configuration assumes single cluster istio deployment. If you run istio in multi-cluster scenarios,
+  # you may need to adjust this configuration.
+  attributes/istio:
+    include:
+      match_type: regexp
+      metric_names:
+        - istio_.*
+    actions:
+      - action: delete
+        key: source_cluster
```
rmfitzpatrick, line 600: would it make sense to add a new value and generate each action so it's more flexible as additional attributes are (un)wanted?

dmitryax: Do you mean having another option in values.yaml for that?

I wouldn't add more configuration options to the already bloated values.yaml for this niche use case.

Overriding the set of attributes is already possible, just a bit more verbose, like:
```yaml
agent:
  config:
    processors:
      attributes/istio:
        actions:
          - action: delete
            key: another_istio_attribute
          ...
```

dmitryax: Also, the current interface to enable istio metrics is pretty unflexible.

```yaml
autodetect:
  istio: true
```

We can probably rethink it to be able to add this list of attributes. Maybe as a follow-up effort. Let me know if you have any ideas

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -586,6 +586,26 @@ processors:
...
+      - action: delete
+        key: destination_cluster
```
owais, line 602: this might be useful for multi-cluster deployments.

dmitryax: I know. I added a comment above :) Please let me know if you think of other better options to exclude

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -586,6 +586,26 @@ processors:
...
+  {{- if or .Values.autodetect.prometheus .Values.autodetect.istio }}
+  # This processor is used to remove excessive istio attributes to avoid running into the dimensions limit.
+  # This configuration assumes single cluster istio deployment. If you run istio in multi-cluster scenarios,
+  # you may need to adjust this configuration.
+  attributes/istio:
+    include:
+      match_type: regexp
+      metric_names:
+        - istio_.*
+    actions:
```
jinja2, line 598: @dmitryax we hit a similar issue when infra upgraded istio in lab recently. This is what I noticed. There were 47 dims on the most common metric `istio_requests_total`. I found the easiest way to fix it was to disable the addition of host/env dimensions on metrics collected from istio-ingress (that freed up like 11 dimensions). In addition to this, infra team also excluded a few tags from being added in envoy itself when generating metrics.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -586,6 +586,26 @@ processors:
...
+      - action: delete
+        key: source_cluster
+      - action: delete
+        key: destination_cluster
+      - action: delete
+        key: source_canonical_service
```
jinja2, line 604: If we are removing the canonical_service then we might as well remove the canonical_revision labels

dmitryax: Makes sense. Removed `source_canonical_revision` and `destination_canonical_revision`

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: We already disabled all the attributes added by the prometheus receiver in https://github.com/signalfx/splunk-otel-collector/pull/3042 and https://github.com/signalfx/splunk-otel-collector-chart/pull/763. So at this point, we are at 36 (and 37 for buckets) dimensions for a typical deployment on GKE. Disabling these attributes gives us enough room I believe

jinja2: Ok, great. Do we want to consider removing the host and the cloud provider metadata that gets added with `sync_host_metadata` and resource detection enabled? Some users might want to keep the istio labels but be okay with these being removed esp, given this is info they can easily look up with the pod/node name.

dmitryax: Host and the cloud provider metadata is essential for the correlation with all other metrics and cloud integrations, I would keep them enabled by default across all the metrics that are being sent by the collector instead of removing them conditionally for istio metrics. Removing these 6 attribute should give us enough room

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: The problem is the transition from bool value to the map in a backward-compatible way. We can try something out but I would rather do it as a follow-up effort not blocking the release

dmitryax: Filed this https://github.com/signalfx/splunk-otel-collector-chart/issues/769

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -586,6 +586,30 @@ processors:
         new_name: container.memory.usage
   {{- end }}
 
+  {{- if or .Values.autodetect.prometheus .Values.autodetect.istio }}
+  # This processor is used to remove excessive istio attributes to avoid running into the dimensions limit.
+  # This configuration assumes single cluster istio deployment. If you run istio in multi-cluster scenarios,
```
jinja2, line 591:
```suggestion
  # This configuration assumes single cluster istio deployment. If you run istio in multi-cluster scenarios or make use the canonical service and revision labels,
```

PR #399, opened 2022-03-07
Add missing processors to metrics pipeline

tongpu:
We've seen that some labels that are present on logs are not present on metrics, add the `k8sattributes` and `resource/add_environment` processors to the metrics pipeline fixes this issue.

PR #791, opened 2023-05-22
Add deployment.environment to metrics always instead of just when splunkPlatform metrics are enabled

rjduffner:
When pulling in some prometheus generated metrics via annotations, I noticed the deployment.environment key was not getting set in SignalFx.

This PR changes it so that key always gets set when environment is set in the values file.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -805,7 +805,7 @@ service:
         {{- end }}
         - resourcedetection
         - resource
-        {{- if (and .Values.splunkPlatform.metricsEnabled .Values.environment) }}
+        {{- if .Values.environment }}
         - resource/add_environment
         {{- end }}
         {{- if .Values.isWindows }}


jvoravong:
Hey @rjduffner,
Thanks for contributing. This change looks valid to make our chart usage more uniform. Could you apply the same change to these other locations and add a CHANGELOG.md entry?
- https://github.com/signalfx/splunk-otel-collector-chart/blob/eb4ba48bad7bfa90a7442f43d276c7567f2f9072/helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl#L222
- https://github.com/signalfx/splunk-otel-collector-chart/blob/eb4ba48bad7bfa90a7442f43d276c7567f2f9072/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl#L150

jvoravong:
You will need to run [make render](https://github.com/signalfx/splunk-otel-collector-chart/blob/eb4ba48bad7bfa90a7442f43d276c7567f2f9072/Makefile#L1) on your local machine so the rendered examples are updated with your changes and for the pre-commit action to to pass.

dmitryax (changes requested):
This is done intentionally. `deployment.environment` is used for tracing and logs. Splunk Observability uses `sf_environment` for metrics which is set from the [trace correlation](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/signalfxexporter) which at the end must be the same value. So if you're using APM product, it should be already there. If you don't use the APM product, can you please share more details on how you use this attribute?

If we decide to make this change, it requires thorough testing before applying. It might introduce some issue for Splunk Observability.

dmitryax:
More context from the git blame https://github.com/signalfx/splunk-otel-collector-chart/pull/399

rjduffner:
Thanks for the tips there @jvoravong. Will update those.

@dmitryax, We need a way to correlate infra metrics coming off each host to a specific environment.

Yes APM metrics are working with sf_environment and traces are working with deployment.environment. What is missing is attributes from receivers like hostmetrics or k8s metrics. While we can filter based on cluster name, it would be nice to also have the environment since we do put it in that values file.

Maybe I am missing something though. Thanks!

dmitryax:
> Yes APM metrics are working with sf_environment and traces are working with deployment.environment. What is missing is attributes from receivers like hostmetrics or k8s metrics. While we can filter based on cluster name, it would be nice to also have the environment since we do put it in that values file.

All the metrics, including hostmetrics or k8s metrics, should have `sf_environment` as long as they are sent via `signalfx` exporter and the traces pipeline has `signalfx` exporter as well. See https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/signalfxexporter#traces-configuration-correlation-only for more details

rjduffner:
Hi @dmitryax, I have the signalfx exporter and checked the metric cpu.utilization.

I can see it contains the dimension, k8s.cluster.name (added here: https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl#L555) but we do not have sf_environment.

Also note, we are using the exporter as default in the chart and only passing in our token through the values file.

dmitryax:
@rjduffner `sf_environment` is sent as a property, not a dimension. You can use properties the same way as dimensions, but they are not visible in the data table. You should be able to see it in the metrics finder https://app.signalfx.com/#/metrics/?query=cpu.utilization

Screenshot 2023-05-24 at 3 42 42 PM

rjduffner:
Ok thanks for the information. Its a bit strange its not in the data table as it makes discoverability very hard as well as monitoring multiple envs at once... tricky to say the least.

But good to know its there. Thanks for the help.

PR #795, opened 2023-05-25
Update eks distro example with receiver to scrape k8s api-server metrics

jinja2:
This PR adds a new receiver to the eks distribution example to show users how they can make use of the prometheus receiver to scrape and report metrics from k8s api-server in a managed cluster like EKS.

### DIFF
--- a/examples/distribution-eks/distribution-eks-values.yaml
+++ b/examples/distribution-eks/distribution-eks-values.yaml
@@ -5,3 +5,34 @@ splunkObservability:
 
 distribution: eks
 cloudProvider: aws
+
+clusterReceiver:
+  config:
+    receivers:
+     # -- Example receiver to collect k8s api server metrics from EKS
+     # You can find the list of metrics available by running
+     # "kubectl get --raw /metrics" command against an EKS cluster
+      prometheus/k8s-api-server:
+        config:
+          scrape_configs:
+            - job_name: 'api-server-metrics'
+              tls_config:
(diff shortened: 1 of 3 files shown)


crobert-1 (approved):

dmitryax (approved):

dmitryax:
I believe we should set it up by default instead of https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl#L159

jinja2:
We can add it to be set as default for gke and eks distros. Haven't tested in AKS yet, but they probably expose these too. We'll have to decide on the default metrics to collect though cause there are a lot of metrics exposed.

jvoravong:
Yeah I think we need to refactor how we collect control plane metrics.
- Our default Splunk dashboard content requires a "metric_source" attribute like ["metric_source: kubernetes-apiserver"](https://github.com/signalfx/splunk-otel-collector-chart/blob/fdb574c7707d96b0ea9df9a6606fda7f0b254971/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl#L167) for control plane metrics, it may be time to change this internally.
- This last year, cloud hosted Kubernetes clusters added support for control plane metrics.
- I'm not sure how well multi-zone control planes metrics are collected when using a service to reach a metrics endpoint, I think this was a potential issue before that may have been resolved. If this isn't an issue I would prefer to use services instead of direct host ports (_otel-agent.tpl#L159).

PR #797, opened 2023-05-26
runAsUser SecurityContext not working for windows 

dloucasfx:
Based on https://github.com/kubernetes/kubernetes/issues/102849 runAsUser is not currently working for windows.
This prevents the agent from starting
````
Events:
  Type     Reason             Age   From                     Message
  ----     ------             ----  ----                     -------
  Normal   Scheduled          16s   default-scheduler        Successfully assigned default/splunk-otel-collector-1685115615-agent-gsb8n to ip-10-10-7-216.ec2.internal
  Normal   ResourceAllocated  16s   vpc-resource-controller  Allocated Resource vpc.amazonaws.com/PrivateIPv4Address: 10.10.11.36/19 to the pod
  Warning  FailedMount        16s   kubelet                  MountVolume.SetUp failed for volume "kube-api-access-26vnh" : chown c:\var\lib\kubelet\pods\3905601f-f6c7-401a-a328-e44bf75d80a6\volumes\kubernetes.io~projected\kube-api-access-26vnh\..2023_05_26_15_40_20.3621507769\token: not supported by windows
  Warning  FailedMount        16s   kubelet                  MountVolume.SetUp failed for volume "kube-api-access-26vnh" : chown c:\var\lib\kubelet\pods\3905601f-f6c7-401a-a328-e44bf75d80a6\volumes\kubernetes.io~projected\kube-api-access-26vnh\..2023_05_26_15_40_20.1631205512\token: not supported by windows
  Warning  FailedMount        14s   kubelet                  MountVolume.SetUp failed for volume "kube-api-access-26vnh" : chown c:\var\lib\kubelet\pods\3905601f-f6c7-401a-a328-e44bf75d80a6\volumes\kubernetes.io~projected\kube-api-access-26vnh\..2023_05_26_15_40_22.3704774755\token: not supported by windows
  Warning  FailedMount        12s   kubelet                  MountVolume.SetUp failed for volume "kube-api-access-26vnh" : chown c:\var\lib\kubelet\pods\3905601f-f6c7-401a-a328-e44bf75d80a6\volumes\kubernetes.io~projected\kube-api-access-26vnh\..2023_05_26_15_40_24.1619713144\token: not supported by windows
  ````
  
  This PR checks if `runAsUser` is defined in the securityContext for a windows agent, cluster-receiver and gateway and remove it if it exists.
  
  Finally, I have noticed that  `revert-patch-log-dirs-hook.yaml` is rendered for windows when `securityContext.runAsUser` and `securityContext.runAsGroup` are defined, although `runAsUser` does not work on windows, I aded an extra check to make sure not to bother when it's windows.

~~Note: I still need to do more validation and testing as we changed some requirements. Still need to recover Fluentd and use the helper in other places~~

### CODE REVIEW

File: helm-charts/splunk-otel-collector/templates/daemonset.yaml [hunk 1]
```diff
@@ -252,8 +252,14 @@ spec:
         {{- if or $agent.securityContext (and (eq (include "splunk-otel-collector.logsEnabled" $) "true") (eq .Values.logsEngine "otel")) }}
         securityContext:
           {{- if $agent.securityContext }}
-          {{- toYaml $agent.securityContext | nindent 10 }}
-          {{- else }}
+            {{- if .Values.isWindows }}
+              {{- with unset $agent.securityContext "runAsUser" }}
+              {{- toYaml $agent.securityContext | nindent 10 }}
```
crobert-1, line 257: Maybe I'm missing something, but if this line should be run whether or not it's a Windows system, we can just have it once outside of the `if..else..` instead of inside both `if` and `else`. Admittedly I'm not very experienced with YAML's formatting here, so I may be misunderstanding this.

Same goes for changes in the other files.

crobert-1 (changes requested):

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -252,8 +252,14 @@ spec:
...
+              {{- end}}
+            {{- else }}
+              {{- toYaml $agent.securityContext | nindent 10 }}
```
crobert-1, line 260:
```suggestion
              {{- toYaml $agent.securityContext | nindent 10 }}
```
The trailing whitespace here is the cause of the `pre-commit` [ci failure](https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/5092612755/jobs/9154186535?pr=797).

helm-charts/splunk-otel-collector/templates/daemonset.yaml
dloucasfx: @crobert-1 
For windows, we are running `with unset...` , if it's windows, then we just use the non-altered  `$agent.securityContext` 
The unset context will be lost if I run it outside the `with unset` block

crobert-1: Okay, thanks for clarifying!

crobert-1 (approved):

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
```diff
@@ -77,10 +77,8 @@ spec:
         ` }}
         {{- $clusterReceiver.affinity | mustMergeOverwrite (fromYaml $clusterReceiverPodAntiAffinity) | toYaml | nindent 8 }}
       {{- end }}
-      {{- if $clusterReceiver.securityContext }}
       securityContext:
-        {{ toYaml $clusterReceiver.securityContext | nindent 8 }}
-      {{- end }}
+        {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $clusterReceiver.securityContext) | nindent 8 }}
```
dmitryax, line 81: We don't need to set `runAsUser: 0` in the gateway and the cluster receiver. Only on the agent, both otel col and fluentd.

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -251,11 +251,7 @@ spec:
         imagePullPolicy: {{ .Values.image.otelcol.pullPolicy }}
         {{- if or $agent.securityContext (and (eq (include "splunk-otel-collector.logsEnabled" $) "true") (eq .Values.logsEngine "otel")) }}
         securityContext:
-          {{- if $agent.securityContext }}
-          {{- toYaml $agent.securityContext | nindent 10 }}
-          {{- else }}
-          runAsUser: 0
-          {{- end }}
+          {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $agent.securityContext) | nindent 10 }}
```
dmitryax, line 254: I don't think you have to pass `.Values.isWindows`. `$.Values.isWindows` should be available there

dloucasfx: @dmitryax 
If the caller is just passing the variable `.` then this would work, but I need to explicitly pass the securityContext of that particular block (ex: `$agent.securityContext`), so in helper func `.` will not have `.Values.isWindows` 

Is there a trick I am not aware of?

dmitryax: I mean `$.Values.isWindows` not `.Values.isWindows`

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
dloucasfx: I added another parameter to the helper func , whether to set runAsUser or not .
This was the helper fun can be called to clean up runAsUser from windows instances for the gw and cluster receiver.

helm-charts/splunk-otel-collector/templates/daemonset.yaml
dloucasfx: I can't access the `$.Values.isWindows` from the helper function, unless I pass `.` when I call the func
```<$.Values.isWindows>: nil pointer evaluating interface```

dmitryax (approved):

Issue #428, opened 2022-04-04
Issues deploying on EKS Windows Nodes


matthewmodestino:
Hi team, in working with a user deploying to EKS Windows Node Groups, we ran into a few items that caused the deployment to fail using default chart values. 

In order to deploy successfully we needed to make customizations to the chart, that like should be done by default, or included in documentation:

Update SecurityContext:

We had to update security context to use `ContainerAdministrator` user for windows. 

```
 # To collect container logs and journald logs, it will run the agent as a root user.
  # To run it as non root user, uncomment below `securityContext` options.
  # Setting runAsUser and runAsGroup to a non root user enables an init container that patches group
  # permissions of container logs directories on the host filesystem to make logs readable by this non root user.

  securityContext:
    windowsOptions:
      runAsUserName: ContainerAdministrator 
```

Then we had to create a new secret to not step on the old one. Although I assume in docs we could advise how to point to the existing one already created by first deploying the linux chart...

```
secret:
  create: true
  name: "splunk-otel-collector-windows"
  # Specifies whether secret provided by user should be validated.
  validateSecret: true
```

halovanic:
Thank you @matthewmodestino, I too had been struggling to make Windows on EKS work since 0.42 and this magic `agent.securityContext` value is indeed the fix.  It seems this should either be incorporated into the Windows documentation or keyed off the `isWindows` value since I don't think any other security options work under Windows.

I think the general direction for the secret problem is to fully deploy the collector chart twice, once for Linux and once for Windows (assuming you have both kinds of nodes).  That's most easily done by using a different namespace like "splunk-win", so this conflict won't happen and you can just manage them separately.

I also found that the default agent memory limit was just a bit too low under Windows to work reliably, although YMMV.

matthewmodestino:
That's another good point, I did find that both the liveness probe guidance in the chart was key, and did see instances where upping the resource limits was helpful. 

```
# This tells the kubelet that it should wait for x seconds before performing the first probe.
# This is required in case you are using windows worker nodes.
# It is recommended to keep it a 60-second window but it depends on cluster specification.
readinessProbe:
  initialDelaySeconds: 0
livenessProbe:
  initialDelaySeconds: 0
```

dmitryax:
@matthewmodestino thanks for reporting this. I've created another issue for secret name conflict https://github.com/signalfx/splunk-otel-collector-chart/issues/439

VihasMakwana:
[...]

PR #809, opened 2023-06-13
Use "ContainerAdministrator" for windows node

VihasMakwana:
For Linux nodes, the runAsUser is set to 0 by default if nothing is specified in values.yaml.
But for windows, it leaves this field empty. 
It makes sense to set runAsUserName inside windowsOptions to `ContainerAdministrator` which is the `root` equivalent for Windows node


rmfitzpatrick:
Are there specific issues this addresses and documented requirements for this user instead of `ContainerUser` or `LocalSystem`?

VihasMakwana:
> 

Ah, actually nope.
I came across https://github.com/signalfx/splunk-otel-collector-chart/issues/428.
It was mainly because of `runAsUser` being set even for Windows deployments. It was fixed AFAIK.
The reason why I raised this PR was to maintain consistency in Linux and Windows nodes.
In Linux, the agent's runAsUser is set to 0 i.e. root user if nothing is specified, but for Windows, it doesn't set anything so I made a few changes.

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -251,7 +251,11 @@ spec:
         imagePullPolicy: {{ .Values.image.otelcol.pullPolicy }}
         {{- if or $agent.securityContext (and (eq (include "splunk-otel-collector.logsEnabled" $) "true") (eq .Values.logsEngine "otel")) }}
         securityContext:
-          {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $agent.securityContext "setRunAsUser" true) | nindent 10 }}
+          {{- if .Values.isWindows}}
+          {{- include "splunk-otel-collector.windowsSecurityContext" (dict "securityContext" $agent.securityContext "setRunAsUser" true) | nindent 10 }}
+          {{- else }}
+          {{- include "splunk-otel-collector.securityContext" (dict "securityContext" $agent.securityContext "setRunAsUser" true) | nindent 10 }}
+          {{- end}}
```
dmitryax, line 258: Can we incorporate this login in the `splunk-otel-collector.securityContext` helper instead of having two of them?

VihasMakwana: Hmm, yes, I didn't go for it because I thought it would be quite messy, but let me give it a shot.

VihasMakwana: Done, please review

VihasMakwana:
@dmitryax can you please re-review this one?

jvoravong:
Should the proposed configuration be set when `isWindows: true`? 
I'm surprised we didn't see these changes affect https://github.com/signalfx/splunk-otel-collector-chart/tree/main/examples/kubernetes-windows-nodes.

jvoravong:
In that example [isWindows: true](https://github.com/signalfx/splunk-otel-collector-chart/blob/0312ea3abde3388647d075b79d31da1c5371515f/examples/kubernetes-windows-nodes/kubernetes-windows-nodes-values.yaml#L12). Your changes should have added content to the [rendered files in the example](https://github.com/signalfx/splunk-otel-collector-chart/tree/main/examples/kubernetes-windows-nodes/rendered_manifests) after running a `make render` command, please look into this.

VihasMakwana:
Okayy, will go through that

jvoravong (approved):

Issue #644, opened 2023-01-24
Logs pipeline operators logic assume UTC on kubernetes node, can cause incorrect timestamp in Splunk


matthewmodestino:
Hey Team!

Was working with a colleague (shout out to Mike Sakahara!) and discovered that our logs pipeline assumes kubernetes host will be in UTC. While this is definitely the most common scenario, and we do encourage users to always use UTC, if the host is not set to UTC, it breaks our hardcoded parsing logic:

First, if the container engine is not set in the helm chart (which i also always encourage) the `get-format` operator assumes Zulu timestamp:

https://github.com/signalfx/splunk-otel-collector-chart/blob/f9abebe4814f6fbc3988f009db0772e76ac1fe65/rendered/manifests/otel-logs/configmap-agent.yaml#L149-L157

Then when passed to the matched format parser (which actually may not even match anymore) the log format parser also fails to extract the timestamp due to hardcoded look for Zulu:

https://github.com/signalfx/splunk-otel-collector-chart/blob/f9abebe4814f6fbc3988f009db0772e76ac1fe65/rendered/manifests/otel-logs/configmap-agent.yaml#L172-L177

Ideally we should improve the regex to be able to handle the chance the host timestamp may not be UTC/Z and parse out the timestamp correctly. 

Here is an example of containerd logs on host set to GMT-6

```
2023-01-23T12:09:28.344779078-06:00 stdout F {"caller":"service_controller.go:103","controller":"ServiceReconciler","end reconcile":"kube-system/microk8s.io-hostpath","level":"info","ts":"2023-01-23T18:09:28Z"}
2023-01-23T12:09:28.344779078-06:00 stdout F {"caller":"service_controller.go:103","controller":"ServiceReconciler","end reconcile":"kube-system/microk8s.io-hostpath","level":"info","ts":"2023-01-23T18:09:28Z"}
root@zach:/var/log/pods/metallb-system_speaker-z78rf_57df697a-e63d-48d5-bccb-562074f1e6c9/speaker# date
Tue Jan 24 09:25:24 AM CST 2023
```
checked our regex parser in regex101 golang and it does not match. likely breaks our parsing completely, tho events still arrive in Splunk, but the timestamp is incorrectly showing 3 AM, when event is from 3 pm. 

<img width="1296" alt="image" src="https://user-images.githubusercontent.com/31549209/214347368-f11bdbd4-b480-4898-8c2c-f7c5cc3dd8c4.png">

[...]

PR #813, opened 2023-06-15
Make time format configruable.

VihasMakwana:
- Issue: https://github.com/signalfx/splunk-otel-collector-chart/issues/644
- If the host is not set to UTC, the current logic breaks. To resolve this issue, the user needs to override current operators in `agent.config` section, and it wouldn't look so good.
- Allow users to modify the timestamp string in the filelog receiver. It uses current values by default, but give the option to override them through `logsCollection.containers.timestampExtraction`


jvoravong:
Hey @VihasMakwana,
Can you describe in more detail what actually breaks or doesn't work as expected? Providing example logs and results would also be very helpful.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -575,6 +575,17 @@ logsCollection:
     # Set to 0 to remove any size limit.
     maxRecombineLogSize: 1048576
 
+    # Regex to extract logs, time for CRI-O or containerd logs
+    # More info: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/regex_parser.md
+    regexp: ""
```
jvoravong, line 580: A Kubernetes environment can use different container runtimes on different nodes at the same time. The suggested override configurations should be granular enough to configure each parser (crio, containerd, docker) individually.

VihasMakwana: Okayy, will take care of this one and update the PR

VihasMakwana:
> Hey @VihasMakwana, Can you describe in more detail what actually breaks or doesn't work as expected? Providing example logs and results would also be very helpful.


From the linked issue:

- **Our logs pipeline assumes kubernetes host will be in UTC. While this is definitely the most common scenario, and we do encourage users to always use UTC, if the host is not set to UTC, it breaks our hardcoded parsing logic.**
- **Ideally we should improve the regex to be able to handle the chance the host timestamp may not be UTC/Z and parse out the timestamp correctly.**

Example of contained runtime with a different timezone:
```
2023-01-23T12:09:28.344779078-06:00 stdout F {"caller":"service_controller.go:103","controller":"ServiceReconciler","end reconcile":"kube-system/microk8s.io-hostpath","level":"info","ts":"2023-01-23T18:09:28Z"}
2023-01-23T12:09:28.344779078-06:00 stdout F {"caller":"service_controller.go:103","controller":"ServiceReconciler","end reconcile":"kube-system/microk8s.io-hostpath","level":"info","ts":"2023-01-23T18:09:28Z"}
```
The above wouldn't match the hardcoded regex.

For more details about reproducing this, refer https://github.com/signalfx/splunk-otel-collector-chart/issues/644#issuecomment-1589344762.

dmitryax:
Is the issue fixed now by https://github.com/signalfx/splunk-otel-collector-chart/pull/817 ? I would like to avoid introducing new configuration options if they are not strictly required

jvoravong:
Left a similar [comment](https://github.com/signalfx/splunk-otel-collector-chart/issues/644#issuecomment-1601143341) as @dmitryax in the related issue. Was waiting for confirmation from involved reporters.

VihasMakwana:
Closing this one, will raise another PR for https://github.com/signalfx/splunk-otel-collector-chart/issues/829

PR #817, opened 2023-06-21
Fix cri-o log format time layout

jvoravong:
Related Issues (duplicates):
- https://github.com/signalfx/splunk-otel-collector-chart/issues/644
- https://github.com/open-telemetry/opentelemetry-helm-charts/issues/718

Description:

- Adopting a bug fix from the community. Basically the root cause of the bug being fixed is containerd log formats are not as finalized as CRI-O and Docker log formats. containerd log formats have recently changed in some K8s environments to look very much like CRI-O log formats. 
- The community has updated the CRI-O parser to accept containerd logs that are formatted very similarly to CRI-O logs, which fixes this bug. We may only need one parser for CRI-O and containerd logs in the future. 
- Keeping our current containerd parser would be a good idea for now for having legacy containerd format support.

### DIFF
--- a/examples/only-logs-otel/rendered_manifests/configmap-agent.yaml
+++ b/examples/only-logs-otel/rendered_manifests/configmap-agent.yaml
@@ -155,7 +155,7 @@ data:
         - id: parser-crio
           regex: ^(?P<time>[^ Z]+) (?P<stream>stdout|stderr) (?P<logtag>[^ ]*) ?(?P<log>.*)$
           timestamp:
-            layout: "2006-01-02T15:04:05.999999999-07:00"
+            layout: 2006-01-02T15:04:05.999999999Z07:00
             layout_type: gotime
             parse_from: attributes.time
           type: regex_parser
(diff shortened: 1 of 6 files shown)


examples/only-logs-otel/rendered_manifests/configmap-agent.yaml
```diff
@@ -155,7 +155,7 @@ data:
         - id: parser-crio
           regex: ^(?P<time>[^ Z]+) (?P<stream>stdout|stderr) (?P<logtag>[^ ]*) ?(?P<log>.*)$
           timestamp:
-            layout: "2006-01-02T15:04:05.999999999-07:00"
+            layout: 2006-01-02T15:04:05.999999999Z07:00
```
rmfitzpatrick, line 158: It's unfortunate format constants like `RFC3339Nano` can't be used.

rmfitzpatrick (approved):

examples/only-logs-otel/rendered_manifests/configmap-agent.yaml
jvoravong: +1

PR #819, opened 2023-06-21
Poll for new java agent every 4 hours

breedx-splk:
Instead of the `splunk-otel-java' release process needing to push to every repo that depends on it, we think it is more appropriate for the dependent repos to check for new versions themselves. This moves the dependency from an external repo back into the repo that needs it.

When a change in agent version is detected, a PR should get created automatically.

Once this is set up and shown to be working, we will remove the related functionality from the agent workflows.


jvoravong (approved):
It would be good to automate this dependency update task. Will keep an eye out for upcoming PRs to update the splunk-otel-java dependency.

PR #825, opened 2023-06-22
Add service.name attribute to OTel logs if istio autodetect is enabled

dmitryax:
Add `service.name` resource attribute to logs if `autodetect.istio` is enabled using transform processor instead of using fluentd jq plugin. This change removes the limitation of `service.name` attribute being available only with logsEngine=fluentd.

Once https://github.com/signalfx/splunk-otel-collector-chart/pull/826 merged, the rendered diff will show the istio related changes

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -156,16 +156,6 @@ resource/logs:
       action: delete
     - key: {{ include "splunk-otel-collector.filterAttr" . }}
       action: delete
-    {{- if .Values.autodetect.istio }}
-    - key: service.name
-      from_attribute: k8s.pod.labels.app
-      action: insert
-    - key: service.name
-      from_attribute: istio_service_name
-      action: insert
-    - key: istio_service_name
-      action: delete
-    {{- end }}
     {{- if .Values.splunkPlatform.fieldNameConvention.renameFieldsSck }}
     - key: container_name
       from_attribute: k8s.container.name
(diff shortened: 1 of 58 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -198,6 +198,10 @@ distribution: ""
 
 autodetect:
   prometheus: false
+  # This option is recommended for istio environments. It does the following things:
+  # - Enables scraping istio control plane metrics from Promethes endpoints.
+  # - Add a `service.name` resource attribute to logs with the same value as istio generates for
+  #   traces to enable correlation between logs and traces usign this attribute.
```
jvoravong, line 204:
```suggestion
  #   traces to enable correlation between logs and traces using this attribute.
```

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -216,6 +206,26 @@ resource/logs:
...
+  log_statements:
+    - context: resource
+      statements:
+        - set(cache["owner_name"], attributes["k8s.pod.name"]) where attributes["k8s.pod.name"] != nil
+        # Name of the object owning the pod is taken from "k8s.pod.name" attribute by striping the pod suffix according
+        # to the k8s name generation rules (we don't want to put pressure on the k8s API server to get the owner name):
+        # https://github.com/kubernetes/apimachinery/blob/ff522ab81c745a9ac5f7eeb7852fac134194a3b6/pkg/util/rand/rand.go#L92-L127
+        - replace_pattern(cache["owner_name"], "^(.+?)-(?:(?:[0-9bcdf]+-)?[bcdfghjklmnpqrstvwxz2456789]{5}|[0-9]+)$$", "$$1") where cache["owner_name"] != nil
+        - set(attributes["service.name"], Concat([cache["owner_name"], attributes["k8s.namespace.name"]], ".")) where cache["owner_name"] != nil and attributes["k8s.namespace.name"] != nil
+        - set(attributes["service.name"], Concat([attributes["k8s.pod.labels.app"], attributes["k8s.namespace.name"]], ".")) where attributes["k8s.pod.labels.app"] != nil and attributes["k8s.namespace.name"] != nil
```
rmfitzpatrick, line 226: `insert` for `service.name` was the existing process but not we're always overwriting?

Apart from that, since the update no longer relies on the dedicated `istio_service_name` attr and also restricts the app label setting to non-nil namespaces w/o a default I don't think these are equivalent, but it'd be hard to do that w/o a dedicated pipeline. I wonder if preserving `service.name` is enough though.

dmitryax: > insert for service.name was the existing process but now* we're always overwriting?

Good catch! I'll revert it.

> Apart from that, since the update no longer relies on the dedicated istio_service_name attr and also restricts the app label setting to non-nil namespaces w/o a default I don't think these are equivalent, but it'd be hard to do that w/o a dedicated pipeline.

Yeah, I changed the logic a bit. I think we can assume the namespace is always available unless users explicitly remove it, in that case, we don't set anything because, if we set a default value, it'll be different from what's set by istio on traces.

dmitryax: Updated to not override existing `service.name`

rmfitzpatrick (approved):

Issue #758, opened 2023-04-27
Error connecting to kubernetes-proxy


atoulme:
The change implemented here https://github.com/signalfx/splunk-otel-collector-chart/pull/711/files fixes the kubernetes-scheduler issue however, it does not fix the kubernetes-proxy issue.

From original post:
> 2023-03-16T15:53:22.550Z error prometheusexporter/prometheus.go:139 Could not get prometheus metrics {"kind": "receiver", "name": "receiver_creator", "pipeline": "metrics", "name": "smartagent/kubernetes-proxy/receiver_creator{endpoint=\"xxx.xxx.xxx.xxx\"}/k8s_observer/d742b91e-685e-4c35-8197-0b56ebc88e39", "monitorID": "smartagentkubernetesproxyreceiver_creatorendpoint2071306682k8s_observerd742b91e685e4c3581970b56ebc88e39", "monitorType": "kubernetes-proxy", "error": "Get \"http://xxx.xxx.xxx.xxx:29101/metrics\": dial tcp xxx.xxx.xxx.xxx:29101: connect: connection refused"}

_Originally posted by @kishah-lilly in https://github.com/signalfx/splunk-otel-collector-chart/issues/697#issuecomment-1526066251_

jvoravong:
I didn't notice this being an issue when testing and validating Openshift 4.12 with the Splunk Otel Collector Chart v0.72.0. 
If this is still an issue can any reporters, please post in depth details about your cluster setup.

kishah-lilly:
@jvoravong 
Helm chart version 0.75.0 on OpenShift 4.10 using **Splunk Observability configuration** only (**not** Splunk Cloud / Splunk Enterprise configuration):

```
agent:
  enabled: true

[...]
  
    proxy:
      # Specifies whether to collect proxy metrics.
      enabled: true
    scheduler:
      # Specifies whether to collect scheduler metrics.
      enabled: true
```

Setting proxy.enabled to false stops the errors.

Errors:
```
[...]

PR #832, opened 2023-06-27
Update the log level for metric scrape failures of the smartagent/kubernetes-proxy receiver from error to debug

jvoravong:
Description:
The default behavior of smart agent receivers based on Prometheus was to log an error when a metric scraping failure occurs. This can result in excessive log noise, especially when dealing with unstable metric source targets. Prometheus native metric collector mechanisms log these failures at the debug level. In the case of our control plane smart agent receivers in this chart, deploying them to a new, untested, or unsupported Kubernetes distribution can generate a significant amount of log error noise. These changes aim to reduce unexpected log noise when deploying to Kubernetes environments that are not directly supported by this chart.

Notes:
- The kube proxy control plane smart agent receiver is a specific case that often generates substantial log error noise in different Kubernetes environments. Multiple related issues have been filed across various repositories regarding kube proxy metric collection, including this repository.
  - [Error connecting to kubernetes-proxy](https://github.com/signalfx/splunk-otel-collector-chart/issues/758)
- An upstream related pull request has been merged, which enables this functionality.
  - [Add scrapeFailureLogLevel config to smartagent prometheus receivers](https://github.com/signalfx/splunk-otel-collector/pulls?q=is%3Apr+is%3Aclosed+scrape)

### DIFF
--- a/docs/advanced-configuration.md
+++ b/docs/advanced-configuration.md
@@ -379,7 +379,7 @@ agent:
 ### Known issues
 
 Kube Proxy
-* `10249: connect: connection refuse`
+* `10249: connect: connection refused`
   * Issue
     * When using a Kubernetes cluster with non-default configurations for kube proxy, there is a reported network connectivity issue that prevents the collection of proxy metrics.
   * Solution
--- a/docs/advanced-configuration.md
+++ b/docs/advanced-configuration.md
@@ -379,14 +379,19 @@ agent:
 ### Known issues
 
 Kube Proxy
-* https://github.com/kubernetes/kops/issues/6472
-  * Problem
-    * When using a kops created Kubernetes cluster, a network connectivity issue has been reported that prevents proxy
-      metrics from being collected.
+* `10249: connect: connection refuse`
+  * Issue
+    * When using a Kubernetes cluster with non-default configurations for kube proxy, there is a reported network connectivity issue that prevents the collection of proxy metrics.
   * Solution
-    * This issue can be addressed updating the kubeProxy metric bind address in the kops cluster spec:
-      * Set "kubeProxy.metricsBindAddress: 0.0.0.0" in the kops cluster spec.
-      * Deploy the change with "kops update cluster {cluster_name}" and "kops rolling-update cluster {cluster_name}".
+    * Update the kube proxy metric bind address (--metrics-bind-address) in the cluster spec.
+Set the kubeProxy metrics bind address to 0.0.0.0 or another value based on your Kubernetes cluster distribution.
+For this particular issue, the solution may vary depending on the Kubernetes cluster distribution. It is recommended to research what your Kubernetes distribution recommends for addressing this issue.
+  * Related Issue Links
+    * [kubernetes - Expose kube-proxy metrics on 0.0.0.0 by default ](https://github.com/kubernetes/kubernetes/pull/74300)
+    * [kubernetes - kube-proxy TLS support](https://github.com/kubernetes/kubernetes/issues/106870)
+    * [splunk-otel-collector-chart - Error connecting to kubernetes-proxy](https://github.com/signalfx/splunk-otel-collector-chart/issues/758)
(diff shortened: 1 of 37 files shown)


docs/advanced-configuration.md
```diff
@@ -379,14 +379,19 @@ agent:
 ### Known issues
 
 Kube Proxy
-* https://github.com/kubernetes/kops/issues/6472
-  * Problem
-    * When using a kops created Kubernetes cluster, a network connectivity issue has been reported that prevents proxy
-      metrics from being collected.
+* `10249: connect: connection refuse`
```
rmfitzpatrick, line 382:
```suggestion
* `10249: connect: connection refused`
```

docs/advanced-configuration.md
```diff
@@ -379,14 +379,19 @@ agent:
 ### Known issues
 
 Kube Proxy
-* https://github.com/kubernetes/kops/issues/6472
-  * Problem
-    * When using a kops created Kubernetes cluster, a network connectivity issue has been reported that prevents proxy
-      metrics from being collected.
+* `10249: connect: connection refused`
```
rmfitzpatrick, line 382: Do you know what logs/signals will persist to demonstrate configuration or connectivity issues with the change in log level? Does this error get reported on failed receiver startup?

jvoravong: After this change, users can view failure information and related stack traces in the collector agent logs. Prior to the change, these failures were logged at the error level, but they will now be logged at the debug level. It's important to note that these failures do not cause the collector to fail to start; the collector continues to operate normally. However, the failure is logged at regular intervals, which is determined by the configured interval. By default, the metric collection interval is set to 10 seconds, so this failure will be logged every 10 seconds.

jinja2 (approved):

docs/advanced-configuration.md
rmfitzpatrick: The agent log level isn't set though right (so it'd be info)? I mean without setting debug service telemetry logging.

jvoravong: That is correct, the agent will still use the default log level info. A user would have to update the [service telemetry logging level to debug](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/troubleshooting.md) in order to view these types of failures after this change.

Debugging this type of failure using debug logs shouldn't be unusual to users who are familiar with how to debug Prometheus issues.

rmfitzpatrick: Thanks, I'm referring to this scenario as being a potential issue. In order to reduce chattiness for some unsupported services on some deployments, we are now defaulting to silencing errors that could mask larger problems. My initial thought would be to leave the level at error for the api server or similar, or even add a new value to set the level for these components that is opt-in to debug.

jvoravong: We find ourselves in a situation where we need to balance the issue of "noisy logs in unsupported environments" with "inadequate log noise in supported environments."

To address this trade-off, I suggest taking a minimalist approach that can offer significant benefits. Based on the previous comments, I propose incorporating the scrapeFailureLogLevel: debug only for the kube-proxy receiver when the distribution is set to [""](https://github.com/signalfx/splunk-otel-collector-chart/blob/715d20e8f541187f0d8733dfabc935328a8c5c70/helm-charts/splunk-otel-collector/values.yaml#L181).

I plan to implement these changes soon, unless someone else presents an alternative approach they would like to suggest.

aunshc: Thanks Josh, focusing on unsupported environments makes sense. Let's start here and make sure the log level is a note in product docs too (customers should be aware of how to update the values.yaml file to override this). While we have this implemented for kube-proxy at this point, do we have similar pain points through issues for other control plane components as well?

jvoravong: @aunshc The other control plane components have barely had barely any of the same pain point issues (when it comes to custom k8s distribution implementation) compared to kube poxy . This could change in the future (seems unlikely at this time). We can focus only on addressing the kube proxy pain point for now.

PR #836, opened 2023-06-27
Add support to create default Instrumentation for operator based auto-instrumentation

jvoravong:
Description:
- This pull request introduces helm templates to facilitate the deployment of a default opentelemetry.io/v1alpha1 Instrumentation, eliminating the need for manual configuration by users.
- The default Instrumentation includes the following features:
  - It sets the endpoint value for exporting traces. If the user does not provide an endpoint value, the traces will be exported to the collector agent or gateway.
  - Users have the flexibility to override default values by utilizing the `operator.instrumentation.spec` field, similar to how the agent configuration can be overridden using `agent.config`.
- These changes bring us closer to the ability to create end-to-end tests for operator-based auto-instrumentation.

Testing:
- Manual testing was conducted by directing auto-instrumented Java applications to send traces to the collector agent and gateway.


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1360,12 +1361,38 @@ networkExplorer:
...
-# Related documentation: docs/auto-instrumentation-install.md
+# Related documentation: https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/auto-instrumentation-install.md
 # Full list of Helm value configurations: https://artifacthub.io/packages/helm/opentelemetry-helm/opentelemetry-operator?modal=values
 ################################################################################
 
 operator:
   enabled: false
+  # For more details, refer to: https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api.md#instrumentation
+  instrumentation:
+    create: true
```
dmitryax, line 1372: Why do we need to expose this option? Isn't the instrumentation the primary goal of introducing the operator?

jvoravong: This can be removed for now and added back later if we ever get a request for it. I could see a user using other methods for deploying and managing Instrumentation objects and not wanting to use our default. Not sure how common it would be though.

Removed.

docs/auto-instrumentation-install.md
```diff
@@ -30,10 +30,28 @@ If a cert-manager is not available in the cluster (or other TLS certificate sour
...
-the exported traces. There are two ways to set this attribute._
-- _Use the values.yaml optional `environment` configuration_
-- _Use the Instrumentation spec with the env var OTEL_RESOURCE_ATTRIBUTES_
+<details>
+<summary>Expand for extended Instrumentation configuration details</summary>
+
+- An [opentelemetry.io/v1alpha1 Instrumentation](https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api.md#instrumentation)
+object is used to configure the auto-instrumentation of your applications. To successfully enable instrumentation, the
+target pod must have an Instrumentation object available.
+- When `operator.enabled=true` and `operator.instrumentation.create=true`, the helm chart deploys a
```
dmitryax, line 39: `operator.instrumentation.create` is still here and a couple of other places

jvoravong: updated

examples/enable-operator-and-auto-instrumentation/enable-operator-and-auto-instrumentation-values.yaml
```diff
@@ -4,6 +4,8 @@ splunkObservability:
   logsEnabled: true
   accessToken: CHANGEME
   realm: us0
+  # Enables Always on Profiling
+  # profilingEnabled: true
```
dmitryax, line 8: Why is this commented out? Shouldn't it be applied?

helm-charts/splunk-otel-collector/templates/operator/_instrumentation-spec.tpl
```diff
@@ -0,0 +1,38 @@
...
+  - name: SPLUNK_PROFILER_MEMORY_ENABLED
+    value: "true"
+  {{- end }}
+python:
+  env:
+    # Required if endpoint is set to 4317.
+    # Python auto-instrumentation uses http/proto by default, so data must be sent to 4318 instead of 4317.
+    # See: https://github.com/open-telemetry/opentelemetry-operator#opentelemetry-auto-instrumentation-injection
+    - name: OTEL_EXPORTER_OTLP_ENDPOINT
+      value: {{- include "splunk-otel-collector.operator.instrumentation.exporter.endpoint" . | replace "4317" "4318" | nindent 6 }}
```
dmitryax, line 29: Can we maybe change `splunk-otel-collector.operator.instrumentation.exporter.endpoint` to return without the port instead of doing the replace?

helm-charts/splunk-otel-collector/templates/operator/_instrumentation-spec.tpl
```diff
@@ -0,0 +1,38 @@
+{{- define "splunk-otel-collector.operator.instrumentation.spec" -}}
```
dmitryax, line 1: I'd suggest we move this yaml to the `operator.instrumentation` section in values.yaml file. We already have pretty much the same snipped but commented out. If we move the actual default config there, it'll be easier for users to make change.

It won't work for the dynamic endpoints tho. We can keep them commented out in values.yaml and set them dynamically in the helper if users don't provide them explicitly.

jvoravong: Done. I moved as much as possible to the value.yaml file within the spec section, this likely improves readability. Need to fix the github workflow for updating the java image.

examples/enable-operator-and-auto-instrumentation/enable-operator-and-auto-instrumentation-values.yaml
jvoravong: From recent directions, profiling is not enabled by default when enabling auto-instrumentation. This is an opt-in after option, was trying to show this exist here with a comment. I'm trying to add "custom sub-case manifest renders" in the examples that will show off options like "profilingEnabled: true" well in a future PR. Let me know if you have input.

helm-charts/splunk-otel-collector/templates/operator/_instrumentation-spec.tpl
jvoravong: The suggested change would be harder to support over time because the upstream project uses host + port. I can refactor this code to be more well intentioned, that should help.
- Added a conditional statement to validate ":4317" is at the end of the endpoint value.

jvoravong: Done

.github/workflows/update_instrumentation_dependencies.yaml
```diff
@@ -1,39 +1,42 @@
...
-  LATEST_API: https://api.github.com/repos/signalfx/splunk-otel-java/releases/latest
+  LATEST_API: https://api.github.com/repos/signalfx/splunk-otel-{lang}/releases/latest
 
 jobs:
-  maybe_update_java:
+  maybe_update:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        language: ['java'] # Add other languages here
```
jvoravong, line 17: We might be able to have automation extract the list value for this line, will add in a future PR if possible.

examples/enable-operator-and-auto-instrumentation/enable-operator-and-auto-instrumentation-values.yaml
jvoravong: Removed these lines since we document them [instrumentation-enable-profiling.yaml](https://github.com/signalfx/splunk-otel-collector-chart/pull/836/files#diff-d0ee0bbabe39234f7ecfcbca2879598f013e0f3209463169c18984618bf5ada4)

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1360,23 +1361,60 @@ networkExplorer:
...
+        - b3
+      # Optional "sampler" parameter for enabling trace sampling, see: https://opentelemetry.io/docs/concepts/sdk-configuration/general-sdk-configuration/#otel_traces_sampler
+      # sampler:
+        # type: traceidratio
+        # argument: "0.95"
+      # Optional "environment variable" parameters that can configure all instrumentation libraries.
+      # env:
+        # Sets the deployment.environment tag on spans.
+        # - name: OTEL_RESOURCE_ATTRIBUTES
+        #   value: "deployment.environment=dev"
```
dmitryax, line 1394: This resource attribute is always set by the collector, I don't think we need to have any logic for `OTEL_RESOURCE_ATTRIBUTES` in the instrumentation. The only use case is deploying the instrumentation without an agent or gateway. I see you have some other logic to support that use case. Why do you think we need to support it? I'd say we can just fail if the agent and gateway are disabled with the instrumentation. It'll reduce a significant portion of code here. If users want instrumentation only, I think they should use the community operator instead. I'm just trying to see if we can avoid adding complexity to this chart for no apparent reason. Let me know if you disagree cc @aunshc

jvoravong: From my understanding I don't think we always set deployment.environment in the collector. Let me know if I am missing something. I would support converting deployment.environment configs to be non-optional.
- https://github.com/signalfx/splunk-otel-collector-chart/blob/ba1f0d348557a91656c68ebf393881ac5f918388/docs/advanced-configuration.md?plain=1#L67
- https://github.com/signalfx/splunk-otel-collector-chart/blob/ba1f0d348557a91656c68ebf393881ac5f918388/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl#L553C8-L553C8

jvoravong: Removed these comments

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
```diff
@@ -0,0 +1,64 @@
...
+    {{- $envSet := false -}}
+    {{- if and .Values.operator.instrumentation.spec .Values.operator.instrumentation.spec.env -}}
+      {{- range .Values.operator.instrumentation.spec.env -}}
+        {{- if and (eq .name "OTEL_RESOURCE_ATTRIBUTES") .value -}}
+          {{- $envSet = true -}}
+        {{- end -}}
+      {{- end -}}
+    {{- end -}}
+    {{- if not $envSet -}}
+      {{- fail "When operator.enabled=true, (splunkPlatform.tracesEnabled=true or splunkObservability.tracesEnabled=true), (agent.enabled=true or gateway.enabled=true), and .Values.operator.instrumentation.spec.exporter.endpoint is not set, either environment must be a non-empty string or operator.instrumentation.spec.env must contain an item with {name: OTEL_RESOURCE_ATTRIBUTES, value: non-empty string}" -}}
```
dmitryax, line 16: Why is this a requirement? Does APM require `development.environment`? if so, why can't we just require `environment` in values.yaml when instrumentation enabled?

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
```diff
@@ -0,0 +1,64 @@
...
+{{- $tracesEnabled := or (include "splunk-otel-collector.platformTracesEnabled" .) (include "splunk-otel-collector.o11yTracesEnabled" .) -}}
+{{- $endpointOverridden := and .Values.operator.instrumentation.spec .Values.operator.instrumentation.spec.exporter .Values.operator.instrumentation.spec.exporter.endpoint (ne .Values.operator.instrumentation.spec.exporter.endpoint "") -}}
+{{- if and .Values.operator.enabled $tracesEnabled (not $endpointOverridden) -}}
+  {{/*  If no endpoint override was provided, the environment variable/tag must be set by the agent, gateway, or instrumentation.*/}}
+  {{- if or (not .Values.environment) (eq .Values.environment "") -}}
+    {{- $envSet := false -}}
+    {{- if and .Values.operator.instrumentation.spec .Values.operator.instrumentation.spec.env -}}
+      {{- range .Values.operator.instrumentation.spec.env -}}
+        {{- if and (eq .name "OTEL_RESOURCE_ATTRIBUTES") .value -}}
+          {{- $envSet = true -}}
```
dmitryax, line 11: Having `OTEL_RESOURCE_ATTRIBUTES` doesn't mean `development.environment` is set, user can set any other resource attribute.

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
```diff
@@ -0,0 +1,64 @@
...
+
+{{- define "splunk-otel-collector.operator.instrumentation.exporter.endpoint" -}}
+{{- if and .Values.operator.instrumentation.spec .Values.operator.instrumentation.spec.exporter .Values.operator.instrumentation.spec.exporter.endpoint (ne .Values.operator.instrumentation.spec.exporter.endpoint "") }}
+  {{ .Values.operator.instrumentation.spec.exporter.endpoint }}
+{{- else if .Values.agent.enabled }}
+  http://$(SPLUNK_OTEL_AGENT):4317
+{{- else if .Values.gateway.enabled }}
+  http://{{ include "splunk-otel-collector.fullname" . }}:4317
+{{- else -}}
+  {{- fail "When operator.enabled=true, (splunkPlatform.tracesEnabled=true or splunkObservability.tracesEnabled=true), (agent.enabled=true or gateway.enabled=true), and .Values.operator.instrumentation.spec.exporter.endpoint is not set, either environment must be a non-empty string or operator.instrumentation.spec.env must contain an item with {name: OTEL_RESOURCE_ATTRIBUTES, value: non-empty string}" -}}
```
dmitryax, line 30: Wrong copy-paste

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
jvoravong: - APM telemetry data does require `development.environment` for default content content
- We can add that new requirement. Adding.

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
jvoravong: True, done.

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
jvoravong: Updated

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
jvoravong: Done

dmitryax (approved):

PR #837, opened 2023-06-28
Set cluster_name for host logs too if renameFieldsSck is enabled

### LINKED ISSUE #682: splunkPlatform.fieldNameConvention.renameFieldsSck doesn't cover journald units or extraFileLogs
According to the SCK migration guide, it's possible to preserve the Splunk field names as used with SCK. Without this, we would need to rework all our custom Splunk Dashboards. 

https://github.com/signalfx/splunk-otel-collector-chart/blob/ff1e1ff2bf415ab8c7f37e03d85c056debf688b8/docs/migration-from-sck.md?plain=1#L405-L414

Now I discovered that the setting only affects the `logs` pipeline but not the `logs/host`. The `resource/logs` is not included in the `logs/host` pipeline. https://github.com/signalfx/splunk-otel-collector-chart/blob/ff1e1ff2bf415ab8c7f37e03d85c056debf688b8/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl#L678

Probably for journald and extraFileLogs only the `cluster_name` field is relevant.

https://github.com/signalfx/splunk-otel-collector-chart/blob/ff1e1ff2bf415ab8c7f37e03d85c056debf688b8/helm-charts/splunk-otel-collector/templates/config/_common.tpl#L121-L123

vinzent:
Up to now the cluster_name field was only set in `resource/logs`. This snippet was only included for the container logs, but not for host logs (files, journald).

This change moves the cluster_name field to the `resource` snippet, which is according inline comment used for all things that pass through otel.

Fixes #682

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Fixed
+
+- Set cluster_name for host logs too if renameFieldsSck is enabled [#837](https://github.com/signalfx/splunk-otel-collector-chart/pull/837)
+
 ### Added
 
 - Update PodDisruptionBudgets API version to allow both `policy/v1beta1` and `policy/v1` [#835](https://github.com/signalfx/splunk-otel-collector-chart/pull/835)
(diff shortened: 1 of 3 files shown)


dmitryax (commented):
LGTM. @vinzent can you please add a changelog entry?

vinzent:
@dmitryax  added changelog entry.

dmitryax (approved):

PR #839, opened 2023-06-28
add emptyDir daemonset volume for SA receiver collectd config

rmfitzpatrick:
In cases like openshift's default SecurityContextConstraints when the agent is set to have a read-only root filesystem, collectd monitor use isn't possible without having a declared volume mount for its run directory:

```
error	collectd/monitorcore.go:126	Could not render collectd plugin config	{"kind": "receiver", "name": "smartagent/collectd/mysql", "data_type": "metrics", "error": "failed to create collectd config dir at /usr/lib/splunk-otel-collector/agent-bundle/run/collectd/monitor-smartagentcollectdmysql/managed_config: mkdir /usr/lib/splunk-otel-collector/agent-bundle/run/collectd/monitor-smartagentcollectdmysql: read-only file system", "monitorID": "smartagentcollectdmysql", "monitorType": "collectd/mysql", "path": "/usr/lib/splunk-otel-collector/agent-bundle/run/collectd/monitor-smartagentcollectdmysql/managed_config/20-mysql.smartagentcollectdmysql.conf"}
```


rmfitzpatrick:
It'd probably make sense to add these to the CR deployment as well.

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -404,11 +404,16 @@ spec:
...
+        - mountPath: /usr/lib/splunk-otel-collector/agent-bundle/run/collectd
+          name: run-collectd
+          readOnly: false
         {{- if $agent.extraVolumeMounts }}
         {{- toYaml $agent.extraVolumeMounts | nindent 8 }}
         {{- end }}
       terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
       volumes:
+      - name: run-collectd
+        emptyDir: {}
```
jinja2, line 416: How about adding a limit? I assume we are only writing collectd config files in this directory.
```suggestion
        emptyDir:
          sizeLimit: 500Mi
```

rmfitzpatrick: good call, though that would be a lot of config files and they should be cleanly removed on receiver shutdown.

jinja2 (approved):

jvoravong (approved):

PR #845, opened 2023-07-11
Adding agent correctness tests

pszkamruk-splunk:
Adding agent pod tests:

test which checks exclude agent logs flag set to true
test for checking metadata of agent log entry (source, sourcetype, index)
test for checking if all agent pod logs are ingested correctly into Splunk


test/config_yaml_files/agent_tests_values.yaml
```diff
@@ -0,0 +1,31 @@
+clusterName: sck-otel
+splunkPlatform:
+  endpoint:
+  insecureSkipVerify: true
+#  token: ${CI_SPLUNK_HEC_TOKEN}
```
jvoravong, line 5: Could you leave a comment for this?

jvoravong: Never mind, it is [documented](https://github.com/signalfx/splunk-otel-collector-chart/blob/6d3ca53d1e70f60a9c166217dd509a8a376cc058/test/README.md?plain=1#L39)

jvoravong (approved):

PR #859, opened 2023-07-20
Prepare release 0.81.0

jvoravong:


CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
+## [0.81.0] - 2023-07-20
+
+This Splunk OpenTelemetry Collector Chart for Kubernetes uses the following components
+
+- Splunk OpenTelemetry Collector - [0.81.0](https://github.com/signalfx/splunk-otel-collector/releases/tag/v0.81.0)
+- Splunk Network Explorer - [latest-v0.9](https://quay.io/repository/signalfx/splunk-network-explorer-kernel-collector?tab=tags)
+- cert-manager chart - [1.12.2](https://charts.jetstack.io)
+- opentelemetry-operator chart - [0.34.0](https://github.com/open-telemetry/opentelemetry-helm-charts/releases/tag/opentelemetry-operator-0.34.0)
+- Java auto-instrumentation - [1.26.0](https://github.com/signalfx/splunk-otel-java/releases/tag/v1.26.0)
```
dmitryax, line 15: It sounds like they are all used by default. And why fluentd isn't mentioned then?

I think we should keep the existing wording meaning that we track the release of the collector image in the chart versioning. Other image updates should be mentioned in the `Changed` section only when they are updated.

jvoravong: Left fluentd out because no one has updated the version in years. Reverted to old style for now.

dmitryax (approved):

atoulme (approved):

Issue #800, opened 2023-05-29
Cluster receiver crashes after helm update when file_storage extension is enabled


wojtekzyla:
## Description

While working on these PRs #675 and #753 I discovered that when `file_storage` was enabled for the cluster receiver, then making any update in the values.yaml file under `clusterReceiver` section and running `helm update` caused crash of the cluster receiver pod. Using `file_storage` in agent works fine.

## Steps to Reproduce

1. Set up cluster receiver config as described in the configuration and start the otel collector.
2. Make any change in the values.yaml under `clusterReceiver` section and run `helm update`.

## Expected Result

Cluster Receiver pod working correctly.

## Actual Result

Cluster Receiver pod crashes.

### Chart version

0.71.0

### Chart configuration

```yaml
Name:         sck-otel-splunk-otel-collector-otel-k8s-cluster-receiver
Namespace:    default
Labels:       app=splunk-otel-collector
              app.kubernetes.io/instance=sck-otel
              app.kubernetes.io/managed-by=Helm
              app.kubernetes.io/name=splunk-otel-collector
              app.kubernetes.io/version=0.71.0
              chart=splunk-otel-collector-0.71.0
              helm.sh/chart=splunk-otel-collector-0.71.0
              heritage=Helm
              release=sck-otel
Annotations:  meta.helm.sh/release-name: sck-otel
              meta.helm.sh/release-namespace: default

Data
====
relay:
----
exporters:
  splunk_hec/platform_metrics:
    disable_compression: true
    endpoint: <MY_ENDPOINT>
    index: metric
    max_connections: 200
    retry_on_failure:
      enabled: true
      initial_interval: 5s
      max_elapsed_time: 300s
      max_interval: 30s
    sending_queue:
      enabled: true
      num_consumers: 10
      queue_size: 5000
      storage: file_storage/persistent_queue
    source: kubernetes
    splunk_app_name: splunk-otel-collector
    splunk_app_version: 0.71.0
    timeout: 10s
    tls:
      insecure_skip_verify: true
    token: ${SPLUNK_PLATFORM_HEC_TOKEN}
extensions:
  file_storage/persistent_queue:
    directory: /var/addon/splunk/persist/clusterReceiver
  health_check: null
  memory_ballast:
    size_mib: ${SPLUNK_BALLAST_SIZE_MIB}
processors:
  batch: null
  memory_limiter:
    check_interval: 2s
[...]

PR #861, opened 2023-07-27
Persistence queue

VihasMakwana:
Things included in this PR:

- enable persistent queues configuration in the helm chart for an agent, 
- Don't enable the persistent queues for following scenarios:

    - If the gateway mode is enabled. Gateway uses multiple replicas by default, and our filestorage extension will try to acquire an exclusive lock for the directory and it will cause a deadlock. Only one pod will be in a healthy state if we enable a persistent queue for gateway.
    - If the gke/autopilot or eks/fargate are used. Both of these distributions don't allow hostPath mounts.

- Will add the same thing in the docs


atoulme:
You did the work for logs and metrics. What about traces? Do they also need that support?

Would you please add an example with persistence enabled so we can see what that looks like?

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -198,11 +203,16 @@ service:
...
       address: 0.0.0.0:8889
-  {{- if eq (include "splunk-otel-collector.distribution" .) "eks/fargate" }}
-  extensions: [health_check, memory_ballast, k8s_observer]
-  {{- else }}
-  extensions: [health_check, memory_ballast]
-  {{- end }}
+  extensions:
+    - health_check
+    - memory_ballast
+    {{- if eq (include "splunk-otel-collector.distribution" .) "eks/fargate" }}
```
atoulme, line 209: This sounds unrelated to this PR.

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -441,6 +441,26 @@ Whether clusterReceiver should be enabled
...
+{{- end -}}
+
+
+{{- define "splunk-otel-collector.persistentQueueEnabledMetrics" -}}
+{{ $gatewayDisabled := ne (include "splunk-otel-collector.gatewayEnabled" .) "true" }}
+{{- and $gatewayDisabled (.Values.splunkPlatform.sendingQueue.persistentQueueEnabled.metrics) (and (ne (include "splunk-otel-collector.distribution" .) "eks/fargate") (ne (include "splunk-otel-collector.distribution" .) "gke/autopilot")) -}}
+{{- end -}}
+
+
+{{- define "splunk-otel-collector.persistentQueueEnabled" -}}
```
atoulme, line 459: I'm a bit fuzzy on why you have this boolean and how you use it. I'll comment below on it.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -11,6 +11,11 @@ extensions:
     directory: {{ .Values.logsCollection.checkpointPath }}
   {{- end }}
 
+  {{- if (eq (include "splunk-otel-collector.persistentQueueEnabled" .) "true") }}
```
atoulme, line 14: so we define both logs and metrics extensions when persistent queue is enabled, regardless if only logs or metrics are enabled. Not sure why?

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -198,11 +203,16 @@ service:
...
+    - k8s_observer
+    {{- end }}
+    {{- if (eq (include "splunk-otel-collector.persistentQueueEnabled" .) "true") }}
```
atoulme, line 212: Same here, we define both

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -106,6 +106,18 @@ splunkPlatform:
     #   requests_per_second is the average number of requests per seconds.
     queueSize: 5000
 
+    # When enabled, it uses file_storage extension to persist the queue data.
+    # This can be used as an alternative redundancy mechanism for data being exported.
+    # NOTE: The File Storage extension will persist state to the node's local file system.
+    # While using the persistent queue it is advised to increase memory limit for agent (agent.resources.limits.memory)
+    # to 1Gi. Persistent queue is currently supported only for logs and metrics of the agent.
```
atoulme, line 113: Where is that memory recommendation coming from?

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -441,6 +441,26 @@ Whether clusterReceiver should be enabled
 {{- end -}}
 
 
+{{/*
+Whether persistentQueue should be enabled
+*/}}
+{{- define "splunk-otel-collector.persistentQueueEnabledLogs" -}}
+{{ $gatewayDisabled := ne (include "splunk-otel-collector.gatewayEnabled" .) "true" }}
+{{- and $gatewayDisabled (.Values.splunkPlatform.sendingQueue.persistentQueueEnabled.logs) (and (ne (include "splunk-otel-collector.distribution" .) "eks/fargate") (ne (include "splunk-otel-collector.distribution" .) "gke/autopilot")) -}}
+{{- end -}}
```
atoulme, line 450: It looks like we only enable log persistence queue if gateway is disabled and we don't run in eks/fargate or autopilot, on top of our boolean field.
Can we document that limitation? Is there a reason we cannot do that on those platforms? Can we instead abort if someone tries to set that boolean to true?

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -5,6 +5,11 @@ The values can be overridden in .Values.clusterReceiver.config
 {{- define "splunk-otel-collector.clusterReceiverConfig" -}}
 {{ $clusterReceiver := fromYaml (include "splunk-otel-collector.clusterReceiver" .) -}}
 extensions:
+  {{- if (eq (include "splunk-otel-collector.persistentQueueEnabled" .) "true") }}
+  {{- include "splunk-otel-collector.persistentQueueLogs" (dict "Values" .Values "forAgent" false) | nindent 2 }}
+  {{- include "splunk-otel-collector.persistentQueueMetrics" (dict "Values" .Values "forAgent" false) | nindent 2 }}
```
atoulme, line 10: you are defining both file storage extensions regardless of whether only one is set, and therefore might get errors because the folder is not created.

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
VihasMakwana: yeah missed this one. will fix it

helm-charts/splunk-otel-collector/templates/_helpers.tpl
VihasMakwana: I mentioned the reasons in PR description.

VihasMakwana: Pasting the same here:
```
Don't enable the persistent queues for following scenarios:

If the gateway mode is enabled. Gateway uses multiple replicas by default, and our filestorage extension will try to acquire an exclusive lock for the directory and it will cause a deadlock. Only one pod will be in a healthy state if we enable a persistent queue for gateway.
If the gke/autopilot or eks/fargate are used. Both of these distributions don't allow hostPath mounts.

```

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
VihasMakwana: yeah, will resolve this.

helm-charts/splunk-otel-collector/values.yaml
VihasMakwana: From sizing guide: https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/troubleshooting.md#sizing-splunk-otel-collector-containers

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
VihasMakwana: Kind of. 
But if I had kept it at it is, it would've been a bit complete, as multiple blocks should be added I guess.
If we just split out that `extensions` array, it provides greater flexibility to add/remove more extensions in the future.

helm-charts/splunk-otel-collector/templates/_helpers.tpl
VihasMakwana: yeah, got rid of it as it served little purpose. It was used only once, in clusterReciever's deployment.

VihasMakwana:
@atoulme it completely makes sense to enable persistent queuing for traces as well. As a result, I have updated this PR.

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
```diff
@@ -227,6 +271,18 @@ spec:
       - name: messages
         emptyDir: {}
       {{- end }}
+      {{- if (eq (include "splunk-otel-collector.persistentQueueEnabledMetrics" .) "true") }}
+      - name: persistent-queue-metrics
+        hostPath:
```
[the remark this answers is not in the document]
VihasMakwana: I think statefulset would be better, right?
I'm not sure if we **strictly** need PVC.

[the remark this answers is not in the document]
jinja2: > I'm not sure if we strictly need PVC.
How would you get a persistent volume for the pod to persist its queue if not with a pvc? Switching to statefulset but still using hostPath will still mean the user needs to pin pod to a specific node. A persistent volume on the other hand means the queue moves with the pod when rescheduled.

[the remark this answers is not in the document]
VihasMakwana: I see.
If we go ahead with PCV, we would need to use either `local` or `hostPath` configs. `hostPath` is out of the picture, as it will only work in single node env. 
Now, for `local` as well, we need to specify `nodeSelector`.
From https://kubernetes.io/docs/concepts/storage/volumes/#local,

> You must set a PersistentVolume nodeAffinity when using local volumes. The Kubernetes scheduler uses the PersistentVolume nodeAffinity to schedule these Pods to the correct node.

I guess this contradicts our approach. 

I think the other way to achieve this would be to use CSI for specific vendors based on `distribution` param.

examples/enable-persistence-queue/enable-persistence-queue-values.yaml
```diff
@@ -0,0 +1,29 @@
...
+    persistentQueueEnabled:
+      logs: true
+      metrics: true
+      traces: true
+      storagePath: "/var/addon/splunk/persist"
+
+logsEngine: otel
+
+agent:
+  securityContext:
```
jvoravong, line 20: Is the securityContext required?
- If yes, please add a note similiar to [this](https://github.com/signalfx/splunk-otel-collector-chart/blame/08be4e5853a7e3d9ce1d42ec5d755a68ecca53e1/docs/advanced-configuration.md#L526) in your newly added advanced-configuration.md documentation.
- If no, I d remove the securityContext here or leave a note it is optional.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -106,6 +106,20 @@ splunkPlatform:
     #   requests_per_second is the average number of requests per seconds.
     queueSize: 5000
 
+    # When enabled, it uses file_storage extension to persist the queue data.
+    # This can be used as an alternative redundancy mechanism for data being exported.
+    # NOTE: The File Storage extension will persist state to the node's local file system.
+    # While using the persistent queue it is advised to increase memory limit for agent (agent.resources.limits.memory)
+    # to 1Gi.
```
jvoravong, line 113: Please add a "# Refer to https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/advanced-configuration.md#XXX" link to your newly added docs.

docs/advanced-configuration.md
```diff
@@ -695,3 +695,29 @@ rbac:
...
+### Support for persistent queue
+
+* `GKE/Autopilot` and `EKS/Fargate` support
+  * Both of the above distributions doesn't allow volume mounts, as they are kind of `serverless` and we don't manage the underlying infrastructure.
+  * Persistent buffering is not supported for them, as directory needs to be mounted via `hostPath`.
+  * Refer [aws/fargate](https://docs.aws.amazon.com/eks/latest/userguide/fargate.html) and [gke/autopilot](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-security#built-in-security).
+* Gateway support
+  * The filestorage extention acquires an exclusive lock for the queue directory.
+  * It is not possible to run the persistent buffering if there are multiple replicas of a pod and `gateway` runs 3 replicas by default.
+  * If we provide support anyhow, only of those pods will be in running state and the other will be blocked to acquire the lock.
```
jvoravong, line 722:
```suggestion
  * Even if support is somehow provided, only one of the pods will be able to acquire the lock and run, while the others will be blocked and unable to operate.
```

docs/advanced-configuration.md
```diff
@@ -695,3 +695,29 @@ rbac:
...
+  * https://github.com/signalfx/splunk-otel-collector-chart/issues/800
```
jvoravong, line 723: The title of #800 indicates the Cluster receiver crashes with trying to use Persistence queue functionality. Does #861 close #800? Is this line out of date once #861 is merged?

helm-charts/splunk-otel-collector/templates/_helpers.tpl
jvoravong: We should add both documention as well as validation rules for these limitations to ensure the best user experience. Please add validation rules with proper error messages to the values.schema.json file or in the the helm templates using the `fail` function.

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -5,6 +5,13 @@ The values can be overridden in .Values.clusterReceiver.config
 {{- define "splunk-otel-collector.clusterReceiverConfig" -}}
 {{ $clusterReceiver := fromYaml (include "splunk-otel-collector.clusterReceiver" .) -}}
 extensions:
+  {{- if (eq (include "splunk-otel-collector.persistentQueueEnabledLogs" .) "true") }}
+  {{- include "splunk-otel-collector.persistentQueueLogs" (dict "Values" .Values "forAgent" false) | nindent 2 }}
+  {{- end }}
+  {{- if (eq (include "splunk-otel-collector.persistentQueueEnabledMetrics" .) "true") }}
+  {{- include "splunk-otel-collector.persistentQueueMetrics" (dict "Values" .Values "forAgent" false) | nindent 2 }}
+  {{- end }}
+
```
jvoravong, line 14: Nit: extra line

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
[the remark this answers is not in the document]
jinja2: It will be left to the user to set the `storageClass` for the pvc. We don't need to get into the details of the type of storage used in our chart with pvc.

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -160,6 +160,24 @@ spec:
...
+          chmod -v g+rwxs {{ .Values.splunkPlatform.sendingQueue.persistentQueueEnabled.storagePath }}/agent/metrics;
+          setfacl -n -Rm d:m::rx,m::rx,d:g:{{ $agent.securityContext.runAsGroup | default 999 }}:rx,g:{{ $agent.securityContext.runAsGroup | default 999 }}:rx {{ .Values.splunkPlatform.sendingQueue.persistentQueueEnabled.storagePath }}/agent/metrics;
+          {{- end }}
+          {{- if (eq (include "splunk-otel-collector.persistentQueueEnabledTraces" .) "true") }}
+          mkdir -p {{ .Values.splunkPlatform.sendingQueue.persistentQueueEnabled.storagePath }}/agent/traces;
+          chown -Rv {{ $agent.securityContext.runAsUser | default 999 }}:{{ $agent.securityContext.runAsGroup | default 999 }} {{ .Values.splunkPlatform.sendingQueue.persistentQueueEnabled.storagePath }}/agent/traces;
+          chmod -v g+rwxs {{ .Values.splunkPlatform.sendingQueue.persistentQueueEnabled.storagePath }}/agent/traces;
+          setfacl -n -Rm d:m::rx,m::rx,d:g:{{ $agent.securityContext.runAsGroup | default 999 }}:rx,g:{{ $agent.securityContext.runAsGroup | default 999 }}:rx {{ .Values.splunkPlatform.sendingQueue.persistentQueueEnabled.storagePath }}/agent/traces;
```
dmitryax, line 180: You don't need different storages per signal. You can define only one storage and they won't conflict, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queued_retry.go#L135

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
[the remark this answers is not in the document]
jinja2: @dmitryax I think you mentioned that this PR might be more of a parity thing for sck to otel, in which case should we keep the clusterReceiver out of this? If desired we can create a different issue to explore how best to add persistent queues for deployment (sts + pvc would be ideal)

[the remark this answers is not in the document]
dmitryax: SGTM, let's exclude it for now and put a not in the values.yaml that the persistence is not applicable to the k8s cluster metrics and k8s events. The VPC's approach can be handled after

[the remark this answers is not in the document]
VihasMakwana: Okay.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -106,6 +106,20 @@ splunkPlatform:
...
+    persistentQueueEnabled:
```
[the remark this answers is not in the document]
VihasMakwana: Done, looks a lot cleaner now.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -616,13 +620,13 @@ exporters:
...
-  {{- include "splunk-otel-collector.splunkPlatformLogsExporter" . | nindent 2 }}
+  {{- include "splunk-otel-collector.addPersistentStorage" (dict "exporter" (include "splunk-otel-collector.splunkPlatformLogsExporter" .) "isEnabled" (include "splunk-otel-collector.persistentQueueEnabled" .)) | nindent 2 }}
   {{- end }}
   {{- if (eq (include "splunk-otel-collector.platformMetricsEnabled" .) "true") }}
-  {{- include "splunk-otel-collector.splunkPlatformMetricsExporter" . | nindent 2 }}
+  {{- include "splunk-otel-collector.addPersistentStorage" (dict "exporter" (include "splunk-otel-collector.splunkPlatformMetricsExporter" .) "isEnabled" (include "splunk-otel-collector.persistentQueueEnabled" .)) | nindent 2 }}
   {{- end }}
   {{- if (eq (include "splunk-otel-collector.platformTracesEnabled" .) "true") }}
-  {{- include "splunk-otel-collector.splunkPlatformTracesExporter" . | nindent 2 }}
+  {{- include "splunk-otel-collector.addPersistentStorage" (dict "exporter" (include "splunk-otel-collector.splunkPlatformTracesExporter" .) "isEnabled" (include "splunk-otel-collector.persistentQueueEnabled" .)) | nindent 2 }}
```
dmitryax, line 629: Can you modify `splunk-otel-collector.splunkPlatformTracesExporter` and other helpers instead of introducing `splunk-otel-collector.addPersistentStorage` helper? This is a scenario when following the DRY principle makes the code too hard to read.

examples/disable-persistence-queue-logs/disable-persistence-queue-logs-values.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+agent:
+  securityContext:
+    runAsUser: 999
+    runAsGroup: 999
+  config:
+    exporters:
+       splunk_hec/platform_logs:
+         sending_queue:
+           storage: null
```
dmitryax, line 25: Let's not create an example per each possible configuration. It's too much to maintain and not that helpful for users. Can we have only one example where we enable the persistent queue for logs and traces and disable it for metrics (seems like the most reasonable scenario)?

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -438,6 +438,14 @@ Whether clusterReceiver should be enabled
 {{- end -}}
 
 
+{{/*
+Whether persistentQueue should be enabled
+*/}}
+{{- define "splunk-otel-collector.persistentQueueEnabled" }}
+{{- .Values.splunkPlatform.sendingQueue.persistentQueue.enabled -}}
+{{- end }}
```
dmitryax, line 446: No need to introduce a helper for this. Just use `.Values.splunkPlatform.sendingQueue.persistentQueue.enabled`

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -235,6 +235,12 @@ filter/logs:
           value: "true"
 {{- end }}
 
+{{- define "splunk-otel-collector.persistentQueue" -}}
+file_storage/persistent_queue:
+  directory: {{ .Values.splunkPlatform.sendingQueue.persistentQueue.storagePath }}/agent
+  timeout: 0
+{{- end }}
```
dmitryax, line 242: It should not go to common if it's used in the agent only

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -198,11 +198,12 @@ service:
...
-  extensions: [health_check, memory_ballast, k8s_observer]
-  {{- else }}
-  extensions: [health_check, memory_ballast]
-  {{- end }}
+  extensions:
+    - health_check
+    - memory_ballast
+    {{- if eq (include "splunk-otel-collector.distribution" .) "eks/fargate" }}
+    - k8s_observer
+    {{- end }}
```
dmitryax, line 206: Nit: let's avoid unrelated changes

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -106,6 +106,17 @@ splunkPlatform:
     #   requests_per_second is the average number of requests per seconds.
     queueSize: 5000
 
+    # When enabled, it uses file_storage extension to persist the queue data.
+    # This can be used as an alternative redundancy mechanism for data being exported.
```
dmitryax, line 110:
```suggestion
    # This option enables the persistent queue to store data on the disk instead of memory before sending it to the backend.
    # It allows setting higher queue limits and preserving the data across restarts of the collector container.
```

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -106,6 +106,17 @@ splunkPlatform:
...
+    # NOTE: The File Storage extension will persist state to the node's local file system.
+    # While using the persistent queue it is advised to increase memory limit for agent (agent.resources.limits.memory)
+    # to 1Gi.
+    # Refer to: https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/advanced-configuration.md#data-persistence
+    persistentQueue:
+      # Specifies whether to persist log/metric/trace data.
+      enabled: false
+      storagePath: "/var/addon/splunk/persist"
```
dmitryax, line 118: Why does it have to be `persist`? Sounds a bit ambiguous. Maybe `/var/addon/splunk/exporter_queue`?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
VihasMakwana: Well, here's a catch.
As it's only used in agent, modifying `splunk-otel-collector.splunkPlatformTracesExporter` directly without any additional data would be not possible. We use the same helper for the gateway.
We don't know which pipeline (agent or gateway) called `splunk-otel-collector.splunkPlatformTracesExporter`. 
I will try to implement in a much cleaner way.

VihasMakwana: I did try something cleaner, https://github.com/signalfx/splunk-otel-collector-chart/pull/861/files#diff-55f9d185ea0bc213a43cf77ba9ebcba2f7ffd650e9535fba0bae53706fd45825R624.
I set a variable in the current scope before calling `splunkPlatformTracesExporter` and family. That way those helpers would know whether to include the `storage` or not. 
Let me know if that's fine by you.

VihasMakwana: For clusterReceiver and gateway, we don't set that field. 
We only set it for the agent and unset it once we're done. 
Kind of weird but does the work.

VihasMakwana: I find it much cleaner than passing extra variables to `include splunk-otel-collector.splunkPlatformTracesExporter` for all three exporters.

examples/disable-persistence-queue-traces/disable-persistence-queue-traces-values.yaml
```diff
@@ -0,0 +1,30 @@
...
+    exporters:
+       splunk_hec/platform_traces:
+         sending_queue:
+           storage: null
+
+clusterReceiver:
+  enabled: true
+  securityContext:
+    runAsUser: 999
+    runAsGroup: 999
```
dmitryax, line 30: Why is this added? seems unrelated

examples/disable-persistence-queue-traces/disable-persistence-queue-traces-values.yaml
```diff
@@ -0,0 +1,30 @@
...
+  sendingQueue:
+    persistentQueue:
+      enabled: true
+
+logsEngine: otel
+
+agent:
+  securityContext:
+    runAsUser: 999
+    runAsGroup: 999
```
dmitryax, line 19: Same Q here. I'd suggest keeping each example solving a specific problem

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: This looks good to me. Thank you!

dmitryax:
The PR LGTM overall. just couple nits

examples/enable-persistence-queue/README.md
```diff
@@ -0,0 +1,8 @@
+# Example of chart configuration
+
+## Enable Persistent Queue for logs, metrics and traces
+
+This example will show how to enable data persistence for log/metric/trace data. Data will be persisted in node's local filesystem. 
+Persistent queue will keep track of unexported data. It will continue from previously saved offsets, if any, after collector restarts and export it to Splunk platform.
+
+Refer to: https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/advanced-configuration.md#data-persistence
```
dmitryax, line 8: I'm not sure if this example is needed. I'd rather put more comments in the other example. e.g.

```
# If you want to disable the persistent queue in a particular pipeline, you can do so using the following config
# update splunk_hec/platform_traces exporter to use memory queue instead
agent:
  config:
    exporters:
       splunk_hec/platform_traces:
         sending_queue:
           storage: null
```

VihasMakwana: yeah, we can do that. but I added one for enabling and one for disabling, just in case.

VihasMakwana:
@dmitryax I added functional tests covering persistent storage, I guess it's helpful.

dmitryax (approved):

jvoravong (approved):

examples/enable-persistence-queue/README.md
jvoravong: Added support for [no-render](https://github.com/signalfx/splunk-otel-collector-chart/blame/main/examples/DEVELOPER_GUIDE.md) examples which could be useful for examples like this in the future.

Issue #868, opened 2023-08-02
Fresh install fails since chart v0.77.0


sleterrier:
## Description
Initial `helm install` fails since chart v0.77.0.

## Steps to Reproduce
```
helm install splunk-collector splunk-otel-collector-chart/splunk-otel-collector -f values.yaml --version 0.77.0
```

## Expected Result
`helm install` should succeed.

## Actual Result
Fails secret validation due to missing serviceAccount.

### Chart version

0.77.0+

## Environment
Cloud: EKS + vanilla Kubernetes
k8s version: 1.25

### Chart configuration

```yaml
agent:
  controlPlaneMetrics:
    apiserver:
      enabled: false
    controllerManager:
      enabled: false
    coredns:
      enabled: false
    etcd:
      enabled: false
    proxy:
      enabled: false
    scheduler:
      enabled: false
  enabled: true
  resources:
    limits:
      cpu: 1
      memory: 500Mi
    requests:
      cpu: 250m
      memory: 500Mi
certmanager:
  enabled: false
  installCRDs: false
clusterName: cluster-name
clusterReceiver:
  enabled: false
environment: dev
extraAttributes:
  fromAnnotations:
  - from: namespace
    key_regex: (example.net/.*)
    tag_name: $$1
  - from: pod
    key_regex: (example.net/.*)
    tag_name: $$1
  fromLabels:
  - from: namespace
    key_regex: (example.net/.*)
    tag_name: $$1
  - from: pod
    key_regex: (example.net/.*)
    tag_name: $$1
logsCollection:
  containers:
    enabled: true
    useSplunkIncludeAnnotation: true
logsEngine: otel
secret:
  create: false
  name: splunk-otel-collector
serviceAccount:
  create: true
  name: splunk-collector
splunkPlatform:
  endpoint: https://http-inputs-obfuscated.splunkcloud.com/services/collector
  index: kubernetes_logs
  logsEnabled: true
  metricsEnabled: false
  tracesEnabled: false
```

### Log output

```shell
Error: INSTALLATION FAILED: failed pre-install: warning: Hook pre-install splunk-otel-collector/templates/secret-splunk-validation-hook.yaml failed: pods "splunk-collector-splunk-otel-collector-validate-secret" is forbidden: error looking up service account splunk/splunk-collector: serviceaccount "splunk-collector" not found
```

### Additional context

https://github.com/signalfx/splunk-otel-collector-chart/pull/781 is likely the culprit.

[...]

Issue #780, opened 2023-05-09
Missing ServiceAccountName configuration for secret-splunk-validation-hook pod


philipsabri:
### Describe the issue you're reporting

Hi, I believe we're missing the ``ServiceAccountName`` configuration for the  [secret-splunk-validation-hook pod](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/secret-splunk-validation-hook.yaml). Is there a reason it hasn't been added to the pod?

Since the  ``imagePullSecrets`` are [added to the ServiceAccount](https://github.com/signalfx/splunk-otel-collector-chart/blob/b36acaaa7b140170ea454ab905049c0f93e187cf/helm-charts/splunk-otel-collector/templates/serviceAccount.yaml#L3-L8), it would be helpful to use it on the hook pod when pulling the image from a private registry.

jvoravong:
Sorry I didn't notice automation closed this ticket.

Updates:
- Verified in the latest version of OpenShift using the default service account works with the validation hook pod, I was concerned it wouldn't since they have higher security standards.
- Created the potential solution I was thinking of (link above) for this issue. While this solution would work I did find some drawbacks and don't think this is the way to go at this time. We might need it in the future.
- After more research, I do think we can go forward with the following fix.
  - Remove the service account in the validation hook
  - If the user provided [imagePullSecrets](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret), add them to the validation hook pod spec.
  -  @omrozowicz-splunk feel free to tackle this if you have time.

PR #871, opened 2023-08-02
fix for non-existing service account in secret validation hook

omrozowicz-splunk:
This is regarding this issue: https://github.com/signalfx/splunk-otel-collector-chart/issues/868

`default-splunk-otel-collector-validate-secret` is being run before service account is created, so every config like this:
```
secret:
  create: false
  name: splunk-otel-collector
serviceAccount:
  create: true
  name: splunk-collector
```
will result in ERROR `Error: INSTALLATION FAILED: failed pre-install: warning: Hook pre-install splunk-otel-collector/templates/secret-splunk-validation-hook.yaml failed: pods "splunk-collector-splunk-otel-collector-validate-secret" is forbidden: error looking up service account splunk/splunk-collector: serviceaccount "splunk-collector" not found
`

We shouldn't explicitly demand having serviceAccount created beforehand either, as this is not the only way to access secrets. If the default namespace service account has access to namespace secrets and we won't put any `serviceAccount` in pre-hook spec, the validation will finish successfully.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/secret-splunk-validation-hook.yaml
+++ b/helm-charts/splunk-otel-collector/templates/secret-splunk-validation-hook.yaml
@@ -11,9 +11,7 @@ metadata:
     "helm.sh/hook": pre-upgrade,pre-install
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 spec:
-{{- if not .Values.serviceAccount.create }}
   serviceAccountName: {{ template "splunk-otel-collector.serviceAccountName" . }}
-{{- end }}
   restartPolicy: Never
   containers:
   - name: validate-secret
--- a/examples/secret-validation/README.md
+++ b/examples/secret-validation/README.md
@@ -1,0 +1,15 @@
+# Example of chart configuration
+
+## Use existing secret
+
+This example shows how reuse existing secret with o11y token. In this case we use the default
+`serviceAccount` settings and assume it is enough to access the `secret`.
+
+Sometimes we might need to use existing `serviceAccount` to be able to read the secret, then
+the configuration must be:
+
+```yaml
+serviceAccount:
+  create: false
+  name: name-of-existing-sa
+```
--- a/examples/secret-validation/rendered_manifests/clusterRole.yaml
+++ b/examples/secret-validation/rendered_manifests/clusterRole.yaml
@@ -1,0 +1,83 @@
+---
+# Source: splunk-otel-collector/templates/clusterRole.yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: default-splunk-otel-collector
+  labels:
+    app.kubernetes.io/name: splunk-otel-collector
+    helm.sh/chart: splunk-otel-collector-0.82.0
+    app.kubernetes.io/managed-by: Helm
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/version: "0.82.0"
+    app: splunk-otel-collector
+    chart: splunk-otel-collector-0.82.0
+    release: default
+    heritage: Helm
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - events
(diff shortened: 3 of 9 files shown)


helm-charts/splunk-otel-collector/templates/secret-splunk-validation-hook.yaml
```diff
@@ -11,7 +11,9 @@ metadata:
     "helm.sh/hook": pre-upgrade,pre-install
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 spec:
+{{- if not .Values.serviceAccount.create }}
   serviceAccountName: {{ template "splunk-otel-collector.serviceAccountName" . }}
```
jinja2, line 15: I think we can just remove the `serviceAccountName` from this pod. Let k8s attach the `default` serviceAccount since the commands run here to test the token does not need any k8s api access.

omrozowicz-splunk: This is where we started, but it was added after this issue was raised: https://github.com/signalfx/splunk-otel-collector-chart/issues/780

jinja2: Ah, interesting, so there are users pulling images from private registries, makes sense to keep the option then.

jinja2 (approved):

helm-charts/splunk-otel-collector/templates/secret-splunk-validation-hook.yaml
jvoravong: The current changes would likely be breaking for some users using pullimagesecrets for private image repos. We can likely create a solution that supports all the mentioned use cases by adding more [helm hooks](https://github.com/signalfx/splunk-otel-collector-chart/pull/871/files/281423b71b2e571324d9c0f007e3782abc7c177b#diff-335f8a73653dc7fd765c77f8a613d9c8f263065bd6cab6fc2ba5d75feb204ef0). Let me see what is possible here shortly unless someone else does it faster.

omrozowicz-splunk: why is this a breaking change? Currently only using the existing serviceAccount (when `.Values.serviceAccount.create` is `false`) works for this case, so this change won't break anything for the users with such configuration I think.

jvoravong: - The update in  https://github.com/signalfx/splunk-otel-collector-chart/issues/780, which added the ServiceAccountName setting for the secret-splunk-validation-hook pod, has caused the https://github.com/signalfx/splunk-otel-collector-chart/issues/868 bug for customers who rely on the default service account with a custom secret. 
- Some customers utilize private image repositories, which mirror the splunk-otel-collector image from https://quay.io/repository/signalfx/splunk-otel-collector, and these repositories require custom credentials and a  custom service account. 
  - A failure can arise when the validation check. If a customer is using a custom service account and secret, the validation check might fail because it can't the splunk-otel-collector image. This is because the image pull secret attached to the custom service account is needed to retrieve the splunk-otel-collector image from the private repository.
- A potential solution could involve the creation of a secondary (derivative) service account that has the same lifecycle of the validation pod. This account would only be created and utilized when .Values.serviceAccount.create=true is set, and would be exclusively used by the validation pod.

jvoravong: For this PR, how about we still add the example and remove the changes in the pod spec. I'll open up a new PR in the next day for adding the secondary (derivative) service account and add @omrozowicz-splunk as a reviewer.

omrozowicz-splunk: done, I ran `make render` again, now it renders the config with `serviceAccount`

jvoravong (approved):

Issue #439, opened 2022-04-27
Fix the secret name to be able to install more than one chart in the same cluster 


dmitryax:
Currently secret name is build from [Chart name](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/_helpers.tpl#L145), but it has to be build from Release name to not introduce conflicts between different installations of the same chart. Similar way as done for other generated resources.

PR #873, opened 2023-08-03
Fix secret name conflicts

VihasMakwana:
Fixes https://github.com/signalfx/splunk-otel-collector-chart/issues/439.


helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -149,7 +149,12 @@ Define name for the Splunk Secret
 {{- if .Values.secret.name -}}
 {{- printf "%s" .Values.secret.name -}}
 {{- else -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{ $name := default .Chart.Name .Values.nameOverride -}}
```
jinja2, line 152: This logic is already defined in the template `splunk-otel-collector.fullname`. I would suggest just using the existing template here instead of c+p the same logic.

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -160,7 +165,12 @@ Define name for the etcd Secret
 {{- if .Values.agent.controlPlaneMetrics.etcd.secret.name -}}
 {{- printf "%s" .Values.agent.controlPlaneMetrics.etcd.secret.name -}}
 {{- else -}}
-{{- default .Chart.Name .Values.nameOverride | printf "%s-etcd" | trunc 63 | trimSuffix "-" -}}
+{{ $name := default .Chart.Name .Values.nameOverride -}}
```
jinja2, line 168: same as above, you can use the existing template

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -149,7 +149,12 @@ Define name for the Splunk Secret
 {{- if .Values.secret.name -}}
 {{- printf "%s" .Values.secret.name -}}
 {{- else -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{ $name := (include "splunk-otel-collector.name" .) -}}
+{{- if contains $name .Release.Name -}}
+{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
```
jinja2, line 157:
```suggestion
{{ default (include "splunk-otel-collector.fullname" .) .Values.secret.name }}
```
I think you can just replace directly with the helper function, and can also remove the if/else before this. `splunk-otel-collector.fullname` already evaluates the same conditions with the addition of also respecting the `fullNameOverride` variable which is missing here.

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -149,7 +149,8 @@ Define name for the Splunk Secret
 {{- if .Values.secret.name -}}
 {{- printf "%s" .Values.secret.name -}}
 {{- else -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
```
jinja2, line 153: @VihasMakwana my comment [here](https://github.com/signalfx/splunk-otel-collector-chart/pull/873#discussion_r1287273674) is still relevant wrt the new changes. The helper `splunk-otel-collector.fullname` will make certain the secret name aligns correctly with the release name and the if/else can be dropped in favor of using the `default` function.

VihasMakwana: I'm not sure if we can directly use `splunk-otel-collector.fullname`, as it takes `.Values.fullnameOverride` into account. We don't have any reference to `.Values.fullnameOverride` in the existing logic of `splunk-otel-collector.secret`.
What we can do is separate out [this block](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/_helpers.tpl#L19-L23 ) into a separate helper and use that instead.

VihasMakwana: But this change would be more like a refactoring, rather than a bug fix.
I would like to keep it in separate PR, if that's fine by you.

jinja2: The `fullnameOverride` variable is meant to be used when a user wants to intentionally replace the complete name of the objects generated as part of the chart. We use the `fullname` helper function to name all other objects already and this gives user the option to change the names of the objects in a consistent manner across the board. So I don't think we are changing the scope of the original issue by using the fullname helper here.

jinja2 (changes requested):

jinja2 (approved):

dmitryax (approved):

dmitryax:
@VihasMakwana please run `make render`

VihasMakwana:
@dmitryax I think we are good to merge.

PR #886, opened 2023-08-14
Add support for "no render" examples, refactor the kafkametrics receiver example

jvoravong:
Description:
- Added the ability to support "no render" examples to avoid bloating git diffs over time. The "no render" examples are still validated to work with "helm template" commands.
- Added a DEVELOPER_GUIDE.md for the examples.
- Refactored the kafkametrics receiver to be more standardized. Included changes so the Kafka demo is more portable.


examples/DEVELOPER_GUIDE.md
```diff
@@ -0,0 +1,44 @@
...
+examples that are derivatives or for any other reason you might not want them rendered.
+
+### Derivative Examples and Code Review Efficiency
+
+When adding derivative examples that showcase specific cases, it's recommended to use the `-values.norender.yaml`
+suffix for the filename. By doing this, the example won't be rendered by the script, ensuring that
+code review diffs remain compact and focused. This practice prevents the inclusion of derivative
+rendered manifests, which can bloat the diffs and make reviews more challenging.
+
+### Adding Examples Without Rendering
```
jinja2, line 40: Imo, we should keep the rendered templates. It allows us to track changes to the examples which we might not catch otherwise when updating the chart. But I don't see any harm in having this capability in the script.

jinja2: Sounds good to me

jinja2 (approved):

PR #887, opened 2023-08-15
DOC Update: Log collection and running the container in non-root user mode

jvoravong:


docs/advanced-configuration.md
```diff
@@ -514,10 +514,18 @@ Here is the summary of performance benchmarks run internally.
...
+agent:
+  securityContext:
+     runAsUser: 20000
+     runAsGroup: 20000
+```
 
-Collecting logs often requires reading log files that are owned by the root user. By default, the container runs with `securityContext.runAsUser = 0` which gives the `root` user permission to read those files. To run the container in `non-root` user mode, set `.agent.securityContext` to `20000` to cause the container to run the required file system operations as UID and GID `20000`. (it can be any other UID & GUI)
-
-Note: `cri-o` container runtime did not work during internal testing.
+Note: Running the collector agent for log collection in non-root mode is currently unsupported in CRI-O and OpenShift environments due to platform limitations, for more details see the related
```
jinja2, line 526: Is this valid for Openshift given we create our own security context constraint with `RunAsUser` set to `RunAsAny`? The [docs on scc](https://docs.openshift.com/container-platform/4.13/authentication/managing-security-context-constraints.html#authorization-SCC-strategies_configuring-internal-oauth) imply we can set the runAsUser config.

Re: CRI-O - Did we try it with a cri-o openshift cluster (that's the only distro using cri-o I can think of)? To my knowledge, cri-o and containerd should have similar functionalities for runAsUser

jvoravong: - We can run cri-o in minikube, kops, and OpenShift. Unfortunately the cri-o minikube driver doesn't function on arm machines.
- Tested the issue outlined in the doc on the latest OpenShift v4.13 cluster, this issue is still real and valid. This issue happens with both otel and fluentd as the log engine. This issue was non-existed on the latest versions of docker and containerd.
- From my research, the root of the issue is initContainers and containers have the same permissions when using cri-o or OpenShift. We would likely have to use a second daemonset or other similar Kubernetes object to update the permissions of all the log data on every host so that the agent daemonset could use the non-root user mode and read the log data on all nodes.

jinja2: I don't think the reason mentioned here is the problem. I did a quick test with cri-o/openshift, and was able to set separate users for init (root) and main containers (random uid). Maybe we just document that it is not supported in the chart and not due to a platform limitation? We can have an issue to fix this for openshift and triage that separately.

jinja2 (approved):

PR #906, opened 2023-09-01
Enable native OTel logs collection in GKE Autopilot

dmitryax:
Native OTel logs collection is now available in GKE Autopilot. The `--metrics-addr` hack still required if fluent is enabled.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/daemonset.yaml
+++ b/helm-charts/splunk-otel-collector/templates/daemonset.yaml
@@ -232,7 +232,7 @@ spec:
         {{- else }}
         - /otelcol
         - --config=/conf/relay.yaml
-        {{- if eq .Values.distribution "gke/autopilot" }}
+        {{- if and (eq .Values.distribution "gke/autopilot") (eq .Values.logsEngine "fluentd") }}
         {{- /* Temporary no-op argument required to run collector on GKE Autopilot */}}
         - --metrics-addr=0.0.0.0:8889
         {{- end }}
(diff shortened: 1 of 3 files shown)


helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -232,7 +232,7 @@ spec:
         {{- else }}
         - /otelcol
         - --config=/conf/relay.yaml
-        {{- if eq .Values.distribution "gke/autopilot" }}
+        {{- if and (eq .Values.distribution "gke/autopilot") (eq .Values.logsEngine "fluentd") }}
         {{- /* Temporary no-op argument required to run collector on GKE Autopilot */}}
         - --metrics-addr=0.0.0.0:8889
```
jinja2, line 237: No related to this PR, but out of curiosity, could this be localhost? I assume this is the listener addr for prometheus endpoint and will be scraped by the locally running collector.

jinja2 (approved):

helm-charts/splunk-otel-collector/templates/daemonset.yaml
dmitryax: This is a noop. The collector ignores this argument. It's just need to bypass the allowlist

dmitryax: But your Q still holds true. It's just provided in the config now https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl#L646. I think you're right it should be localhost. We should discuss it more.

PR #917, opened 2023-09-07
Add ability to update and track operator auto-instrumentation images

jvoravong:
### Description
- Introduced the ability to automatically track both native OpenTelemetry and Splunk instrumentation libraries within this Helm chart project.
  - Rationale: Due to the limitations of fetching native OpenTelemetry instrumentation library information, we've implemented CI/CD workflows to automate this process. This ensures that every time the operator subchart is updated, the corresponding instrumentation library information in the `values.yaml` file is also updated.
  - A running operator doesn't disclose the version information of the instrumentation libraries until deployed. Helm release values cannot be dynamically updated after an install/upgrade to include this information, necessitating this enhancement.
  - No new instrumentation libraries were added. This update merely makes explicit what was previously implicit, enabling easier tracking and management ([example](https://github.com/signalfx/splunk-otel-collector-chart/pull/917/files#diff-1c372e398662351c026254d2137a8912a34e16503f4edefc76d05274e8fa2391)). 
  - With the instrumentation libraries being explicitly defined, we can now track them in version control and use the otel resource attribute _splunk.zc.method_ to track usage.
- Began refactoring the GitHub workflow to make the code reusable both in GitHub Actions and locally. Added a `--debug` parameter for debugging purposes.

### Testing
- Validated by deploying the operator with the latest changes. Ensured that the `OTEL_RESOURCE_ATTRIBUTES` and `OTEL_EXPORTER_OTLP_ENDPOINT` values were correctly managed.

### Documentation
- Increased inline code documentation and comments for better maintainability.
- Updated `NOTES.txt` to provide users with clearer information regarding the status and capabilities of auto-instrumentation components.

### DIFF
--- a/.github/workflows/update_chart_dependencies.yaml
+++ b/.github/workflows/update_chart_dependencies.yaml
@@ -1,10 +1,23 @@
 name: Check for new chart dependency updates
 
+# Description:
+# This workflow automates the process of checking for and updating Helm chart dependencies.
+# Specifically, it:
+#   1. Checks for new versions of (subchart) dependencies listed in chart.yaml.
+#   2. Updates chart.yaml with new versions where applicable.
+#   3. If the 'opentelemetry-operator' subchart is updated in chart.yaml, it also updates related
+#      image tags in values.yaml.
+
 on:
   schedule:
     # Run every Monday at noon.
     - cron: "0 12 * * 1"
   workflow_dispatch:
+    inputs:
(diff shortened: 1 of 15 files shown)


jvoravong:
> I hate how much shell is needed but it's some of the most readable shell I've ever seen

Agreed. My longterm vision is to centralize the non-trivial shell code as much as possible so it can easily be executed from the Makefile.

PR #937, opened 2023-09-22
Change default logs collection engine to OpenTelemetry

dmitryax:
Change the default logs collection engine (`logsEngine`) to the native OpenTelemetry logs collection (`otel`).

If you want to keep using Fluentd sidecar for the logs collection, set `logsEngine: fluentd` in your values.yaml.


atoulme (approved):

jinja2:
There are a couple of examples, specifically `fluentd-multiline-logs-java-stack-traces` and `only-logs-fluentd` which might need to be updated with the `logsEngine: fluentd` in values.yaml or we update the use case.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -164,12 +164,10 @@ splunkObservability:
...
 # - `otel`: utilize native OpenTelemetry log collection.
 #
-# Change it to `otel` to get higher throughput performance and avoid installing
-# an extra container for fluentd.
+# `fluentd` will be deprecated soon, so it's recommended to use `otel` instead.
 ################################################################################
 
-# The default value is "fluentd" which will be changed to "otel" in 0.86.0.
-logsEngine: ""
+logsEngine: otel
```
jinja2, line 170: should we bump the default resource limits with the logs collection being done by the collector container? I didn't see any guidelines for how much more resources the collector container should be provided

examples/add-filter-processor/rendered_manifests/daemonset.yaml
```diff
@@ -47,74 +46,41 @@ spec:
         - effect: NoSchedule
           key: node-role.kubernetes.io/control-plane
       initContainers:
-        - name: prepare-fluentd-config
-          image: splunk/fluentd-hec:1.3.3
+        - name: migrate-checkpoint
+          image: quay.io/signalfx/splunk-otel-collector:0.85.0
           imagePullPolicy: IfNotPresent
-          command: [ "sh", "-c"]
+          command: ["/migratecheckpoint"]
```
jinja2, line 52: I am not aware how migratecheckpoint works, but from the context I assume this is used to switch the fluentd written position tracking file to otel tracking file. Would there be any issues if this is run on a node which didn't have fluentd running? Basically if `/var/log/splunk-fluentd-containers.log.pos` does not exist. Along the same lines, would this clean up the fluents pos file after it migrates to otel the 1st time? Should this be configurable then? A user can remove this extra initContainer after the first helm upgrade with this switch is deployed.

helm-charts/splunk-otel-collector/values.yaml
dmitryax: I think we should keep it as is until we find out that the current defaults are too low. Logs collection is disabled by default for o11y. For `splunkPlatform`, it always depends on the volume. From my tests, I remember it was not that significant increase after switching from fluentd.

dmitryax:
> There are a couple of examples, specifically fluentd-multiline-logs-java-stack-traces and only-logs-fluentd which might need to be updated with the logsEngine: fluentd in values.yaml or we update the use case.

Thanks, I updated them

examples/add-filter-processor/rendered_manifests/daemonset.yaml
dmitryax: > Would there be any issues if this is run on a node which didn't have fluentd running? 

No, the script just exists in that case https://github.com/signalfx/splunk-otel-collector/blob/894d7ef2564c709c92fb94e4a71ced970020d475/cmd/migratecheckpoint/main.go#L175

> Along the same lines, would this clean up the fluents pos file after it migrates to otel the 1st time? Should this be configurable then? A user can remove this extra initContainer after the first helm upgrade with this switch is deployed.

I don't think we need to complicate that just to avoid running the initContainer. It's the same container as the agent, and it returns right away if the otel checkpointing file already exists https://github.com/signalfx/splunk-otel-collector/blob/894d7ef2564c709c92fb94e4a71ced970020d475/cmd/migratecheckpoint/main.go#L168

jinja2: ok, looks like an innocuous addition then

helm-charts/splunk-otel-collector/values.yaml
jinja2: Ok, hopefully users are mindful of setting the limits to match their usage.

jinja2 (approved):

Issue #938, opened 2023-09-27
Helm chart not working if profiling and operator enabled without further configuration.


conrad784:
## Description
Upgrading from 0.84 to 0.85 leads to errors in our deployment.
There is a "breaking change" but this should according to the changelog only impact if we would specify our instrumentation images. We are not using any custom instrumentation images.
We have enables profiling and also enabled the operator without any kustomizations.

## Steps to Reproduce
wget https://github.com/signalfx/splunk-otel-collector-chart/releases/download/splunk-otel-collector-0.85.0/splunk-otel-collector-0.85.0.tgz
tar xfz splunk-otel-collector-0.85.0.tgz

helm template splunk-otel-collector splunk-otel-collector --namespace otel -f otel.yml --debug

## Expected Result
rendered helm chart yaml

## Actual Result
install.go:200: [debug] Original chart version: ""
install.go:217: [debug] CHART PATH: /tmp/otel-debug/splunk-otel-collector

Error: template: splunk-otel-collector/templates/operator/instrumentation.yaml:30:11: executing "splunk-otel-collector/templates/operator/instrumentation.yaml" at <not>: wrong number of args for not: want 1 got 3
helm.go:84: [debug] template: splunk-otel-collector/templates/operator/instrumentation.yaml:30:11: executing "splunk-otel-collector/templates/operator/instrumentation.yaml" at <not>: wrong number of args for not: want 1 got 3

### Chart version

0.85.0

## Environment
helm version
version.BuildInfo{Version:"v3.12.3", GitCommit:"3a31588ad33fe3b89af5a2a54ee1d25bfe6eaa5e", GitTreeState:"clean", GoVersion:"go1.20.7"}
OS: MacOS ARM + Linux x86

### Chart configuration

```yaml
# test.yml 
---
cloudProvider: "gcp"
distribution: ""
clusterName: "test"
splunkPlatform:
  endpoint: "https://example.com"

splunkObservability:
  realm: "eu0"

  profilingEnabled: true

  # Options to disable or enable particular telemetry data types.
  metricsEnabled: true
  tracesEnabled: true
  logsEnabled: false

gateway:
  enabled: false

tolerations:
  - operator: "Exists"
    effect: "NoExecute"
  - operator: "Exists"
    effect: "NoSchedule"

agent:
  config:
    exporters:
      signalfx:
        include_metrics:
          - metric_names:
            - k8s.node.cpu.utilization
            - k8s.pod.cpu.utilization
[...]

PR #943, opened 2023-10-03
Fix Operator helm template issues

jvoravong:
**Description:**

This PR resolves multiple smaller issues in the Helm template related to the of non-default Operator auto-instrumentation values, which previously could trigger deployment failures.

[Link to Splunk Idea: #938](<insert_link_here>) (Refers to Issue 1)

**Issue 1:**
- **Affected Values:**
    - `splunkObservability.profilingEnabled`
- **Root Issue:**
    - The helper `splunk-otel-collector.operator.extract-name-keys-from-dict-list` returns a list that wasn’t being casted accurately where used, necessitating a `toYaml` or `toJson` for proper interpretation.
- **Fix:**
    - Refactor `splunk-otel-collector.operator.extract-name-keys-from-dict-list` to `splunk-otel-collector.operator.env-has` to exclusively return "true" or "false" string values.

**Issue 2:** 
- **Affected Values:**
    - `operator.instrumentation.spec.env`
    - `operator.instrumentation.spec.sampler`
- **Root Issue:**
    - Incorrect usage of the `include` function; a combination of "with" + "toYaml" should have been utilized instead.
- **Fix:**
    - Employ the appropriate "with" + "toYaml" mechanism instead of the "include" function.

**Issue 3:** 
- **Affected Values:**
    - `operator.instrumentation.spec.{instrumentation_library}.env`
- **Root Issue:**
    - Environment variable values equal to "true" and "false" (strings) are being converted into booleans, presumably due to the involved `printf` function. Environment variables set in `opentelemetry.io/v1alpha1` Instrumentation must be strings.
- **Fix:**
    - Guarantee that quotes are rendered in the final manifests for environment variables using the quote function.

**Testing:** 
The new example has been tested; it is both renderable and deployable locally.


helm-charts/splunk-otel-collector/templates/operator/instrumentation.yaml
```diff
@@ -19,29 +19,30 @@ spec:
     - tracecontext
     - baggage
     - b3
-  {{- if .Values.operator.instrumentation.spec.sampler }}
-  {{ include .Values.operator.instrumentation.spec.sampler }}
+  {{- with .Values.operator.instrumentation.spec.sampler }}
+  sampler:
+  {{- toYaml . | nindent 4 }}
```
dmitryax, line 24: This seems like a fix of another issue? If so, please submit a separate PR. Let's try to keep the changeset serving one urpose

jvoravong: This is needed to fix the regression. I'll fix only the regressions and move the new example out to a new PR.

dmitryax: I understand that this is a regression, but it's another regression that is unrelated to the reported issue which is meant to be fixed in this PR, right? The reported issue is about `instrumentation.spec.env` not `instrumentation.spec.sampler`

helm-charts/splunk-otel-collector/templates/operator/instrumentation.yaml
```diff
@@ -19,29 +19,30 @@ spec:
...
-    {{- if .Values.operator.instrumentation.spec.env }}
-    {{- include .Values.operator.instrumentation.spec.env }}
-    {{- end }}
-    {{- if .Values.splunkObservability.profilingEnabled }}
-    {{- if not hasKey (include "splunk-otel-collector.operator.extract-name-keys-from-dict-list" .Values.operator.instrumentation.spec.env) "SPLUNK_PROFILER_ENABLED" }}
+  {{- with .Values.operator.instrumentation.spec.env }}
+  {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- if .Values.splunkObservability.profilingEnabled }}
+  {{- if eq (include "splunk-otel-collector.operator.env-has" (dict "env" .Values.operator.instrumentation.spec.env "envName" "SPLUNK_PROFILER_ENABLED")) "false" }}
```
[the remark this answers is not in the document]
jvoravong: Fixed

helm-charts/splunk-otel-collector/templates/operator/instrumentation.yaml
```diff
@@ -19,29 +19,30 @@ spec:
...
     - name: SPLUNK_PROFILER_ENABLED
       value: "true"
-    {{- end }}
-    {{- if not hasKey (include "splunk-otel-collector.operator.extract-name-keys-from-dict-list" .Values.operator.instrumentation.spec.env) "SPLUNK_PROFILER_MEMORY_ENABLED" }}
+  {{- end }}
+  {{- if eq (include "splunk-otel-collector.operator.env-has" (dict "env" .Values.operator.instrumentation.spec.env "envName" "SPLUNK_PROFILER_MEMORY_ENABLED")) "false" }}
```
dmitryax: Can we just set the following as the default in the values.yaml instead of applying this magic under the hood?
```yaml
operator:
  instrumentation:
    spec:
      env:
        - name: SPLUNK_PROFILER_ENABLED
          value: "true"
        - name: SPLUNK_PROFILER_MEMORY_ENABLED
          value: "true"
```

.chloggen/operator-template-fix.yaml
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, networkExplorer, operator, chart, other)
+component: operator
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix helm template bug where some non-default Operator auto-instrumentation values can cause deployment failures
+# One or more tracking issues related to the change
+issues: [938,943]
```
dmitryax, line 8: I think we should not link both issue and PR for the issue

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
```diff
@@ -138,7 +143,7 @@ Helper for generating environment variables for each instrumentation library.
       {{- $otelResourceAttributes = printf "%s,%s" $env.value $otelResourceAttributes }}
     {{- else }}
       {{- printf "- name: %s" $env.name | nindent 6 -}}
-      {{- printf "  value: %s" $env.value | nindent 6 -}}
+      {{- printf "  value: \"%s\"" $env.value | nindent 6 -}}
```
dmitryax, line 146: Unrelated fix?

dmitryax: Anyway, `quote` should be used instead to avoid double-quoting.

dmitryax:
```yaml
        value: {{ quote $env.value }}
```

helm-charts/splunk-otel-collector/templates/operator/instrumentation.yaml
jvoravong: I've updated the description of this PR to outline the changes more accurately. I'll try to use GH issues more appropriately in the future. For at least this time since this change-set is small I would prefer not to split this PR out more.

.chloggen/operator-template-fix.yaml
jvoravong: Going to only link the PR for now.

dmitryax: In the collector, we link issue and use PR as a fallback if there is no issue for the change

helm-charts/splunk-otel-collector/templates/operator/instrumentation.yaml
dmitryax: It's not about issues specifically. We just need to make sure each PR addresses a particular problem/improvement outlined in the description or the linked issue. It makes the changeset clean for the git history and makes the review process much less confusing.

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
jvoravong: Done

helm-charts/splunk-otel-collector/templates/operator/instrumentation.yaml
jvoravong: Unfortunately I would not recommend this without other changes or trade offs.
- If we instrument applications with profiling enabled and the collector isn't setup to receiver profiling data (splunkObservability.profilingEnabled: true) , the instrumented application will waste resources and log error messages.
- I think it is easier for customers to get lost if we don't apply some magic under the hood.
  - We tried using only validation rules at the start, this was not trivial so we moved onto our current implementation.
  - I wish we could use [yaml-anchors](https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors) in the values file and link the value of splunkObservability.profilingEnabled to these env vars. However since profilingEnabled would be a boolean and these env vars have to have string values it is either not possible or would require some magic.

jvoravong: Unfortunately I would not recommend this without other changes or trade offs.
- If we instrument applications with profiling enabled and the collector isn't setup to receiver profiling data (splunkObservability.profilingEnabled: true) , the instrumented application will waste resources and log error messages.
- I think it is easier for customers to get lost if we don't apply some magic under the hood.
  - We tried using only validation rules at the start, this was not trivial so we moved onto our current implementation.
  - I wish we could use [yaml-anchors](https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors) in the values file and link the value of splunkObservability.profilingEnabled to these env vars. However since profilingEnabled would be a boolean and these env vars have to have string values it is either not possible or would require some magic.

dmitryax: Makes sense. Thanks for clarifying

.chloggen/operator-template-fix.yaml
jvoravong: Done

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
```diff
@@ -103,22 +102,28 @@ Helper to define entries for instrumentation libraries.
...
+    {{- end -}}
   {{- end -}}
 
-  {{- /* Return the list of 'name' keys */ -}}
-  {{- $keyList -}}
+  {{- if $found -}}
+    {{- "true" -}}
+  {{- else -}}
+    {{- "false" -}}
+  {{- end -}}
```
dmitryax, line 126: You don't need this. Just `{{- $found -}}` should work

dmitryax (approved):

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
jvoravong: Done

PR #949, opened 2023-10-05
[chore] e2e tests

atoulme:
**Description:**
add e2e tests that test cluster receiver metrics and an attempt to test trace auto-instrumentation.

The auto-instrumentation test is skipped for now as deployment fails on kind right now, probably tied to the port mapping.


e2e_tests/e2e_test.go
```diff
@@ -0,0 +1,424 @@
...
+					"otlp": map[string]interface{}{
+						"endpoint": fmt.Sprintf("%s:4317", hostEndpoint(t)),
+						"tls": map[string]interface{}{
+							"insecure": true,
+						},
+					},
+				},
+			},
+		},
+	}
```
dmitryax, line 85: Let's put this in a yaml file and read it :)

dmitryax (commented):
I think it's a good start assuming we are going to replace the python tests without losing any coverage instead on maintaining both

Makefile
```diff
@@ -95,3 +95,28 @@ chlog-preview: chlog-validate ## Provide a preview of the generated CHANGELOG.md
 chlog-update: chlog-validate ## Creates an update to CHANGELOG.md for a release entry from content in .chloggen
 	$(CHLOGGEN) update --version "[$(VERSION)] - $$(date +'%Y-%m-%d')" || exit 1; \
 	ci_scripts/chloggen-update.sh || exit 1
+
+CERTMANAGER_VERSION ?= 1.10.0
```
jvoravong, line 99:
```suggestion
CERTMANAGER_VERSION ?= $(shell yq eval ".dependencies[] | select(.name == \"cert-manager\") | .version" helm-charts/splunk-otel-collector/Chart.yaml)
```

atoulme: thanks, added to the latest.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,63 @@
+name: e2e-tests
+
+
+on: pull_request
```
dmitryax, line 4: Should run on main as well

Makefile
dmitryax: looks like it's not applied

e2e_tests/e2e_test.go
```diff
@@ -0,0 +1,299 @@
...
+		return value[(strings.LastIndex(value, "/") + 1):]
+	}
+
+	latestTrace := tracesConsumer.AllTraces()[len(tracesConsumer.AllTraces())-1]
+	actualSpan := latestTrace.ResourceSpans().At(0).ScopeSpans().At(0).Spans().At(0)
+	host, _ := actualSpan.Attributes().Get("http.host")
+	assert.Equal(t, "localhost:3000", host.AsString())
+
+	statusCode, _ := actualSpan.Attributes().Get("http.status_code")
+	assert.Equal(t, "200", statusCode.AsString())
```
dmitryax, line 168: Why checking this additionally? I see the expected files have these attributes

e2e_tests/e2e_test.go
```diff
@@ -0,0 +1,299 @@
...
+
+	expectedSpan := expectedTraces.ResourceSpans().At(0).ScopeSpans().At(0).Spans().At(0)
+	expectedSpan.Attributes().Range(func(k string, _ pcommon.Value) bool {
+		_, ok := actualSpan.Attributes().Get(k)
+		assert.True(t, ok)
+		return true
+	})
```
dmitryax, line 175: Same Q here. Not sure why it's needed

e2e_tests/e2e_test.go
atoulme: I removed this entirely in the next version. Good point.

e2e_tests/e2e_test.go
atoulme: yes same

Makefile
atoulme: I applied it in the next version, leaving this PR untouched. The change is not enough, I had to change the make target a bit to remove the `v` of the version.

e2e_tests/testdata/operator_values.yaml
dmitryax, line 1: Nit: `operator_values.yaml` name is confusing. Should be just `values.yaml` or `test_values.yaml`?

dmitryax (commented):
Great start! Thank you, @atoulme.

The tests are still failing. PTAL

dmitryax (approved):

Makefile
```diff
@@ -95,3 +95,28 @@ chlog-preview: chlog-validate ## Provide a preview of the generated CHANGELOG.md
 chlog-update: chlog-validate ## Creates an update to CHANGELOG.md for a release entry from content in .chloggen
 	$(CHLOGGEN) update --version "[$(VERSION)] - $$(date +'%Y-%m-%d')" || exit 1; \
 	ci_scripts/chloggen-update.sh || exit 1
+
+CERTMANAGER_VERSION ?= $(shell yq eval ".dependencies[] | select(.name == \"cert-manager\") | .version" helm-charts/splunk-otel-collector/Chart.yaml)
+
+.PHONY: cert-manager
+cert-manager: cmctl
+	# Consider using cmctl to install the cert-manager once install command is not experimental
+	kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/${CERTMANAGER_VERSION}/cert-manager.yaml
```
jinja2, line 104: is cert-manager installation for the operator webhook? if so, then can we just [enable the cert-manager dependency](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/values.yaml#L1451) when installing helm chart?

atoulme: No, that didn't work. I copied the cert-manager testing code from the operator project instead.

jvoravong: Jina's suggestion should theoretically be plausible, I wonder if this a limitation of using kind in a GH workflow. This solution is acceptable for now.

The operator project runs tests in parallel using 1 kind cluster with the kuttl test framework which results in each test case getting a unique k8s namespace to run the test case in. Since you can only have 1 cert-manager per k8s cluster, the operator project runs 'make cert-manager' when setting up the kind cluster before any tests are run.

jvoravong (approved):

Makefile
atoulme: It really seemed like a weird ordering issue, as in the operator would be failing to get a cert and just hang there. It is probably because of the nature of the deployment, if you install cert-manager at the same time as the operator.

The good news is that we can patch this to use the cert-manager in the helm chart with a smaller delta.

PR #951, opened 2023-10-06
End to end tests in Golang

atoulme:
This PR is an effort to create a set of tests that can listen to the different data inputs from the chart, and refer to them for testing. The cluster can continuously report data in, and we can use this data to assert based on the deployment.

This test runs against kind with a variety of kubernetes versions.

Issue #896, opened 2023-08-24
Replace NetworkExplorer kernel-collector-podsecuritypolicy since PodSecurityPolicy was removed in K8s v1.25


jvoravong:
## Description

- PodSecurityPolicy was [deprecated](https://kubernetes.io/blog/2021/04/08/kubernetes-1-21-release-announcement/#podsecuritypolicy-deprecation) in Kubernetes v1.21, and removed from Kubernetes in v1.25.
- The Network explorer 
https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/network-explorer/kernel-collector-podsecuritypolicy.yaml uses the no longer available object. 
- Starting in Kubernetes 1.25, you can't use the Network Explorer out of the box

## Steps to Reproduce
Try deploying the network explorer to a Kubernetes +1.25 cluster.

## Expected Result
Should be able to deploy the Network Explorer to current Kubernetes versions.

## Actual Result
Can only deploy to older versions of Kubernetes.

Tested with the chart v0.83.0

### Describe the solution you'd like

Refactor the kernel-collector K8s resources to use proper K8s objects instead of a PodSecurityPolicy.

kumachop2:
I have tested with the chart v0.85.0. Please do advice on timelines.

kumachop2:
@jvoravong I'm still getting error with Pod Sec Policy and here is error. Our K8s Version: v1.25.6

2023-10-20T22:56:52.7227095Z Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "***-splunk-otel-collector-kernel-collector" namespace: "" from "": no matches for kind "PodSecurityPolicy" in version "policy/v1beta1"
2023-10-20T22:56:52.7230520Z ensure CRDs are installed first
2023-10-20T22:56:52.7233232Z helm.go:84: [debug] resource mapping not found for name: "***-splunk-otel-collector-kernel-collector" namespace: "" from "": no matches for kind "PodSecurityPolicy" in version "policy/v1beta1"
2023-10-20T22:56:52.7235682Z ensure CRDs are installed first

kumachop2:
@jvoravong  Please do respond with your findings. Thanks in advance.

jinja2:
[...]

PR #953, opened 2023-10-09
delete pod security policy, enable network explorer functional tests

omrozowicz-splunk:
Delete deprecated PodSecurityPolicy so Network Explorer can work again.

**Issue:** https://github.com/signalfx/splunk-otel-collector-chart/issues/896

**Testing:** Manual tests + enabling network explorer in functional tests

**Documentation:** No documentation needed.

_______
I tested it on microk8s cluster, network explorer pods were initialised  successfully and network metrics were gathered:
image

I don't see all the documented ones, but I don't have any errors so it might be they're not present on my env.

+ I wanted to use functional tests we have now to test if behaviour is the same as before deprecation, but looks like network metrics were never tested there ([as the condition was events >= 0](https://github.com/signalfx/splunk-otel-collector-chart/blob/ed11bd9542864e99946ef5161f6e9c09dbb55c24/test/k8s_metrics_test/test_config_metrics.py#L183C9-L183C9)), I spent some time to try to make network explorer work on minikube to make it a part of testing CI, but that still not sure if it is achievable. Now I see it's going to be disabled completely: https://github.com/signalfx/splunk-otel-collector-chart/pull/951, so no need for me to work any further on it.

### DIFF
--- a/.chloggen/delete-deprecated-podsecuritypolicy.yaml
+++ b/.chloggen/delete-deprecated-podsecuritypolicy.yaml
@@ -1,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, networkExplorer, operator, chart, other)
+component: networkExplorer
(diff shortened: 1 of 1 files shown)


jvoravong (approved):

atoulme:
Indeed. We will bring back network explorer tests in a bit.

PR #955, opened 2023-10-10
Drop time attribute from log records

### LINKED ISSUE #912: Remove "time" from the "fields" sent to splunk_hec endpoints
### Is your feature request related to a problem? Please describe.

I am doing some testing on the fields and formats (event and raw) that the containerd logging pipeline sends to splunk hec. 

When I send the event format to the raw endpoint, i can inspect what OTel is sending:

image

In the payload we can see the "indexed fields" we send. In those fields we should avoid sending massive cardinality for Splunk to index, like "time". 

I believe we spoke about this in the past, and may have done it for docker, but im using containerd rendered logging pipeline from the chart. 

### Describe the solution you'd like

remove timestamps from fields sent by default to splunk to avoid indexing fields with super high cardinality. 

its already being correctly set at the root `time` field. It is also in the raw `event` field if the user ever wants to extract it at search time.

### Describe alternatives you've considered

I can workaround it by trying to drop this field downstream in splunk. It requires configuring a props/transforms to "null out" the indexed field value (i think). 

Will also try dropping via config override on otel agent side. 

### Additional context

we should check all logging pipelines to ensure we are checking the fields we send for auto indexed extractions to avoid indexing/stroage impact, where it makes sense

atoulme:
**Description:**
Fixes #912 

This drops the time attribute from the log records parsed, so it is not sent as an extra field to Splunk.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -414,8 +414,12 @@ receivers:
       # Clean up log record
       - type: move
         id: clean-up-log-record
+        output: clean-up-time-attribute
         from: attributes.log
         to: body
+      - type: remove
+        id: clean-up-time-attribute
+        field: attributes.time
   {{- end }}
 
   {{- if .Values.logsCollection.extraFileLogs }}
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -414,11 +414,9 @@ receivers:
       # Clean up log record
       - type: move
         id: clean-up-log-record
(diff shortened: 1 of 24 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -414,8 +414,12 @@ receivers:
       # Clean up log record
       - type: move
         id: clean-up-log-record
+        output: clean-up-time-attribute
         from: attributes.log
         to: body
+      - type: remove
+        id: clean-up-time-attribute
+        field: attributes.time
```
dmitryax, line 422: No need to add id unless it's being used as a reference in a non-sequential flow

```suggestion
        from: attributes.log
        to: body
      - type: remove
        field: attributes.time
```

dmitryax (approved):

PR #956, opened 2023-10-10
Clean up job name and constant declaration in Makefile

atoulme:
Peeled off these changes from #951

### DIFF
--- a/.github/workflows/functional_test.yaml
+++ b/.github/workflows/functional_test.yaml
@@ -3,7 +3,7 @@ name: Splunk Platform Functional Test
 on: pull_request
 
 jobs:
-  e2e-test:
+  functional-test:
     name: Kubernetes ${{ matrix.kubernetes_version }} ${{ matrix.container_runtime }}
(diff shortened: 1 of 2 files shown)


.github/workflows/functional_test.yaml
```diff
@@ -3,7 +3,7 @@ name: Splunk Platform Functional Test
 on: pull_request
 
 jobs:
-  e2e-test:
+  functional-test:
```
jvoravong, line 6: Nit: I honestly think we should rename the functional tests to e2e where ever possible. It is one of the odd things this project has compared to others.

jvoravong (approved):

.github/workflows/functional_test.yaml
atoulme: I am following the feedback from @rmfitzpatrick that these are not truly e2e tests since they don't involve the o11y product, so renaming to functional tests seems to be closer to what they do, plus they replace existing functional tests. No strong opinion about it.

PR #959, opened 2023-10-11
Release 0.86.0

rmfitzpatrick:


functional_tests/testdata/expected_cluster_receiver.yaml
```diff
@@ -29,7 +29,7 @@ resourceMetrics:
             unit: "1"
         scope:
           name: otelcol/k8sclusterreceiver
-          version: v0.85.0
+          version: v0.86.0
```
jinja2, line 32: should we update this to `latest` instead? The [value is ignored in the test](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/functional_tests/functional_test.go#L203).

jvoravong: I'd prefer keeping declarations like this version locked so over time we know what was tested locally and in CI/CD. open to other input and opinions though.

rmfitzpatrick: I updated to `ignored` so it's less misleading.

jvoravong (approved):

PR #963, opened 2023-10-13
[chore] Remove configuration for disabling opencensus.resourcetype

dmitryax:
The resource attribute was disabled upstream in 0.87.0 and will be removed in the next version.

This change has no effect on the end users, so no need to mention in the changelog

### DIFF
--- a/examples/add-filter-processor/rendered_manifests/configmap-cluster-receiver.yaml
+++ b/examples/add-filter-processor/rendered_manifests/configmap-cluster-receiver.yaml
@@ -71,9 +71,6 @@ data:
         auth_type: serviceAccount
         metadata_exporters:
         - signalfx
-        resource_attributes:
-          opencensus.resourcetype:
-            enabled: false
       prometheus/k8s_cluster_receiver:
         config:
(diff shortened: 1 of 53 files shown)


atoulme:
The tests fail because we need to update the collector to 0.87.0 to remove this attribute.

dmitryax:
> The tests fail because we need to update the collector to 0.87.0 to remove this attribute.

That's because the version of e2e tests that this PR is build on uses OTLP exporter. Rebasing should solve it since now we use signalfx exporter in the tests which removes that attributes

dmitryax:
> The tests fail because we need to update the collector to 0.87.0 to remove this attribute.

You're right. I confused the version. Turning it to draft until 0.87.0 is released

dmitryax:
This should be merged along with the 0.87.0 release or after 0.87.0 image adoption and before the release. Otherwise, there will be some deprecation messages thrown

dmitryax:
@atoulme, rebased. PTAL

dmitryax:

jinja2 (approved):

Issue #973, opened 2023-10-16
docs: Add the minimum helm version required in both the README and Splunk documentation


jek-bao-choo:
For example, when using Helm version 3.5 and earlier, you may encounter an 'unclosed action' error. However, with Helm version 3.6 and later, it functions without issues.

Please make sure to document the minimum Helm version required, such as 'minimum Helm version 3.x.x,' in both the README and Splunk documentation.

PR #978, opened 2023-10-17
Helm info in docs

sbylica-splunk:
**Description:** 

Updated docs to specify minimal version of helm required.

Will fix this issue: https://github.com/signalfx/splunk-otel-collector-chart/issues/973

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ require additional configurations applied to
 
 The following prerequisites are required to use the helm chart:
 
-- [Helm 3](https://helm.sh/docs/intro/install/) (Helm versions lower than 3.6.0 is not supported)
+- [Helm 3](https://helm.sh/docs/intro/install/) (Helm versions lower than 3.6.0 are not supported)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,7 +109,7 @@ require additional configurations applied to
 
 The following prerequisites are required to use the helm chart:
 
-- [Helm 3](https://helm.sh/docs/intro/install/) (Helm 2 is not supported)
+- [Helm 3](https://helm.sh/docs/intro/install/) (Helm versions lower than 3.6.0 are not supported)
```
rmfitzpatrick, line 112: What is the origin of this restriction? https://github.com/signalfx/splunk-otel-collector-chart/issues/973 sounds like a bug to me and not necessarily something we want to enforce as a floor.

sbylica-splunk: I added it because of this issue. So far I can confirm that trying to use helm version below 3.6.0 causes an "unclosed action" error. I will try to investigate what causes this error

jvoravong: Thanks for opening up this PR and bring attention to this. There are a few other things to consider when making a change like this. We opened up https://github.com/signalfx/splunk-otel-collector-chart/pull/979 to help address this. Closing this PR in favor of 979.

sbylica-splunk: Makes sense! Thanks :)

PR #979, opened 2023-10-17
Update Kubernetes and Helm 3 version support in README.md

jvoravong:
Description:
Update Kubernetes and Helm 3 version support in README.md to align with https://docs.splunk.com/observability/en/gdi/opentelemetry/install-k8s.html#supported-kubernetes-distributions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,16 +88,15 @@ This distribution currently supports:
 
 ## Supported Kubernetes distributions
 
-This helm chart is tested and works with default configurations on the following
-Kubernetes distributions:
+The Helm chart works with default configurations of the main Kubernetes distributions. Use actively supported versions:
 
-- [Vanilla (unmodified version) Kubernetes](https://kubernetes.io)
-- [Amazon Elastic Kubernetes Service](https://aws.amazon.com/eks)
+- [Vanilla (unmodified version) Kubernetes](https://endoflife.date/kubernetes)
+- [Amazon Elastic Kubernetes Service](https://endoflife.date/amazon-eks)
   including [with Fargate profiles](docs/advanced-configuration.md#eks-fargate-support)
-- [Azure Kubernetes Service](https://docs.microsoft.com/en-us/azure/aks)
-- [Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine)
+- [Azure Kubernetes Service](https://endoflife.date/azure-kubernetes-service)
+- [Google Kubernetes Engine](https://endoflife.date/google-kubernetes-engine)
   including [GKE Autopilot](docs/advanced-configuration.md#gke-autopilot-support)
-- [Red Hat OpenShift](https://www.redhat.com/en/technologies/cloud-computing/openshift)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,16 +88,15 @@ This distribution currently supports:
 
 ## Supported Kubernetes distributions
 
-This helm chart is tested and works with default configurations on the following
-Kubernetes distributions:
+The Helm chart works with default configurations of the main Kubernetes distributions. Use actively supported versions:
```
rmfitzpatrick, line 91: >Use actively supported versions

Can we add some clarity around this like "support is only provided for <insert provider names> maintained versions per their ~license agreements" and "support for expired versions is best effort and backward incompatible feature adoption cannot be circumvented" (cc @aunshc)

README.md
```diff
@@ -88,16 +88,15 @@ This distribution currently supports:
...
 
-- [Vanilla (unmodified version) Kubernetes](https://kubernetes.io)
-- [Amazon Elastic Kubernetes Service](https://aws.amazon.com/eks)
+- [Vanilla (unmodified version) Kubernetes](https://endoflife.date/kubernetes)
```
rmfitzpatrick, line 93: endoflife.date is a great service, but I'd think linking to primary sources is preferable.

README.md
```diff
@@ -109,7 +108,9 @@ require additional configurations applied to
 
 The following prerequisites are required to use the helm chart:
 
-- [Helm 3](https://helm.sh/docs/intro/install/) (Helm 2 is not supported)
+- [Helm 3](https://helm.sh/docs/intro/install/)
+  - Helm 3 version support may vary depending on the [supported Kubernetes version](#supported-kubernetes-distributions) you are using and the [Helm supported Kubernetes versions skew](https://helm.sh/docs/topics/version_skew/#supported-version-skew).
```
rmfitzpatrick, line 112: Would be helpful to state we use 3.11 in maintaining this project for some frame of reference.

README.md
jvoravong: I recommend keeping the "End of Life" section as it provides valuable links to the primary source at the beginning of each page. Using "End of Life" also standardizes the format for presenting version support across various distributions, making it easier to understand. Parsing the version support documentation for each distribution at the source can be overwhelming. Open to other input.

README.md
jvoravong: I do want this, but I want it to update the value automatically as needed. I'll see if we can make this a badge.

jvoravong: Only added a static reference for now. Will try to add automation later for this in another ticket.

README.md
jvoravong: I'll propose we iterate on updating this in future PRs and in customer docs and move forward with this current version for this PR.

README.md
jvoravong: Done

atoulme (approved):

rmfitzpatrick (approved):

Issue #908, opened 2023-09-04
Error: failed to create containerd task: failed to create shim task:   hcs::CreateComputeSystem otel-collector: The container operating system does   not match the host operating system.: unknown


surya1150:
## Description
Splunk Otel deployment is failing with below error:
Error: failed to create containerd task: failed to create shim task:
  hcs::CreateComputeSystem otel-collector: The container operating system does
  not match the host operating system.: unknown
## Steps to Reproduce

- Install Splunk Otel with helm chart into AKS cluster to run on windows nodes.
- Kubernetes Cluster Version: 1.26.6
- Node image version - AKSWindows-2022-containerd-20348.1906.230809
- 
## Expected Result

- Splunk Otel agent has to be installed correctly

## Actual Result
`Got below error:
Error: failed to create containerd task: failed to create shim task:   hcs::CreateComputeSystem otel-collector: The container operating system does   not match the host operating system.: unknown
`

### Chart version

0.83.0

## Environment
Cloud: "AKS"
k8s version: 1.26.6
OS: Windows

Helm Values.yaml
`clusterName: "aks-poc-7"
splunkPlatform:
  endpoint: ""
  token: "xxxxx"
  index: "xxxxx"
  metricsIndex: "xxxxxx"
  insecureSkipVerify: true
  metricsEnabled: true
  logsEnabled: true
logsEngine: otel
cloudProvider: "azure"
distribution: "aks"  
clusterReceiver:
  enabled: false
image:
  otelcol:
    repository: quay.io/signalfx/splunk-otel-collector-windows                    
tolerations:
  - key: os
    effect: NoSchedule
    value: windows    
isWindows: true 
readinessProbe:
  initialDelaySeconds: 60
livenessProbe:
  initialDelaySeconds: 60  `

rmfitzpatrick:
@surya1150 thanks for your report and please try the `latest-2022` tag (pointing to `0.86.0-2022` at this time).

Feel free to open a support case for issues like this for more immediate assistance.

PR #988, opened 2023-10-23
document 2019/2022 tags

rmfitzpatrick:
Currently we don't document the available platform tags for windows installations, which can lead to installation errors like those reported in https://github.com/signalfx/splunk-otel-collector-chart/issues/908.


docs/advanced-configuration.md
```diff
@@ -108,6 +112,7 @@ isWindows: true
 image:
   otelcol:
     repository: quay.io/signalfx/splunk-otel-collector-windows
+    tag: <appVersion>-2022
```
rmfitzpatrick, line 115: Opted not to document latest example since ideally tag bumps occur w/ upgrades only (and tracking latest can introduce breaking changes w/o notice).

jinja2 (approved):

PR #993, opened 2023-10-30
add combineWith to multiline options

### LINKED ISSUE #756: Expose `combine_with` option in `multilineConfig` section
### Is your feature request related to a problem? Please describe.

When using `multilineConfig` in the helm chart, it defaults to using the `filelog operator` default option for `combinie_with` which is `\n`

This can cause extra whitespace/lines to be rendered in the Splunk back end, where what I actually want for some recombined logs is to set `combine_with` to `""`

image

### Describe the solution you'd like

Expose the option for `combine_with` in the `multiline_config` section. 

### Describe alternatives you've considered

To workaround this, I would have to grab the rendered `router` and `recombine` operators from the config map and move them to the `extraOperators` section in the helm chart instead. 

### Additional context

We had a similar experience in Splunk Connect for Kubernetes, and eventually exposed more of the `concat` plugin features to control these things in the helm chart as well. 

https://github.com/splunk/splunk-connect-for-kubernetes/blob/a65f108f0e41b6ababf811be9217ae10f0f079ba/helm-chart/splunk-connect-for-kubernetes/values.yaml#L266

atoulme:
**Description:**
Fixes #756 

**Documentation:**
Added to advanced docs.

### DIFF
--- a/docs/advanced-configuration.md
+++ b/docs/advanced-configuration.md
@@ -474,6 +474,7 @@ logsCollection:
         containerName:
           value: server
         firstEntryRegex: ^[^\s].*
+        combineWith: ""
 ```
 
 Use https://regex101.com/ to find a golang regex that works for your format and specify it in the config file for the config option `firstEntryRegex`.
--- a/examples/only-logs-otel/rendered_manifests/configmap-agent.yaml
+++ b/examples/only-logs-otel/rendered_manifests/configmap-agent.yaml
@@ -232,10 +232,32 @@ data:
           - expr: (resource["k8s.namespace.name"]) == "default" && (resource["k8s.pod.name"])
               matches "buttercup-app-.*" && (resource["k8s.container.name"]) == "server"
             output: default_buttercup-app-.*_server
+          - expr: (resource["k8s.namespace.name"]) == "k8s-nm2" && (resource["k8s.pod.name"])
+              matches "foo-app-.*" && (resource["k8s.container.name"]) == "client"
+            output: k8s-nm2_foo-app-.*_client
+          - expr: (resource["k8s.namespace.name"]) == "k8s-nm3" && (resource["k8s.pod.name"])
+              matches "foo-app-.*" && (resource["k8s.container.name"]) == "server"
+            output: k8s-nm3_foo-app-.*_server
           type: router
         - combine_field: attributes.log
           combine_with: "\t"
           id: default_buttercup-app-.*_server
+          is_first_entry: (attributes.log) matches "^[^\\s].*"
+          max_log_size: 1048576
+          output: clean-up-log-record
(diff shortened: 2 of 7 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -409,6 +409,7 @@ receivers:
         combine_field: attributes.log
         is_first_entry: '(attributes.log) matches {{ .firstEntryRegex | quote }}'
         max_log_size: {{ $.Values.logsCollection.containers.maxRecombineLogSize }}
+        combine_with: {{ .combineWith | default "\n" | quote }}
```
jinja2, line 412: Do we want to `quote` the value? A values.yaml entry like `combineWith: \t`, will provide the output `combine_with: "\\t"`. Maybe we should clarify a bit if the user needs to set literal or double-quoted scalar in values yaml if we quote in the template.

atoulme: I need an example as test. I wasn't sure.

atoulme: turns out it was a bit more involved than I thought!

examples/only-logs-otel/rendered_manifests/configmap-agent.yaml
```diff
@@ -227,6 +227,42 @@ data:
...
+        - combine_field: attributes.log
+          combine_with: ""
+          id: k8s-nm2_foo-app-.*_client
+          is_first_entry: (attributes.log) matches "^[^\\s].*"
+          max_log_size: 1048576
+          output: clean-up-log-record
+          source_identifier: resource["com.splunk.source"]
+          type: recombine
+        - combine_field: attributes.log
+          combine_with: \n
```
jinja2, line 259: Should this be`"\n"` like the tab char? The[ doc mentions](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/recombine.md#configuration-fields) to give input as double quoted instead of literals.

atoulme: Maybe I'll just omit combine_with for the default case since it's assumed to be "\n".

docs/advanced-configuration.md
```diff
@@ -474,6 +474,7 @@ logsCollection:
         containerName:
           value: server
         firstEntryRegex: ^[^\s].*
+        combineWith: ""
```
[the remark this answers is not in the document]
atoulme: I think you're saying you want a comment? Sure.

jinja2 (approved):

PR #1026, opened 2023-11-20
Deprecate the networkExplorer option

dmitryax:
Add instruction on how to replace networkExplorer with the upstream OpenTelemetry eBPF Helm Chart.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,48 @@
 # Upgrade guidelines
 
+## 0.87.0 to 0.88.0
+
+The `networkExplorer` option is deprecated now. Please use the upstream OpenTelemetry eBPF Helm chart to collect
+the network metrics by following the next steps:
+
+1. Make sure the Splunk OpenTelemetry Collector helm chart is installed with the gateway enabled:
+
+```yaml
+gateway:
+  enabled: true
+```
+
+2. Disable the network explorer:
+
+```yaml
+networkExplorer:
+  enabled: false
+```
+
+3. Grab name of the Splunk OpenTelemetry Collector gateway service:
+
+```bash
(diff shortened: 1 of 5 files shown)


atoulme (approved):

docs/advanced-configuration.md
```diff
@@ -780,3 +780,35 @@ agent:
   * Cluster receiver is a 1-replica deployment of Open-temlemetry collector.
   * As any available node can be selected by the Kubernetes control plane to run the cluster receiver pod (unless we explicitly specify the `clusterReceiver.nodeSelector` to pin the pod to a specific node), `hostPath` or `local` volume mounts wouldn't work for such envrionments.
   * Data Persistence is currently not applicable to the k8s cluster metrics and k8s events.
+
+### Using OpenTelemetry eBPF helm chart with Splunk OpenTelemetry Collector for Kubernetes
```
aunshc, line 784: Question: Is there a set of generic instructions that customers will be able to leverage if they are using the upstream helm chart instead?

UPGRADING.md
```diff
@@ -1,5 +1,12 @@
 # Upgrade guidelines
 
+## 0.87.0 to 0.88.0
+
+The `networkExplorer` option is deprecated now. Please use the upstream OpenTelemetry eBPF Helm chart to collect
```
jinja2, line 5: Should we add more instructions about how to switch? Is it recommended for users to first disable the nw explorer from Splunk distro release and then install the new release? I assume we might not want both the running simultaneously.

dmitryax: Good point. updated 👍

docs/advanced-configuration.md
dmitryax: discussed offline

PR #1051, opened 2023-11-29
[chore] add EKS to test upgrades

atoulme:
This PR adds eks upgrade testing. It is using a helm command to install the latest main for a PR or the previous commit when running on main, and then we run the tests which try to install then upgrade helm with the latest.

### DIFF
--- a/.github/workflows/functional_test_v2.yaml
+++ b/.github/workflows/functional_test_v2.yaml
@@ -102,3 +102,71 @@ jobs:
         run: |
           cd functional_tests
           go test -v
+  eks-upgrade-test:
+    concurrency:
+      group: eks-access
+    env:
+      KUBECONFIG: /tmp/kube-config-splunk-otel-collector-chart-functional-testing-eks
+      SKIP_TESTS: "true" # we need to skip functional tests as we have not set probes to listen to the traffic.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Checkout main if in a PR
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@v4
+        with:
+          path: base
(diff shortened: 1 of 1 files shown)


.github/workflows/functional_test_v2.yaml
```diff
@@ -102,3 +102,71 @@ jobs:
...
+          aws eks update-kubeconfig --name rotel-eks --region us-west-1
+      - name: Update dependencies
+        run: |
+          cd base && make dep-update
+      - name: Deploy cert-manager
+        run: |
+          cd base && make cert-manager
+      - name: Deploy previous version of the chart
+        run: |
+          cd base && helm install sock helm-charts/splunk-otel-collector --set clusterName=dev-operator --set splunkObservability.realm=us0 --set splunkObservability.accessToken=xxxxx
```
jvoravong, line 160:
```suggestion
          cd base && helm install splunk-otel-collector helm-charts/splunk-otel-collector --set clusterName=dev-operator --set splunkObservability.realm=us0 --set splunkObservability.accessToken=xxxxx
```
We should remove any sck or sock references and avoid adding more.

jvoravong:
It would be useful to add a make actions like these. Can be added in a future PR, I'll likely try to add these in a future PR.
- prepare-e2e: Make sure kind and needed test resources are setup
- functional-test/functional-test-{subset}: Run the functional tests or subset in such a way that github workflows or local devs can run can use this action and run the tests.

jvoravong (approved):

.github/workflows/functional_test_v2.yaml
atoulme: that's just the name of the deployment, also used in the functional tests. We would need to change both references.

PR #1053, opened 2023-12-01
Metrics index from annotations

sbylica-splunk:
**Description:** 
Added support for annotations in metrics index

**Testing:** 
Unit test

**Documentation:** 
Added short explanation to advanced-configuration.md


rmfitzpatrick:
@sbylica-splunk why is a new k8sattributes processor being added for all components instead of enhancing the existing agent's? I don't believe these attrs are necessary for o11y cloud cases so limiting to platform w/ metrics configurations seems preferable as well.

atoulme:
Please rebase and rerun make render

sbylica-splunk:
> Please rebase and rerun make render

Done

rmfitzpatrick:
> As for you second question, do you mean something like that?

Yes, though I think we'll also want to ensure that the new attributes aren't exported as o11y metric dimensions via the signalfx exporter. I'm not sure if there's a way to do that other than duplicating pipelines or using a connector w/ an additional transform processor. Maybe @dmitryax has suggestions since I think he's better versed on the use case than I am.

examples/add-kafkametrics-receiver/rendered_manifests/configmap-agent.yaml
```diff
@@ -95,6 +95,41 @@ data:
         - sources:
           - from: resource_attribute
             name: host.name
+      k8sattributes/metrics:
+        extract:
+          annotations:
+          - from: pod
+            key: splunk.com/sourcetype
```
dmitryax, line 102:
```suggestion
            key: splunk.com/sourcetype
            tag_name: com.splunk.sourcetype
```

examples/add-kafkametrics-receiver/rendered_manifests/configmap-agent.yaml
```diff
@@ -95,6 +95,41 @@ data:
...
+          - from: namespace
+            key: splunk.com/metricsIndex
+            tag_name: com.splunk.index
+          - from: pod
+            key: splunk.com/metricsIndex
+            tag_name: com.splunk.index
+          metadata:
+          - k8s.node.name
+          - k8s.namespace.name
```
dmitryax, line 111: Why are these two needed? All the k8s metrics should go with these attributes already.

examples/add-kafkametrics-receiver/rendered_manifests/configmap-agent.yaml
```diff
@@ -95,6 +95,41 @@ data:
...
+        pod_association:
+        - sources:
+          - from: resource_attribute
+            name: k8s.node.name
```
dmitryax, line 115: this pod association is won't work, not helpful

examples/add-kafkametrics-receiver/rendered_manifests/configmap-agent.yaml
```diff
@@ -95,6 +95,41 @@ data:
...
+          - from: resource_attribute
+            name: k8s.pod.name
+        - sources:
+          - from: resource_attribute
+            name: ip
+        - sources:
+          - from: connection
+        - sources:
+          - from: resource_attribute
+            name: host.name
```
dmitryax, line 132: same for this one

dmitryax:
@sbylica-splunk please make it conditional. At least if `splunk-otel-collector.platformMetricsEnabled` as @rmfitzpatrick suggested. Also, keep in mind that it introduces a significant memory utilization overhead

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -484,6 +484,8 @@ processors:
     filter:
       node_from_env_var: K8S_NODE_NAME
 
+  {{- include "splunk-otel-collector.k8sAttributesProcessorMetrics" . | nindent 2 }}
```
jinja2, line 487: I think this should be limited to pulling pod data local to the agent node only. So adding the config to this processor
```
    filter:
      node_from_env_var: K8S_NODE_NAME
```

examples/add-kafkametrics-receiver/rendered_manifests/configmap-agent.yaml
jinja2: I think this one does work for the case when pod's IP is set as the `host.name`

jinja2: Since we also use this association in our default k8sattribute processor too, want to confirm that it can [work with pod ip](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/pod_association.go#L50)

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -138,6 +138,47 @@ k8sattributes:
     {{- end }}
 {{- end }}
 
+{{/*
+Common config for K8s attributes processor adding k8s metadata to metrics resource attributes.
+*/}}
+{{- define "splunk-otel-collector.k8sAttributesProcessorMetrics" -}}
```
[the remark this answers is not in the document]
sbylica-splunk: Good idea, changed it.

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -212,7 +212,7 @@ service:
     # k8s metrics pipeline
     metrics:
       receivers: [k8s_cluster]
-      processors: 
+      processors:
```
atoulme, line 215: Can you revert whitespace changes to make it easier to see the diff?

atoulme:
```suggestion
      processors: 
```

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -231,10 +231,10 @@ service:
...
     metrics/eks:
       receivers: [receiver_creator]
-      processors: 
-        - memory_limiter 
-        - batch 
-        - resource 
+      processors:
+        - memory_limiter
+        - batch
+        - resource
```
atoulme, line 237:
```suggestion
      processors: 
        - memory_limiter 
        - batch 
        - resource 
```

examples/add-kafkametrics-receiver/rendered_manifests/configmap-agent.yaml
sbylica-splunk: I'll leave it as is then

sbylica-splunk:
Ok, so is there anything else to do? I'll squash current commits after this pr is good to go :)

examples/add-kafkametrics-receiver/rendered_manifests/configmap-agent.yaml
dmitryax: > I think this one does work for the case when pod's IP is set as the host.name

I don't think if we still have any receiver setting pod IP to this attribute. Prometheus receiver used to set this resource attribute. It's been setting `net.host.name` for a while now which is now deprecated in sem conv https://github.com/open-telemetry/semantic-conventions/blob/bef6b4b39f8fc50c266742f56b23c7024a44bb68/docs/attributes-registry/network.md#deprecated-network-attributes. We will need to update the default config of k8sattributes processor after Prometheus receiver is updated

examples/add-kafkametrics-receiver/rendered_manifests/configmap-agent.yaml
dmitryax: I used to be accidentally required due to a bug which was fixed recently https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/28838. Make sure you use +0.90.0 image

sbylica-splunk: Yeah, just checked it out with a new image and it seems to be working fine. Thanks :)

examples/add-kafkametrics-receiver/rendered_manifests/configmap-agent.yaml
sbylica-splunk: Ok, so I removed host.name but kept k8s.node.name

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -484,6 +484,10 @@ processors:
     filter:
       node_from_env_var: K8S_NODE_NAME
 
+  {{- include "splunk-otel-collector.k8sAttributesSplunkPlatformMetrics" . | nindent 2 }}
```
jinja2, line 487: Would you mind adding this optionally only when Splunk Platform metrics are enabled? I see you are adding the processor to the pipeline conditionally already, setting the same check here would keep the config more streamlined for users not using Splunk platform metrics.

.chloggen/metric-index-from-annottaions.yaml
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, networkExplorer, operator, chart, other)
+component: other
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Added support for rerouting metrics index with pod/namespace annotations
+# One or more tracking issues related to the change
+issues: []
```
jinja2, line 8:
```suggestion
issues: [1053]
```
failing changelog validation

jinja2:
@sbylica-splunk please fix trailing whitespace issues as seen in the [failing pre-commit job](https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/7207840021/job/19635502680?pr=1053)

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
sbylica-splunk: Makes sense! Added it

atoulme (approved):

dmitryax (approved):

PR #1056, opened 2023-12-01
Make the clusterName value optional in EKS and GKE

crobert-1:

Now that the resource detection processor supports detecting the `k8s.cluster.name` in GKE and EKS environments, the `clusterName` value can be optional in these environments.

A simplified resource detection processor has been added to more pipelines in this change as well, in every place the `clusterName` was manually being added.

Tested in GKE and kOps.  GKE sets cluster name properly without it specified, and overrides the detected `k8s.cluster.name` when `clusterName` is set. kOps fails without `clusterName` with the following message:
```
crobert$ ~/dev/splunk-otel-collector-chart $ helm upgrade ...
splunk-otel-collector:
- (root): Must validate "then" as "if" was valid
- clusterName: String length must be greater than or equal to 1
```
The `- (root)` lines are a bit confusing, but that's simply saying that the check failed, AFAICT.

Updated README and relevant comments in code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ The following parameter is required or optional depending on the Kubernetes serv
 
 - `clusterName`: arbitrary value that identifies your Kubernetes cluster. The value will be associated with every trace, metric and log as "k8s.cluster.name" attribute.
   * Optional: EKS and GKE. If `clusterName` is specified it will overwrite detected value.
-  * Required: All other services
+  * Required: For all other distributions.
 
 Run the following commands, replacing the parameters above with their appropriate values.
 
--- a/README.md
+++ b/README.md
@@ -152,9 +152,11 @@ For Splunk Observability Cloud the following parameters are required:
 - `splunkObservability.accessToken`: Your Splunk Observability org access
   token.
 
-The following parameter is required for any of the destinations:
+The following parameter is required or optional for any of the destinations, depending on the Kubernetes service:
 
 - `clusterName`: arbitrary value that identifies your Kubernetes cluster. The value will be associated with every trace, metric and log as "k8s.cluster.name" attribute.
+  * Optional: EKS and GKE. Detected value will be overwritten by this parameter if specified.
+  * Required: All other services
 
 Run the following commands, replacing the parameters above with their appropriate values.
 
--- a/README.md
+++ b/README.md
@@ -152,10 +152,10 @@ For Splunk Observability Cloud the following parameters are required:
 - `splunkObservability.accessToken`: Your Splunk Observability org access
   token.
 
-The following parameter is required or optional for any of the destinations, depending on the Kubernetes service:
+The following parameter is required or optional depending on the Kubernetes service:
 
 - `clusterName`: arbitrary value that identifies your Kubernetes cluster. The value will be associated with every trace, metric and log as "k8s.cluster.name" attribute.
-  * Optional: EKS and GKE. Detected value will be overwritten by this parameter if specified.
+  * Optional: EKS and GKE. If `clusterName` is specified it will overwrite detected value.
   * Required: All other services
 
 Run the following commands, replacing the parameters above with their appropriate values.
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ For Splunk Observability Cloud the following parameters are required:
 The following parameter is required or optional depending on the Kubernetes service:
 
 - `clusterName`: arbitrary value that identifies your Kubernetes cluster. The value will be associated with every trace, metric and log as "k8s.cluster.name" attribute.
-  * Optional: EKS and GKE. If `clusterName` is specified it will overwrite detected value.
+  * Optional: If `distribution` is set to EKS, EKS/fargate, GKE, and GKE/autopilot. If `clusterName` is specified it will overwrite detected value.
   * Required: For all other distributions.
 
 Run the following commands, replacing the parameters above with their appropriate values.
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -106,10 +106,8 @@ Common config for adding k8s.cluster.name using the resourcedetection processor
 {{- define "splunk-otel-collector.resourceDetectionProcessorKubernetesClusterName" -}}
 resourcedetection/k8s_cluster_name:
   detectors:
-    # Note: Kubernetes distro detectors need to come first so they set the proper cloud.platform
-    # before it gets set later by the cloud provider detector.
     - env
-    {{- if or (hasPrefix "gke" (include "splunk-otel-collector.distribution" .)) (eq (include "splunk-otel-collector.cloudProvider" .) "gcp") }}
+    {{- if hasPrefix "gke" (include "splunk-otel-collector.distribution" .) }}
     - gcp
     {{- else if hasPrefix "eks" (include "splunk-otel-collector.distribution" .) }}
     - eks
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -80,8 +80,58 @@ resourcedetection:
     # The `system` detector goes last so it can't preclude cloud detectors from setting host/os info.
     # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor#ordering
     - system
+  {{- if hasPrefix "gke" (include "splunk-otel-collector.distribution" .) }}
+  gcp:
+    resource_attributes:
+      k8s.cluster.name:
+        enabled: true
+  {{- else if hasPrefix "eks" (include "splunk-otel-collector.distribution" .) }}
+  eks:
+    resource_attributes:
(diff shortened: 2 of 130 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
```diff
@@ -150,7 +152,12 @@ service:
         - memory_limiter
         - k8sattributes
         - batch
+        {{- if or (hasPrefix "gke" (include "splunk-otel-collector.distribution" .)) (hasPrefix "eks" (include "splunk-otel-collector.distribution" .)) }}
```
crobert-1, line 155: I was thinking it might be good to create a central boolean variable called `isClusterNameOptional` (or something like that) that's simply set to be whether the distribution started with `eks` or `gke`. This would be a lot cleaner in lots re-used places, and would allow easier extension in the future.

This could be done by adding an undocumented variable in `values.schema.json` and set it to `true` when the distribution doesn't require the `clusterName` setting. I thought this might open us up to some bad behavior though if a user found the option and set it to `true`.

Would another place work to add some kind of internal variable that could be referenced in templates? Or some maybe the `values.schema.json` option is acceptable? I'm open to suggestions, I'd really prefer to reference a single bool rather than many duplicate if-checks.

jvoravong:
To get the EKS functional-tests up and running, you will have to close this PR and open up a different PR in a specific way due to recent project test updates. The EKS functional tests need to run with credentials that are only available in this repo.
* Create your feature branch in the https://github.com/signalfx/splunk-otel-collector-chart repo and not your personal fork.
* Create the PR with the feature branch in the https://github.com/signalfx/splunk-otel-collector-chart repo against the main branch.

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
dmitryax: sounds reasonable to move this in a helper but not sure how `values.schema.json` can help. We usualy keep all the templating helpers in _common.tpl

README.md
```diff
@@ -152,9 +152,11 @@ For Splunk Observability Cloud the following parameters are required:
 - `splunkObservability.accessToken`: Your Splunk Observability org access
   token.
 
-The following parameter is required for any of the destinations:
+The following parameter is required or optional depending on the Kubernetes service:
 
 - `clusterName`: arbitrary value that identifies your Kubernetes cluster. The value will be associated with every trace, metric and log as "k8s.cluster.name" attribute.
+  * Optional: EKS and GKE. If `clusterName` is specified it will overwrite detected value.
+  * Required: All other services
```
jvoravong, line 159:
```suggestion
  * Required: All other distributions.
```

README.md
```diff
@@ -152,9 +152,11 @@ For Splunk Observability Cloud the following parameters are required:
 - `splunkObservability.accessToken`: Your Splunk Observability org access
   token.
 
-The following parameter is required for any of the destinations:
+The following parameter is required or optional depending on the Kubernetes service:
 
 - `clusterName`: arbitrary value that identifies your Kubernetes cluster. The value will be associated with every trace, metric and log as "k8s.cluster.name" attribute.
+  * Optional: EKS and GKE. If `clusterName` is specified it will overwrite detected value.
```
jvoravong, line 158:
```suggestion
  * Optional: If `distribution` is set to EKS, EKS/fargate, GKE, and GKE/autopilot. If `clusterName` is specified it will overwrite detected value.
```
Wasn't sure if EKS/fargate and GKE/autopilot were supported at a first glance.

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -1548,6 +1547,45 @@
         }
       ]
     },
+    {
+      "if": {
```
rmfitzpatrick, line 1551: What's the benefit of enhancing the validation to the json schema as opposed to migrating it to a helm helper in total? Not sure from https://github.com/signalfx/splunk-otel-collector-chart/pull/1056#discussion_r1412656557 if you're moving to that approach but imo the cumbersome nature of conditionals in this dsl is enough to drop it alone.

jvoravong: My Notes:
- The validation rules and descriptions in [values.schema.json](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.schema.json) are crucial for guiding users on the chart's functionality. While JSON schema validation has its limits in handling complex business logic, it's vital for maintainability as the values.schema.json is used in many ways to present to the user how the chart is utilized.
- Storing as much information and validation in the values.schema.json file is preferable. It simplifies the presentation of usage instructions to users, a method commonly adopted by many Kubernetes providers that utilize chart add-ons. This approach contrasts with using the values.yaml or template logic.
  - For example, a new distribution/use-case will utilize the values.schema.json file as the primary source to instruct users on chart usage.
- Implementing complex business logic in the values.schema.json file can become unsustainable or overly complicated. In such scenarios, I recommend shifting the validation logic to the Helm Chart templates.
  - In this PR, the current changes and validation logic seem suitable to me as they can be maintained within the values.schema.json file easily. However, should the conditionals expand significantly or if we need to accommodate significantly more variables in the conditional logic for our supported use cases, transferring this validation logic to the Helm templates would be advisable.
  - Transferring the validation logic to Helm templates makes describing this logic to users harder.

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
jvoravong: My Notes:
- Some Kubernetes environments will present the whole values.schema.json file as the information on how to use the chart.
- I think defining a validation helper is acceptable when the validation logic is complicated and/or can be reused in multiple places. We do something similar with the operator in these locations.
  - https://github.com/signalfx/splunk-otel-collector-chart/blob/79312ca68dc3dfda114ef45c4bc10a1a22bfdd99/helm-charts/splunk-otel-collector/templates/operator/instrumentation.yaml#L2
  - https://github.com/signalfx/splunk-otel-collector-chart/blob/79312ca68dc3dfda114ef45c4bc10a1a22bfdd99/helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl#L6

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -1548,6 +1547,45 @@
...
+        "properties": {
+          "distribution": { "pattern": "^[^g].[^k]*[^e].*" }
+        },
+        "required": ["distribution"]
+      },
+      "then": {
+        "if": {
+          "properties": {
+            "distribution": { "pattern": "^[^e].[^k]*[^s].*" }
```
jvoravong, line 1560: Is it possible to simplify validation logic for new folks? I was thinking of combining the 2 value checks into 1 with a value like this.
`"distribution": { "pattern": "^(gke|eks).*" }`

helm-charts/splunk-otel-collector/values.schema.json
crobert-1: _Marking as resolved based on Josh's comments, seems like this is a good place for now._

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -80,8 +80,56 @@ resourcedetection:
...
+  {{- end }}
+{{- end }}
+
+{{/*
+Common config for adding k8s.cluster.name using the resourcedetection processor
+*/}}
+{{- define "splunk-otel-collector.resourceDetectionProcessorKubernetesClusterName" -}}
+resourcedetection/k8s_cluster_name:
+  detectors:
+    - env
```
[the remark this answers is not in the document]
crobert-1: No, I'll remove it.

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -80,8 +80,56 @@ resourcedetection:
...
+    {{- if hasPrefix "gke" (include "splunk-otel-collector.distribution" .) }}
+    - gcp
+    {{- else if hasPrefix "eks" (include "splunk-otel-collector.distribution" .) }}
+    - eks
+    {{- end }}
+  {{- if hasPrefix "gke" (include "splunk-otel-collector.distribution" .) }}
+  gcp:
+    resource_attributes:
+      k8s.cluster.name:
+        enabled: true
```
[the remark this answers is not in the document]
crobert-1: Good call, I've disabled all other resource attributes by default.

helm-charts/splunk-otel-collector/values.schema.json
crobert-1: Good call, updated.

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
crobert-1: I've added a bool definition to the [helpers template file.](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/_helpers.tpl)

crobert-1:
Closing in favor of #1067

PR #1067, opened 2023-12-07
Make clusterName optional on EKS and GKE

crobert-1:
**Description:**
Now that the resource detection processor supports detecting the k8s.cluster.name in GKE and EKS environments, the clusterName value can be optional in these environments.

A simplified resource detection processor has been added to more pipelines in this change as well, in every place the clusterName was manually being added.

**Testing:**
Tested in GKE and kOps. GKE sets cluster name properly without it specified, and overrides the detected k8s.cluster.name when clusterName is set. kOps fails without clusterName with the following message:
```
crobert$ ~/dev/splunk-otel-collector-chart $ helm upgrade ...
splunk-otel-collector:
- (root): Must validate "then" as "if" was valid
- clusterName: String length must be greater than or equal to 1
```
The - (root) lines are a bit confusing, but that's simply saying that the check failed, AFAICT.

Also, removed `clusterName` option from functional tests and changed expected `k8s.cluster.name` value in telemetry to be `rotel-eks` instead of the manually set value.

**Documentation:**
Updated README and relevant comments in code.


crobert-1:
Note, this is a second attempt of #1056 to enable testing to work properly. I have not yet implemented changes requested in that review, other than README wording.

crobert-1:
I believe I've addressed all comments on the original PR now, so I'm marking as ready for review.

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -457,3 +457,11 @@ Build the securityContext for Linux and Windows
 {{- end }}
 {{- toYaml .securityContext }}
 {{- end -}}
+
+{{/*
+Whether the clusterName configuration option is optional
+*/}}
+{{- define "splunk-otel-collector.clusterNameOptional" -}}
+{{- if or (hasPrefix "gke" (include "splunk-otel-collector.distribution" .)) (hasPrefix "eks" (include "splunk-otel-collector.distribution" .)) }}
+{{- end -}}
```
jinja2, line 466:
```suggestion
{{- or (hasPrefix "gke" (include "splunk-otel-collector.distribution" .)) (hasPrefix "eks" (include "splunk-otel-collector.distribution" .)) }}
```
I think with the if statement this template will return an empty string.

README.md
```diff
@@ -152,9 +152,11 @@ For Splunk Observability Cloud the following parameters are required:
 - `splunkObservability.accessToken`: Your Splunk Observability org access
   token.
 
-The following parameter is required for any of the destinations:
+The following parameter is required or optional depending on the Kubernetes service:
```
dmitryax, line 155:
```suggestion
The following parameter is required or optional depending on the Kubernetes distribution:
```

examples/add-receiver-creator/rendered_manifests/configmap-cluster-receiver.yaml
```diff
@@ -66,6 +66,10 @@ data:
         - system
         override: true
         timeout: 10s
+      resourcedetection/k8s_cluster_name:
+        detectors: null
+        override: true
+        timeout: 10s
```
dmitryax, line 72: Can we avoid introducing this processor if it's not need?

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -80,8 +80,91 @@ resourcedetection:
...
+      k8s.cluster.name:
+        enabled: true
+  {{- end }}
   override: true
+  {{/* Testing determined that the EKS cluster name detection required a slightly longer default timeout. */}}
+  {{- if hasPrefix "eks" (include "splunk-otel-collector.distribution" .) }}
+  timeout: 15s
+  {{- else -}}
   timeout: 10s
+  {{- end }}
```
dmitryax, line 100: I think that's ok to increase timeout from 10s to 15s unconditionally. I won't affect the start time anyway

.chloggen/optionalclustername.yaml
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, networkExplorer, operator, chart, other)
+component: chart
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Make clusterName optional in EKS and GKE
+# One or more tracking issues related to the change
+issues: [1056]
```
jvoravong, line 8:
```suggestion
issues: [1056,1067]
```

jvoravong (approved):
One nit. Mostly LGTM. Approving. Please address other review comments before merging.

examples/add-receiver-creator/rendered_manifests/configmap-cluster-receiver.yaml
crobert-1: There are checks to only include the processor in the pipeline if it is detecting the GKE/EKS cluster name, but I agree, there's no point in defining a processor if it's not being used. This processor will now only be defined if it's going to be used (if cluster name is optional).

jinja2:
@crobert-1 I think it is worth updating the eks functional_tests to test with the resourcedetector added cluster-name. The test currently [sets the cluster name in values](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/functional_tests/testdata/test_values.yaml.tmpl#L2) which if removed, should default to the resourcedetector provided attr. This would need a change to the testdata, since the name of the eks cluster being tested on is `rotel-eks` and not the provided `dev-operator`.

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -80,8 +80,81 @@ resourcedetection:
     # The `system` detector goes last so it can't preclude cloud detectors from setting host/os info.
     # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor#ordering
     - system
+  {{- if hasPrefix "gke" (include "splunk-otel-collector.distribution" .) }}
```
jinja2, line 83:
```suggestion
  {{- if and (hasPrefix "gke" (include "splunk-otel-collector.distribution" .)) .Values.clusterName  }}
```
How about not enabling this attr if clusterName variable is set? The ordering of the processors means that even in eks/gke distros, if clusterName is set, the attr value will be equal to the one set by `resource` processor. So we might as well not detect when clusterName var exists. This would potentially save the extra 5-10s delay with the EKS detector on startup.

crobert-1: Good idea, I'll update 👍

crobert-1: I ended up just creating a new variable called `autoDetectClusterName`, which allows us to only add the cluster name logic if `clusterName` isn't specified, and the distribution is `eks*` or `gke*`. I tested in `gke` and `kops` with and without `clusterName`, and found no cluster name related config additions (outside of what was already there before this PR) when `clusterName` was specified.

crobert-1:
I've updated functional tests to account for the EKS optional cluster name. These tests are currently being skipped as the functionality is not complete, but the logic is in place for when they're running. I've moved the `kind` cluster test data to a sub-directory, along with the `eks` cluster data to its own sub-directory. The EKS data is not yet being used.

dmitryax (approved):

atoulme (approved):

jinja2 (approved):

jinja2:
hmm, weird that the EKS tests are being skipped after the rebase since the source isn't a fork

crobert-1:
Regarding the [failing upgrade test](https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/7389980540/job/20104114757?pr=1067): It's failing because the cluster name is no longer passed in to the install command. Since the test is running a previous version of the chart without this PR's changes, the `clusterName` is still required, even in EKS environments. I think it's okay to merge this change as once this PR is no longer the `latest` commit in `main`, it should pass again.

atoulme:
Please go ahead and merge.

PR #1089, opened 2023-12-18
[chore] skip eks tests if PR repo is different

jinja2:

Skip EKS tests which require access to secrets when PR opened from a different repo/fork.

### DIFF
--- a/.github/workflows/functional_test_v2.yaml
+++ b/.github/workflows/functional_test_v2.yaml
@@ -61,7 +61,7 @@ jobs:
           go test -v
   eks-test:
     needs: kubernetes-test
-    if: github.repository == 'signalfx/splunk-otel-collector-chart'
+    if: github.event.pull_request.head.repo.name == github.repository
     concurrency:
       group: eks-access
     env:
@@ -106,7 +106,7 @@ jobs:
           go test -v
   eks-upgrade-test:
     needs: kubernetes-test
-    if: github.repository == 'signalfx/splunk-otel-collector-chart'
+    if: github.event.pull_request.head.repo.name == github.repository
     concurrency:
       group: eks-access
     env:
--- a/.github/workflows/functional_test_v2.yaml
+++ b/.github/workflows/functional_test_v2.yaml
@@ -61,7 +61,7 @@ jobs:
           go test -v
 
   eks-test:
-    name: Test in EKS cluster - Credentials required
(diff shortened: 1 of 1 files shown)


.github/workflows/functional_test_v2.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
           go test -v
   eks-test:
     needs: kubernetes-test
-    if: github.repository == 'signalfx/splunk-otel-collector-chart'
+    if: github.event.pull_request.head.repo.name == github.repository
```
crobert-1, line 64: Would there be some way to log a message to make it clear to the user that the test was skipped for this reason? It might cause some confusion to users who think everything is working perfectly when CI passes, or they might be unsure why a test isn't running when they expect it to.

crobert-1 (approved):

atoulme (approved):

.github/workflows/functional_test_v2.yaml
atoulme: +1

jinja2: There doesn't seem to be a way in github actions to log why a test was skipped, can't even see an option to have github log the condition which made the test skip.

I guess we can add a step in the `eks-test` jobs but then the jobs will not show as skipped and become more confusing in my opinion. The other option would be have a different job which evaluates this condition and logs, but again the UX will not be good, and requires the user to see the output of this other job to make sense of why a following job was skipped. Any suggestions?

jvoravong (approved):

.github/workflows/functional_test_v2.yaml
jvoravong: There are a few (not great) options here. Here are my suggestions.
- Rename EKS Test Steps:
  - Option A: "eks-test (specific to signalfx/splunk-otel-collector-chart repo)"
  - Option B: "eks-test (credentials dependent)"
- Separate Workflow for "Smoke" Functional Tests:
  - Propose creating a distinct workflow for tests that require real Kubernetes credentials. This could be termed as "Smoke Tests".
  - This workflow would be triggered only if the necessary credentials are available or if the repository is "signalfx/splunk-otel-collector-chart".

jinja2: I have added more descriptive names to the EKS jobs, ptal!

atoulme (approved):

PR #1099, opened 2024-01-02
[chore] do not update PRs if already open with no diff on helm chart

atoulme:
We have a PR [open](https://github.com/signalfx/splunk-otel-collector-chart/pull/1094) for the last few weeks that would be recreated every day based off the cron job updating the PR. Because the workflow creates the PR and then updates it, it forgets the previous state and forces update the PR.

* We get a notification a day of the PR refresh
* We lose work on the PR, such as fixing tests.

To fix this, we want to avoid recreating the PR if no changes are detected after performing the update against the helm-chart folder.

### DIFF
--- a/.github/workflows/update_docker_images.yaml
+++ b/.github/workflows/update_docker_images.yaml
@@ -69,10 +69,9 @@ jobs:
           branch: "update-${{ matrix.name }}" # Same branch name for all PRs
           base: main
           delete-branch: true
-          modify-outputs: false
 
       - name: Apply Version Update and Generate Changelog
-        if: ${{ steps.check_for_update.outputs.NEED_UPDATE == 1 && steps.check_if_pr_open.outputs.PR_NEEDED == 1 }}
+        if: ${{ steps.check_for_update.outputs.NEED_UPDATE == 1 && steps.check_if_pr_open.outputs.PR_NEEDED == 1 && steps.open_pr.outputs.pull-request-number != '' }}
         run: |
           # Apply the version update, update the rendered examples with the version update, and create a changelog entry
           # We run `make update-docker-image` again here because the open_pr peter-evans/create-pull-request step before clears out the update changes locally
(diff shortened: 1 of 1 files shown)


.github/workflows/update_docker_images.yaml
```diff
@@ -50,9 +52,15 @@ jobs:
 
           make update-docker-image FILE_PATH=${{ matrix.yaml_file_path }} QUERY_STRING='${{ matrix.yaml_value_path }}' DEBUG_MODE=$DEBUG_MODE
 
+      - name: Check if PR is already open
+        id: check_if_pr_open
+        run: |
+          git diff --exit-code master..update-${{ matrix.name }} -- helm-charts
```
jinja2, line 58:
```suggestion
          git diff --exit-code main..update-${{ matrix.name }} -- helm-charts
```

.github/workflows/update_docker_images.yaml
```diff
@@ -50,9 +52,15 @@ jobs:
 
           make update-docker-image FILE_PATH=${{ matrix.yaml_file_path }} QUERY_STRING='${{ matrix.yaml_value_path }}' DEBUG_MODE=$DEBUG_MODE
 
+      - name: Check if PR is already open
```
jinja2, line 55: Can you clarify what is the motivation for doing this? We don't want open PR to be updated on changes to the base branch/main? I don't see any harm in it, so I am probably missing something.

.github/workflows/update_docker_images.yaml
```diff
@@ -50,9 +52,15 @@ jobs:
...
+        id: check_if_pr_open
+        run: |
+          git diff --exit-code main..update-${{ matrix.name }} -- helm-charts
```
jinja2, line 58: [default is to run shell with set -e](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell), so I think on diff or if the branch doesn't exist, the job will simply fail here.

.github/workflows/update_docker_images.yaml
atoulme: Clarified in the description of the PR.

.github/workflows/update_docker_images.yaml
atoulme: makes sense, I'll update. Thanks!

jvoravong (approved):

PR #1108, opened 2024-01-08
agent: support discovery mode on linux

rmfitzpatrick:
**Description:**
Adding a feature - These changes adopt discovery mode for agents, configurable with a new `agent.discovery` value mapping (disabled by default).

Right now only an `enabled` value and `extensions` and `receivers` properties are configurable. I originally planned on adding a `duration` field that will influence the `SPLUNK_DISCOVERY_DURATION` env var, but decided to defer for a followup since the liveness/readiness probe `initialDelaySeconds` would also need to be adjusted.

**Testing:**
My plan is for the integration tests be in the SOC project where the feature exists, with initial "test" being pushed implicitly to the example yaml content (and lack of changes in others).

**Documentation:**
Updated advanced configuration doc.

### DIFF
--- a/examples/discovery-mode/discovery-mode-values.yaml
+++ b/examples/discovery-mode/discovery-mode-values.yaml
@@ -1,0 +1,30 @@
+clusterName: CHANGEME
+splunkObservability:
+  realm: CHANGEME
+  accessToken: CHANGEME
+agent:
+  discovery:
+    enabled: true
+    properties:
+      receivers:
+        redis:
+          config:
+            password: '${env:REDIS_PASSWORD}'
+            username: '${env:REDIS_USERNAME}'
+  # these env vars reference a secret that must be created manually
+  extraEnvs:
+    - name: REDIS_USERNAME
+      valueFrom:
+        secretKeyRef:
+          name: redis-secret
+          key: username
+    - name: REDIS_PASSWORD
+      valueFrom:
+        secretKeyRef:
+          name: redis-secret
+          key: password
+
+
+clusterReceiver:
+  # disabled for example only. unrelated to discovery mode
+  enabled: false
--- a/examples/discovery-mode/discovery-mode-values.yaml
+++ b/examples/discovery-mode/discovery-mode-values.yaml
@@ -24,7 +24,6 @@ agent:
           name: redis-secret
           key: password
 
-
 clusterReceiver:
   # disabled for example only. unrelated to discovery mode
   enabled: false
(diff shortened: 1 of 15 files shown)


examples/discovery-mode/discovery-mode-values.yaml
```diff
@@ -0,0 +1,30 @@
...
+agent:
+  discovery:
+    enabled: true
+    properties:
+      receivers:
+        redis:
+          config:
+            password: '${env:REDIS_PASSWORD}'
+            username: '${env:REDIS_USERNAME}'
+  # these env vars reference a secret that must be created manually
```
jvoravong, line 14: Can you add the steps to deploy redis and the associated secret to the example README.md?

examples/discovery-mode/discovery-mode-values.yaml
```diff
@@ -0,0 +1,30 @@
...
+        secretKeyRef:
+          name: redis-secret
+          key: username
+    - name: REDIS_PASSWORD
+      valueFrom:
+        secretKeyRef:
+          name: redis-secret
+          key: password
+
+
```
jvoravong, line 27:
```suggestion
```

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -424,6 +424,13 @@ agent:
   # existing fields can be disabled by setting them to null value.
   config: {}
 
+  # Discovery mode attempts to automatically configure the agent with bundled metric receiver configuration
```
jvoravong, line 427:
```suggestion
  # Discovery mode attempts to automatically configure the agent with bundled metric receiver configuration.
  # For more details, refer to: https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/advanced-configuration.md#discovery-mode

```

examples/discovery-mode/discovery-mode-values.yaml
rmfitzpatrick: This seems a bit out of scope for these examples and what I'd think their purpose to be (values and manifests relevant to this helm chart). Is there a precedent for things like this? (I'd think they should be moved to another project or ~docs if so).

jvoravong: - The example README.md should provide all essential setup details. Specifically for this example, I'd recommend at least adding notes to the README.md that mention the prerequisite of having Redis and associated secret being deployed.
- The other non-k8s related receiver example I know of in this project is the the [Kafka example](https://github.com/signalfx/splunk-otel-collector-chart/tree/main/examples/add-kafkametrics-receiver).

I don't think this is worth holding up this PR since everything else looks good.

jvoravong (approved):

dmitryax (approved):

PR #1109, opened 2024-01-09
[chore] Test journald log collection with new tests

atoulme:
**Description:**
This PR has no functional changes to the helm chart. It redefines a test to check that journald log ingestion works as it should by using go code.


atoulme:
Rebasing.

PR #1110, opened 2024-01-09
[agent] change the default journald directory

atoulme:
**Description:**
Change the default journald directory to match the default of the receiver:
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/journaldreceiver

### DIFF
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -599,8 +599,8 @@ logsCollection:
 # Configuration for collecting journald logs using otel collector
   journald:
     enabled: false
-    # Please update directory path for journald if it's different from below default value "/var/log/journal"
-    directory: /var/log/journal
+    # Please update directory path for journald if it's different from below default value "/run/log/journal"
+    directory: /run/log/journal
     # List of service units to collect journald logs for and configuration for each.
     units:
       - name: kubelet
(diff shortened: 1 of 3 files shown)


jvoravong (approved):

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -599,8 +599,8 @@ logsCollection:
 # Configuration for collecting journald logs using otel collector
   journald:
     enabled: false
-    # Please update directory path for journald if it's different from below default value "/var/log/journal"
-    directory: /var/log/journal
+    # Please update directory path for journald if it's different from below default value "/run/log/journal"
+    directory: /run/log/journal
```
rmfitzpatrick, line 603: Why would we need to specify a value to begin with?

atoulme: oh true, the default value can be an empty string.

rmfitzpatrick: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/0d4683eb03d3f1aebb427fa094e0baff638deab6/pkg/stanza/operator/input/journald/journald.go#L101 empty string would still cause setting the `--directory` option so I think leaving the field commented would be preferable?

atoulme: understood, I'll revert and use the `/run/log/journal` then.

PR #1117, opened 2024-01-16
Add .Net Auto-instrumentation Developer Documentation

jvoravong:
**Description:** 
- Add support developer documentation for Splunk .Net auto-instrumentation in Kubernetes.
- The mentioned .NET demo will be added in a follow up PR.
- Adding tests in a follow up PR to keep the PR diff low.

**Testing:** <Describe what testing was performed and which tests were added.>
- Adding functional tests in a follow up PR.

**Documentation:** <Describe the documentation added.>
- Added documentation for how to use .Net auto-instrumentation. [Customer documentation](https://docs.splunk.com/observability/en/gdi/opentelemetry/auto-instrumentation/auto-instrumentation-dotnet-k8s.html) has been created based off of initial changes and feedback.

PR #1129, opened 2024-01-24
Fix issue where the collector agent exporter endpoint used in operator auto-instrumentation was missing the proper IP address 

jvoravong:
- This PR addresses an issue where the Collector Agent Exporter endpoint, utilized in the Operator configuration for .NET and Python auto-instrumentation, lacked the correct IP address assignment. 
  - This flaw specifically impacted setups where trace data was exported to the Collector Agent without using the service (`agent.service.enabled: false`), which resulted in failed trace transmissions.
  - Trace data exported by other auto-instrumentation languages that are not .NET  and Python were unaffected. 
  - Trace data exported to the gateway and related service (`agent.enabled: false` and `gateway.enabled: true`) were unaffected.
  
Testing:
- Adding test coverage as part of https://github.com/signalfx/splunk-otel-collector-chart/pull/1117

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
+++ b/helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
@@ -32,7 +32,7 @@ Helper to ensure the correct usage of the Splunk OpenTelemetry Collector Operato
 {{/*
 Helper to define an endpoint for exporting telemetry data related to auto-instrumentation.
 - Determines the endpoint based on user-defined values or default agent/gateway settings.
-- Order of precedence: User-defined > Agent endpoint > Gateway endpoint
+- Order of precedence: User-defined > Agent service endpoint > Agent host port endpoint > Gateway endpoint
 */}}
 {{- define "splunk-otel-collector.operator.instrumentation-exporter-endpoint" -}}
   {{- /* Initialize endpoint variable */ -}}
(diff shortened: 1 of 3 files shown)


atoulme (approved):

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
```diff
@@ -163,6 +166,11 @@ Helper for generating environment variables for each instrumentation library.
 
   {{- /* Output final OTEL_EXPORTER_OTLP_ENDPOINT, if applicable based on input conditions */ -}}
   {{- if $customOtelExporterEndpoint }}
+    {{- /* {{- /* Ensure the SPLUNK_OTEL_AGENT env var is set with per language env vars to successfully use it in env var substitution */ -}}
```
jinja2, line 169:
```suggestion
    {{- /* Ensure the SPLUNK_OTEL_AGENT env var is set with per language env vars to successfully use it in env var substitution */ -}}
```

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
```diff
@@ -163,6 +166,11 @@ Helper for generating environment variables for each instrumentation library.
...
+    {{- if contains "SPLUNK_OTEL_AGENT" $customOtelExporterEndpoint -}}
+      {{- printf "- name: SPLUNK_OTEL_AGENT\n  valueFrom:\n    fieldRef:\n      apiVersion: v1\n      fieldPath: status.hostIP" | indent 6 }}
```
jinja2, line 171: Small feedback - the readability of these helpers would be better if we could keep yaml snippets as-is. But I understand that would require additional changes to how this is organized. Maybe something we can slowly improve upon.

jinja2 (approved):

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
jvoravong: I agree that readability could be improved. Initially, we aimed to use multiple helm template files/snippets, but the complexity escalated quickly for our needs, pushing the limits of what templates could handle. This design choice was driven by the need for seamless adoption of new operator versions, automatically integrating new upstream instrumentation languages through automation, which necessitated dynamic values and Helm Helpers.

Adding unit tests to showcase the creation and management of the Operator Instrumentation config would indeed be beneficial.

PR #1149, opened 2024-02-09
prepare release v0.93.0

jinja2:

Release chart with Splunk Opentelemetry Collector v0.93.0


dmitryax:
Looks like step 4 from https://github.com/signalfx/splunk-otel-collector-chart/blob/main/RELEASE.md#release-procedure is missed

jinja2:
Updated the chloggen-validation script

ci_scripts/chloggen-pr-validate.sh
```diff
@@ -51,11 +51,10 @@ for file in $CHANGED_FILES; do
 done
 
 # ---- Changelog Entry Validation ----
-if { [[ $HELM_CHART_UPDATED -eq 1 ]] || [[ $RENDERED_MANIFESTS_UPDATED -eq 1 ]]; } && [[ $CHLOGGEN_FILE_PRESENT -eq 0 ]]; then
+if { [[ $HELM_CHART_UPDATED -eq 1 ]]; } && [[ $CHLOGGEN_FILE_PRESENT -eq 0 ]]; then
```
jvoravong, line 54: I think we should keep the previous logic in this file. The Changelog validate workflow should pass without these changes since there were git updates (file removals) in the .chloggen dir.

jvoravong (approved):

ci_scripts/chloggen-pr-validate.sh
jvoravong: I think these changes are fine to move forward with as is. I'm going to add some improvements in a future PR for this code.

jinja2: The validation script is run before the one which updates so on local, `make chlog-update` errors due to this check, the ordering should be updated in the task. But also, I don't think there are many cases when changes to examples would necessitate a changelog entry.

jvoravong: - We can make improvements. I do think this issue is reproducible, wasn't expecting that. I'm really surprised [[chore] Prepare release 0.92.1 #1145](https://github.com/signalfx/splunk-otel-collector-chart/pull/1145/files#) didn't hit this issue.
- I think we really need to scan the rendered manifest for changes because we include subcharts. The subchart templates are stored in our our main chart template area. I'd be pro keeping the scanning of the rendered_manifest dirs, but I can be convinced otherwise.
  - Maybe an alternative is we can scan the Chart.yaml for appVersion or subchart version updates.

jinja2: Re: different helm version - Not sure why would that change the rendered manifests. It would most likely just error for the user when running render if the helm version does not support the functions we have in the chart. The only case I can see the rendered example changing is if somehow helm version has a breaking change which changes the output generated by function, and I expect we'll catch breaking changes like this if they ever occur.

I think scanning for updates to dependencies in Chart.yaml should be good enough, but if you prefer also scanning for examples, feel free to add back the logic once the ordering is fixed.

PR #1156, opened 2024-02-13
remove networkExplorer option

atoulme:
**Description:**
Remove `networkExplorer` option from the chart. This option was first deprecated in 0.88.0.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,9 @@
 # Upgrade guidelines
 
+# 0.93.0 to 0.94.0
+
+The `networkExplorer` option is removed.
+
 ## 0.87.0 to 0.88.0
 
 The `networkExplorer` option is deprecated now. Please use the upstream OpenTelemetry eBPF Helm chart to collect
--- a/helm-charts/splunk-otel-collector/templates/NOTES.txt
+++ b/helm-charts/splunk-otel-collector/templates/NOTES.txt
@@ -82,6 +82,3 @@ Splunk OpenTelemetry Collector is installed and configured to send data to Splun
   - Some libraries may be enabled by default. For current status, see: https://github.com/open-telemetry/opentelemetry-operator#controlling-instrumentation-capabilities
   - Splunk provides best-effort support for native OpenTelemetry libraries, and full support for Splunk library distributions. For used libraries, refer to the values.yaml under "operator.instrumentation.spec".
 {{- end }}
-{{- if not (eq (toString .Values.networkExplorer) "<nil>")}}
-[WARNING] `networkExplorer` option is no longer supported. Please remove it from your custom values.yaml.
-{{- end }}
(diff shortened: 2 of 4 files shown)


UPGRADING.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade guidelines
 
+# 0.93.0 to 0.94.0
+
+The `networkExplorer` option is removed.
```
jinja2, line 5:
```suggestion
The `networkExplorer` variable is removed from the chart. Please remove it from your custom values.yaml to avoid schema errors.
```

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -1409,10 +1409,6 @@
       "type": "boolean",
       "deprecated": true
     },
-    "networkExplorer": {
```
dmitryax, line 1412: If we remove it, the chart installation will be failing if `networkExplorer` is submitted since we don't allow unknown values. Do we want this?

helm-charts/splunk-otel-collector/templates/NOTES.txt
```diff
@@ -82,9 +82,3 @@ Splunk OpenTelemetry Collector is installed and configured to send data to Splun
   - Some libraries may be enabled by default. For current status, see: https://github.com/open-telemetry/opentelemetry-operator#controlling-instrumentation-capabilities
   - Splunk provides best-effort support for native OpenTelemetry libraries, and full support for Splunk library distributions. For used libraries, refer to the values.yaml under "operator.instrumentation.spec".
 {{- end }}
-{{- if and (not (eq (toString .Values.networkExplorer) "<nil>")) .Values.networkExplorer.enabled }}
-{{ fail "Network explorer is not part of this helm chart anymore. Please follow https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md#0870-to-0880 to replace it with the upstream OpenTelemetry eBPF helm chart." }}
-{{- end }}
```
dmitryax, line 87: I'm still not sure how this triggers the reported bug, but I think we should remove just this section (85-87 lines) and leave everything else for now.

dmitryax (approved):
I'm good with either: removing the section now and fail if the option is provided or just fixing the reported issue and keep the second warning for longer. Probably a product call

helm-charts/splunk-otel-collector/templates/NOTES.txt
atoulme: OK I'll revert to remove just the NOTES.txt.

helm-charts/splunk-otel-collector/values.schema.json
jinja2: I think it should be okay to drop it now. We have had the warnings for more than 5 versions now. Do you have a specific timeline in mind?

dmitryax: No specific timeline. If we think it's ready to be drop, I'm ok with that

atoulme:
OK we will first remove the use in 0.93.1, and keep this PR open for 0.94.0.

samiura (approved):

PR #1183, opened 2024-02-26
Enable validation of `enable` field

kristeey:
Since `additionalProperties: false` in the `values.schema.json` file, it is not possible to use common pattern to include `splunk-otel-collector` as a subchart using `condition` field in the parent charts `charts.yaml` (as explained in the helm documentation https://helm.sh/docs/topics/charts/#tags-and-condition-fields-in-dependencies). 

This PR adds a validation for `enabled` boolean field to solve the issue.

### DIFF
--- a/helm-charts/splunk-otel-collector/values.schema.json
+++ b/helm-charts/splunk-otel-collector/values.schema.json
@@ -8,6 +8,10 @@
   "title": "Values",
   "additionalProperties": false,
   "properties": {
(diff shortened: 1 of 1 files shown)


kristeey:

kristeey:
recheck

jvoravong (approved):

jvoravong:
This change looks useful, thanks for contributing @kristeey.

PR #1185, opened 2024-02-27
removed memory balast and using GOMEMLIMIT. OTL-2488

samiura:
**Description:** removed all instances of memory_ballast and using GOMEMLIMIT

**Link to Splunk idea:** [OTL-2488](https://splunk.atlassian.net/browse/OTL-2488)

**Testing:** added unit tests


helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -269,6 +269,8 @@ spec:
           {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $agent.securityContext "setRunAsUser" true) | nindent 10 }}
         {{- end }}
         env:
+          - name: GOMEMLIMIT
+            value: {{ include "splunk-otel-collector.convertMemToMib" $agent.resources.limits.memory | int64 }}
           - name: SPLUNK_MEMORY_TOTAL_MIB
```
pjanotti, line 274: Shouldn't `SPLUNK_MEMORY_TOTAL_MIB` be removed? If not how it interacts with `GOMEMLIMIT`?

pjanotti (approved):
LGTM, but, I have one Q.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -17,8 +17,6 @@ extensions:
     timeout: 0
   {{- end }}
 
-  memory_ballast:
-    size_mib: ${SPLUNK_BALLAST_SIZE_MIB}
```
jvoravong, line 20:
```suggestion
```

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -269,6 +269,8 @@ spec:
           {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $agent.securityContext "setRunAsUser" true) | nindent 10 }}
         {{- end }}
         env:
+          - name: GOMEMLIMIT
+            value: {{ include "splunk-otel-collector.convertMemToMib" $agent.resources.limits.memory | int64 }}
```
jvoravong, line 273: The test might be failing here due to a quotation or some other issue, likely also affects deployment-gateway.yaml.
```suggestion
            value: "{{ include "splunk-otel-collector.convertMemToMib" $agent.resources.limits.memory | int64 }}"
```

helm-charts/splunk-otel-collector/templates/daemonset.yaml
samiura: As far as I understand it has no effect from the tests on the internal realms. this is a good point, may be we can investigate the idea of removing it.

.chloggen/.OTL2488.yaml.un~
dmitryax, line 1: Please remove this file :)

helm-charts/splunk-otel-collector/templates/daemonset.yaml
dmitryax: If we can set it dynamically in https://github.com/signalfx/splunk-otel-collector/blob/91f039d015bdd5792e172ea635bc9b1dfd07a101/internal/settings/settings.go from `SPLUNK_MEMORY_TOTAL_MIB`, that would be great. So we don't need to provide two separate env vars. I remember @atoulme mentioned that he tried that.

Issue #1011, opened 2023-11-09
Include namespace in metadata 


dmitchsplunk:
When using helm to generate templates with the splunk-otel-collector chart: 

`helm template --namespace mynamespace --values values.yaml --output-dir <rendered_manifests_dir> --generate-name splunk-otel-collector-chart/splunk-otel-collector`

The desired namespace is not added to the metadata section of the resulting templates.   For example, [daemonset.yaml](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/daemonset.yaml) does not include the namespace. 

The request is to add the namespace to the generated templates, similar to how it's added in the upstream collector chart.  See [daemonset.yaml](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/daemonset.yaml) as an example: 

```
metadata:
  name: {{ include "opentelemetry-collector.fullname" . }}-agent
  namespace: {{ template "opentelemetry-collector.namespace" . }}
```
  
The workaround is to add the namespace manually to the rendered manifests, or add the "--namespace" argument when applying the manifests with kubectl.

atoulme:
We will be working towards implementing this support. Please follow up internally for updates. Thanks!

PR #1189, opened 2024-02-28
Add namespace to metadata

sbylica-splunk:
**Description:** Add namespace to metadata

**Link to Splunk idea:** https://github.com/signalfx/splunk-otel-collector-chart/issues/1011

**Testing:** Manual testing

**Documentation:**-


jvoravong:
Hey, could we discuss the potential benefits of this change for specific use cases?

- Currently, our [official customer documentation](https://docs.splunk.com/observability/en/gdi/opentelemetry/collector-kubernetes/install-k8s.html#resource-yaml-manifests) mentions that the rendered manifests can be directly used, provided the CHANGEME values are properly updated. Introducing a default namespace value might conflict with this use case, unless we consider additional adjustments.
- Since Helm automatically includes the namespace metadata value for all deployed resources, it might be more efficient to adjust the code generating the rendered manifests, rather than modifying each template individually.

jvoravong:
I'm using Helm v3.12.3 and the namespace metadata is added to deployed objects like the Daemonset for the agent. I wonder if this was not true when using older versions of Helm.

helm-charts/splunk-otel-collector/templates/clusterRole.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: {{ template "splunk-otel-collector.fullname" . }}
+  namespace: {{ .Release.Namespace }}
```
jinja2, line 6: This is a cluster scoped object, no need to add namespace here.

helm-charts/splunk-otel-collector/templates/clusterRoleBinding.yaml
```diff
@@ -3,6 +3,7 @@ kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ template "splunk-otel-collector.fullname" . }}
+  namespace: {{ .Release.Namespace }}
```
jinja2, line 6: This is a cluster scoped object, no need to add namespace here.

helm-charts/splunk-otel-collector/templates/configmap-agent.yaml
```diff
@@ -7,6 +7,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ template "splunk-otel-collector.fullname" . }}-otel-agent
+  namespace: {{ .Release.Namespace }}
```
jinja2, line 10: It would be good to have an additional variable `namespaceOverride` in the chart. This would give users the option to create collector resources in a different namespace (different from where the original chart release is stored) when this chart is used as a sub-chart. Also makes it easy to generate manifests cause the user can just add the var to values and not have to change commands.

A helper template like this can be used then
```
{{- define "splunk-otel-collector.namespace" -}}
{{- default .Release.Namespace .Values.namespaceOverride | trunc 63 | trimSuffix "-" -}}
{{- end -}}
```

In the objects, use
```
namespace: {{ template "splunk-otel-collector.namespace" . }}
```

You'd probably need to replace `{{ .Release.Namespace }}` with the above template in a few other places.

helm-charts/splunk-otel-collector/templates/securityContextConstraints.yaml
```diff
@@ -41,6 +41,7 @@ kind: SecurityContextConstraints
 apiVersion: security.openshift.io/v1
 metadata:
   name: {{ template "splunk-otel-collector.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
```
jinja2, line 44: I think this resource is also cluster-scoped. @jvoravong can you confirm if `SecurityContextConstraints` in OC is cluster or namespace scoped?

jinja2 (changes requested):

jinja2:
> I'm using Helm v3.12.3 and the namespace metadata is added to deployed objects like the Daemonset for the agent. I wonder if this was not true when using older versions of Helm.

@jvoravong While the final objects generated by `helm install` for deploy has the namespace field, `helm template --namespace NS` does not output manifests with `metadata.namespace` set unless we explicitly set it as it is being done in this PR. I can see this addition to the chart being useful for customers who do not want to use helm for deployment and would rather generate complete manifests with `helm template` to apply with kubectl or other client.

jinja2:
> Hey, could we discuss the potential benefits of this change for specific use cases?
> 
> * Currently, our [official customer documentation](https://docs.splunk.com/observability/en/gdi/opentelemetry/collector-kubernetes/install-k8s.html#resource-yaml-manifests) mentions that the rendered manifests can be directly used, provided the CHANGEME values are properly updated. Introducing a default namespace value might conflict with this use case, unless we consider additional adjustments.

Yeah, the official documentation will need to be updated. Even right now the instructions in the doc for using the manifests as-is really only work when deploying to `default` namespace since we have configs (for e.g. [namespace of sa in binding](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/examples/add-filter-processor/rendered_manifests/clusterRoleBinding.yaml#L24)) which use the `Release.Namespace` field which is `default` for the rendered manifests.
My preference would be to only keep the section about generating manifests from templates for users not wanting to use helm for deployment. This would provide a better experience for users since they do not have to deal with updating fields in manifests and only need to set relevant info in value.yaml. But if we must support direct manifests also, then a compromise could be to change the namespace to `CHANGEME-NS` or something, and update the docs with a note for the user to update this value in the manifests which the namespace they are trying to deploy to. 

> * Since Helm automatically includes the namespace metadata value for all deployed resources, it might be more efficient to adjust the code generating the rendered manifests, rather than modifying each template individually.
The change will make it possible for `helm template --namespace` to update the examples with namespace w/o the need to manually update the yaml.

helm-charts/splunk-otel-collector/templates/securityContextConstraints.yaml
jvoravong: [Cluster scoped](https://docs.openshift.com/container-platform/4.15/authentication/managing-security-context-constraints.html#role-based-access-to-ssc_configuring-internal-oauth).

jvoravong:
> > * Since Helm automatically includes the namespace metadata value for all deployed resources, it might be more efficient to adjust the code generating the rendered manifests, rather than modifying each template individually.
> 
> The change will make it possible for `helm template --namespace` to update the examples with namespace w/o the need to manually update the yaml.

Started https://github.com/signalfx/splunk-otel-collector-chart/pull/1192 to adjust the code generating the rendered manifests so we can customize render outputs using custom values inputs.

helm-charts/splunk-otel-collector/templates/clusterRole.yaml
sbylica-splunk: Removed it, but is there a particular reason to not include a namespace in cluster scoped objects?

helm-charts/splunk-otel-collector/templates/clusterRoleBinding.yaml
sbylica-splunk: Removed

helm-charts/splunk-otel-collector/templates/securityContextConstraints.yaml
sbylica-splunk: Also removed

helm-charts/splunk-otel-collector/templates/configmap-agent.yaml
sbylica-splunk: Ok, implemented it this way and it looks good!

helm-charts/splunk-otel-collector/templates/clusterRole.yaml
jvoravong: - I've seen the namespace metadata be ignored or cause Helm manifest validation errors.
- I know this would cause Helm validation errors in our EKS add-on environment.

helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
```diff
@@ -181,3 +181,12 @@ Helper for generating environment variables for each instrumentation library.
...
 {{- end }}
+
+{{/*
+Helper used to define a namspace.
+- Returns namespace from a release
+- If namespaceOverride value is filled in it will replace the namespace
+*/}}
+{{- define "splunk-otel-collector.namespace" -}}
+  {{- default .Release.Namespace .Values.namespaceOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
```
jvoravong, line 192: I'd move this helper up to the helm-charts/splunk-otel-collector/templates/_helpers.tpl file.

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -19,6 +19,10 @@
       "description": "Override name of the chart used in Kubernetes object names.",
       "type": "string"
     },
+    "namespaceOverride": {
+      "description": "Override namespace used in Kubernetes object names.",
```
jinja2, line 23:
```suggestion
      "description": "Override the deployment namespace of chart resources; default namespace is .Release.Namespace",
```

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -7,6 +7,8 @@
 nameOverride: ""
 # fullnameOverride completely replaces the generated name.
 fullnameOverride: ""
+# namespaceOverride replaces namespace in templates.
```
jinja2, line 10:
```suggestion
# namespaceOverride can be used override the deployment namespace for collector resources
# Useful when including this chart as a subchart, so it can be released into a different namespace than the parent
```

jvoravong:
@sbylica-splunk I think you will have to pull latest changes from main and run `make render` to clear up the conflicts.

jinja2:
@sbylica-splunk there was a new version of chart release after this PR was opened which changes the calculated checksum in the examples. You'll have to locally rebase and resolve conflicts to move this along

sbylica-splunk:
Merged changes from the new release

jvoravong (approved):

jinja2 (approved):

PR #1193, opened 2024-02-29
add test for helm install in AKS with windows

jinja2:

### DIFF
--- a/functional_tests/testdata/values/aks_test_values.yaml.tmpl
+++ b/functional_tests/testdata/values/aks_test_values.yaml.tmpl
@@ -1,12 +1,9 @@
 ---
-metricsEnabled: true
-tracesEnabled: true
 splunkObservability:
   realm:       CHANGEME
   accessToken: CHANGEME
   ingestUrl: {{ .AgentEndpoint }}
   apiUrl: {{ .ApiURLEndpoint }}
-  metricsEnabled: true
 
 splunkPlatform:
   token: foobar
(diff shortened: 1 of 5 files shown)


functional_tests/testdata/values/aks_test_values.yaml.tmpl
```diff
@@ -0,0 +1,82 @@
...
+
+clusterName: test
+environment: test
+cloudProvider: azure
+distribution: aks
+isWindows: true
+image:
+  otelcol:
+    repository: quay.io/signalfx/splunk-otel-collector-windows-dev
+    tag: latest
```
jinja2, line 69: this is to test with multi platform image since the cluster has both windows 2019/2022

jvoravong: Might be worth adding an line comment if we are going to use a different image for these tests compared to what a customer might use (quay.io/signalfx/splunk-otel-collector-windows)

jvoravong (approved):
LGTM, a couple small things to check.

functional_tests/testdata/values/aks_test_values.yaml.tmpl
```diff
@@ -0,0 +1,82 @@
+---
+metricsEnabled: true
+tracesEnabled: true
+splunkObservability:
```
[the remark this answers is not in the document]
jinja2: good catch, will remove it

PR #1233, opened 2024-03-22
Update spring petclinic

jvoravong:
- Ran the update-demos.sh script to get the latest changes from the up stream https://github.com/spring-petclinic/spring-petclinic-microservices project for this project's spring-petclinic
- These changes aim to update spring-petclinic using automation. This update add spring-petclinic demo update enhancements. Due to maintenance constraints with our tools, resource memory values are in bytes rather than Mi units.

### DIFF
--- a/examples/enable-operator-and-auto-instrumentation/spring-petclinic/spring-petclinic.yaml
+++ b/examples/enable-operator-and-auto-instrumentation/spring-petclinic/spring-petclinic.yaml
@@ -128,24 +128,12 @@ spec:
           name: admin-server
           ports:
             - containerPort: 9090
+              hostPort: 9090
+              protocol: TCP
           resources:
             limits:
               memory: "536870912"
       restartPolicy: Always
----
-apiVersion: networking.k8s.io/v1
-kind: NetworkPolicy
-metadata:
-  name: spring-petclinic-default
-spec:
-  ingress:
-    - from:
-        - podSelector:
-            matchLabels:
-              io.kompose.network/spring-petclinic-default: "true"
-  podSelector:
-    matchLabels:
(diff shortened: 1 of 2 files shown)


jvoravong:
Several tests are currently failing that are unrelated to these changes.

examples/enable-operator-and-auto-instrumentation/spring-petclinic/spring-petclinic.yaml
```diff
@@ -128,25 +128,13 @@ spec:
           name: admin-server
           ports:
             - containerPort: 9090
+              hostPort: 9090
```
jinja2, line 131: hmm, don't think we need to expose host port for these. Let me check why this is added by kompose ( i guess the docker file has it exposed as 9090:9090 but still no reason to expose on k8s node on conversion)

jinja2 (approved):

examples/enable-operator-and-auto-instrumentation/spring-petclinic/spring-petclinic.yaml
jvoravong: Fair, I'll look into it more. Kompose automation is adding this and I would like to keep these demos automatically generated to a certain extend.

jinja2: yeah, maybe just remove it in the script after kompose comvert is run with something like `yq -i 'del(.spec.template.spec.containers[].ports[].hostPort) | del(.spec.template.spec.containers[].ports[].protocol)' file`

jvoravong: Done

PR #1249, opened 2024-04-12
Remove SAPM exporter, use OTLP exporter

atoulme:
**Description:**
Remove the SAPM exporter and use OTLP exporter for all trace ingest.

**Testing:**
Regenerated examples. Functional tests already use otlp for trace exports.

**Documentation:**
Replaced mention in README.
Added upgrade guidelines, and a note in NOTES.txt to display. I also added an example showing a situation that will trigger the warning in NOTES.txt.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/NOTES.txt
+++ b/helm-charts/splunk-otel-collector/templates/NOTES.txt
@@ -82,3 +82,7 @@ Splunk OpenTelemetry Collector is installed and configured to send data to Splun
   - Some libraries may be enabled by default. For current status, see: https://github.com/open-telemetry/opentelemetry-operator#controlling-instrumentation-capabilities
   - Splunk provides best-effort support for native OpenTelemetry libraries, and full support for Splunk library distributions. For used libraries, refer to the values.yaml under "operator.instrumentation.spec".
 {{- end }}
+{{- if and .Values.agent.config .Values.agent.config.exporters (not (eq (toString .Values.agent.config.exporters.sapm) "<nil>")) }}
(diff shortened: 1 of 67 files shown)


samiura (approved):

helm-charts/splunk-otel-collector/templates/NOTES.txt
```diff
@@ -82,3 +82,7 @@ Splunk OpenTelemetry Collector is installed and configured to send data to Splun
   - Some libraries may be enabled by default. For current status, see: https://github.com/open-telemetry/opentelemetry-operator#controlling-instrumentation-capabilities
   - Splunk provides best-effort support for native OpenTelemetry libraries, and full support for Splunk library distributions. For used libraries, refer to the values.yaml under "operator.instrumentation.spec".
 {{- end }}
+{{- if not (eq (toString ((.Values.agent.config).exporters).sapm) "<nil>") }}
```
dmitryax, line 85: This should work
```suggestion
{{- if not (eq (toString .Values.agent.config.exporters.sapm) "<nil>") }}
```

dmitryax (approved):
One comment. Otherwise, LGTM

helm-charts/splunk-otel-collector/templates/NOTES.txt
dmitryax: Also, need to check the same for gateway

atoulme: No, that fails with `make render`:
```
Error: template: splunk-otel-collector/templates/NOTES.txt:85:32: executing "splunk-otel-collector/templates/NOTES.txt" at <.Values.agent.config.exporters.sapm>: nil pointer evaluating interface {}.sapm
```

PR #1281, opened 2024-05-03
[chore] Update functional tests to add v1.30 Kubernetes version coverage

jvoravong:


jvoravong:
Looks like the latest major versions (1.29, 1.28, 1.27, 1.26) of Kubernetes provided by Kind are causing the functional v2 tests to fail. Going to only remove updating the Kind Kubernetes versions for now, will only update the Minikube versions.

.github/workflows/functional_test.yaml
```diff
@@ -11,14 +11,14 @@ jobs:
...
-          - v1.29.0 # Support: 28 Dec 2024 - 28 Feb 2025
-          - v1.28.4 # Support: 28 Aug 2024 - 28 Oct 2024
-          - v1.27.8  # Support: 28 Apr 2024 - 28 Jun 2024
-          - v1.26.11 # Support: 28 Dec 2023 - 28 Feb 2024
-          - v1.25.16 # Support: 27 Aug 2023 - 27 Oct 2023
+          - v1.30.0 # Support: 28 Apr 2025 - 28 Jun 2025
+          - v1.29.4 # Support: 28 Dec 2024 - 28 Feb 2025
+          - v1.28.9 # Support: 28 Aug 2024 - 28 Oct 2024
+          - v1.27.13  # Support: 28 Apr 2024 - 28 Jun 2024
+          - v1.26.15 # Support: 28 Dec 2023 - 28 Feb 2024
```
crobert-1, line 21: If support for this has ended should we still be testing it?

jvoravong: We should continue this practice for now. An existing comment above indicates that we test the current native Kubernetes versions plus one out-of-date version, primarily to accommodate distributions like EKS, which supports older versions not covered by native Kubernetes, and we support EKS.

I'll update the comments to highlight the point above.

crobert-1 (approved):

PR #1292, opened 2024-05-14
Fix issue where SPLUNK_OTEL_AGENT env var was set before custom opeartor.instrumentation.spec.env

jvoravong:
- This small bug fix makes it so customer provided env vars (using opeartor.instrumentation.spec.env) can use the env var SPLUNK_OTEL_AGENT with var substitution. 
- Before, SPLUNK_OTEL_AGENT was defined after the customer provided env vars and value var substitution was not possible.

### DIFF
--- a/.chloggen/inst-update.yaml
+++ b/.chloggen/inst-update.yaml
@@ -1,5 +1,5 @@
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: bug
+change_type: bug_fix
 # The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
 component: operator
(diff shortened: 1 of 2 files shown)


.chloggen/inst-update.yaml
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug
```
crobert-1, line 2:
```suggestion
change_type: bug_fix
```
Fixes failing CI/CD

crobert-1 (approved):

.chloggen/inst-update.yaml
jvoravong: Ty

jinja2 (approved):

PR #1301, opened 2024-05-22
Add support debug info script

jvoravong:


atoulme:
This probably wouldn't go under ci_scripts?

jvoravong:
> Should this script be under ci_scripts?
- I still need to address a couple of minor issues before changing the status of this PR.
- We can consider placing the script in a different location. One option is to make it runnable via a Helm chart value and template.
  - By leveraging a Helm value, Helm template, and a Kubernetes job, we might simplify its usability. I am currently verifying the feasibility of this approach.
  - Alternatively, creating a new subdirectory called something like /tools might be an appropriate location for this script in this repo. Taking suggestions for a new subdirectory.

jvoravong:
> This probably wouldn't go under ci_scripts?

@atoulme moved to ./tools

examples/add-filter-processor/rendered_manifests/secret-splunk.yaml
```diff
@@ -7,12 +7,12 @@ metadata:
   namespace: default
   labels:
     app.kubernetes.io/name: splunk-otel-collector
-    helm.sh/chart: splunk-otel-collector-0.103.0
+    helm.sh/chart: splunk-otel-collector-0.104.0
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: default
-    app.kubernetes.io/version: "0.103.0"
+    app.kubernetes.io/version: "0.104.0"
```
dmitryax, line 13: Why the upgrade here? Can you please break it down into several PRs? Each PR to serve a particular purpose

jvoravong: Sorry about that, accidentally had my git command automation push to the wrong branch. Reverted the chart upgrade changes, only two files are created or updated in this PR.

PR #1325, opened 2024-06-18
Update OpenShift security context constraints to support the operator service account

jvoravong:
**Description:**
- **Bug Fix:** Addressed an issue where the Operator was not always deployable to OpenShift environments due to the lack of a validating Security Context Constraint.
- **Updates:**
  - Enhanced Security Context Constraints (SCC) for OpenShift to fix formatting issues and include support for the operator service account.
  - Changed the SCC `fsGroup` type from `MustRunAs` to `RunAsAny` to avoid issues with dynamic default `fsGroup` constraint value ranges, which could prevent the operator from deploying in OpenShift environments. Users can still override SCC settings using the [`securityContextConstraintsOverwrite`](https://github.com/signalfx/splunk-otel-collector-chart/blob/320b40a492bc479b12beb4aad20a85e1a9fd12c1/helm-charts/splunk-otel-collector/values.yaml#L1052) value.
  - Refer to the OpenShift documentation for details: [OpenShift SCC Strategies](https://docs.openshift.com/container-platform/4.15/authentication/managing-security-context-constraints.html#authorization-SCC-strategies_configuring-internal-oauth)
    - _FSGroup - MustRunAs - Requires at least one range to be specified if not using pre-allocated values. Uses the minimum value of the first range as the default. Validates against the first ID in the first range._

**Testing:**
- Reproduced the bug and tested the fix with OpenShift 4.12.11 (Kubernetes v1.25.7) to ensure successful deployment of the Operator. Testing other OpenShift versions as well.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/securityContextConstraints.yaml
+++ b/helm-charts/splunk-otel-collector/templates/securityContextConstraints.yaml
@@ -10,11 +10,11 @@ allowHostDirVolumePlugin: true
 allowHostIPC: false
 allowPrivilegedContainer: false
 volumes:
-- configMap
-- downwardAPI
-- emptyDir
-- hostPath
-- secret
+  - configMap
+  - downwardAPI
+  - emptyDir
+  - hostPath
+  - secret
 # Allow podman/crio socket and /proc access
 seLinuxContext:
   type: MustRunAs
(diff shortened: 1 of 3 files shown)


jvoravong:
The functional tests are failing with the current main branch, looking into fixing them.

helm-charts/splunk-otel-collector/templates/securityContextConstraints.yaml
```diff
@@ -49,7 +49,9 @@ metadata:
     heritage: {{ .Release.Service }}
 users:
 - system:serviceaccount:{{ template "splunk-otel-collector.namespace" . }}:{{ template "splunk-otel-collector.serviceAccountName" . }}
-
+{{- if .Values.operator.enabled }}
+- system:serviceaccount:{{ include "splunk-otel-collector.namespace" . }}:operator
```
jinja2, line 53:
```suggestion
- system:serviceaccount:{{ include "splunk-otel-collector.namespace" . }}:{{ template "opentelemetry-operator.serviceAccountName" .Subcharts.operator }}
```

jvoravong: Good point

jinja2 (approved):

PR #1330, opened 2024-06-26
feat: add optional extra configs for agent and cluster receiver

chenlujjj:

Now the chart has defined some configs in the `templates/config` folder, and allow users to override them in the `config` part of the `values.yaml` file. For most use cases, it works perfect. 

However, in our company there are more than 100 k8s clusters and the collector configuration differs here and there across these clusters. At the beginning we maintain the specific configs in each cluster's collector `values.yaml` file, but it quickly becomes a heavy burden. We need `templating` functionality for these extra configs, but helm doesn't support templating for values at the moment. So we come up with an idea: allow the user provide self-defined configmap as the collector configuration, which will be appended to the running command of the collector container.

For example, the command is `/otelcol --config=/conf/relay.yaml` now. If there exists extra configs, it will be `/otelcol --config=/conf/relay.yaml --config=/conf-extra/relay.yaml`

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/daemonset.yaml
+++ b/helm-charts/splunk-otel-collector/templates/daemonset.yaml
@@ -249,6 +249,11 @@ spec:
         {{- else }}
         - /otelcol
         - --config=/conf/relay.yaml
+        {{- if $agent.extraConfigs.enabled }}
+        {{- range $key := $agent.extraConfigs.configMapKeys }}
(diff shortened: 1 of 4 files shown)


jinja2:
Hi @chenlujjj. Thank you for opening this PR. It would be best to file a Splunk idea to add this feature to our helm chart. Thanks!

chenlujjj:
Hi @jinja2 ,I've file a idea here: https://ideas.splunk.com/ideas/EID-I-2345

PR #1349, opened 2024-07-25
Add support for splunk-otel-java v2.x.x

jvoravong:
- Adopt Java 2.x in the collector and chart
- Update any failing tests as needed
- Publish any breaking changes notes as needed


atoulme:
` --- FAIL: Test_Functions/java_traces_captured (181.00s)` as expected, this changes the format of the traces we expose quite a bit.

atoulme:
Quick hunch, can you ignore the resource attribute `network.peer.port` ?

aunshc:
Quick note once this is out of the draft state: we should add this to the changelog and reference that runtime metrics will now be enabled by default with Java 2.x by linking to this migration guide as well https://docs.splunk.com/observability/en/gdi/get-data-in/application/java/migrate-metrics.html.

atoulme:
very close, please add one ignore for the resource attribute "telemetry.distro.version"

atoulme (approved):
LGTM, asking @breedx-splk to review, esp docs.

breedx-splk (approved):
Seems fine to me. I didn't cross-check the entire list of metric name and attribute changes, but it seems about right.

PR #1376, opened 2024-08-08
Add metricsSourceType tracesSourceType

mstopa-splunk:
**Description:** Added `metricsSourceType` and `tracesSourceType` to the chart
**Link to Splunk idea:**
**Testing:** manual
**Documentation:** N/A


helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -170,6 +176,13 @@
               "const": true
             }
           }
+        },
+        {
+          "properties": {
+            "tracesEnabled": {
```
sbylica-splunk, line 182: What's this?

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -391,6 +391,9 @@ splunk_hec/platform_metrics:
   token: "${SPLUNK_PLATFORM_HEC_TOKEN}"
   index: {{ .Values.splunkPlatform.metricsIndex | quote }}
   source: {{ .Values.splunkPlatform.source | quote }}
+  {{- if .Values.splunkPlatform.metricsSourceType }}
+  sourcetype: {{ .Values.splunkPlatform.metricsSourceType | quote }}
+  {{- end }}
```
sbylica-splunk, line 396: I feel like it may get confusing since we are adding a sourcetype to logs with "resource/logs" processor. Maybe we could create something similar for metrics/traces?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -49,6 +49,10 @@ splunkPlatform:
   # Optional. Default value for `sourcetype` field. For container logs, it will
   # be container name.
   sourcetype: ""
```
jvoravong, line 51: - If the sourcetype is the same across logs, metrics, and traces, we should stick with only using `sourcetype` and avoid introducing `metricsSourceType` and` tracesSourceType`.
- If the sourcetype can be different for each telemetry type (logs, metrics, and traces), we should migrate sourcetype to logsSourcetype. Then we would have `logsSourcetype`, `metricsSourceType` and `tracesSourceType` accordingly.

I'm not certain which use case is more common, so it might be better to keep things simple (using the first option) if we're unsure which one to choose.

sbylica-splunk: I think that we should stick to 1) unless there is a good reason to change it

helm-charts/splunk-otel-collector/values.schema.json
mstopa-splunk: @sbylica-splunk this allows you a setup for splunkPlatform with only traces enabled.
Without it you must enable either logs or traces. Do we want to remove this change?

helm-charts/splunk-otel-collector/values.yaml
mstopa-splunk: thanks @jvoravong. @sbylica-splunk please see the new diff

helm-charts/splunk-otel-collector/templates/config/_common.tpl
mstopa-splunk: @sbylica-splunk I can work on that.
On the other hand we set all other Splunk metadata for logs/metrics/traces in the hec exporters, so it's also quite an intuitive place for setting the sourcetype as well.
What do you think?

helm-charts/splunk-otel-collector/values.schema.json
sbylica-splunk: Ah, makes sense now that I'm looking at it. Is there any particular reason for this change?

helm-charts/splunk-otel-collector/templates/config/_common.tpl
sbylica-splunk: Yeah, that makes more sense in my opinion. Do mind though that this would add a sourcetype everywhere - to the gateway and receiver too. If that's okay then I think we can move it to hec_exporters

helm-charts/splunk-otel-collector/values.schema.json
mstopa-splunk: @sbylica-splunk it's inconsistent when we accept either only logs or only metrics, but not traces, but we don't need it necessarily but, this can be removed

helm-charts/splunk-otel-collector/values.yaml
mstopa-splunk: @jvoravong we finished on our end, can you please review?

helm-charts/splunk-otel-collector/templates/config/_common.tpl
jvoravong: I think there are still some gaps here for telemetry data support for a `splunk-platform-enable-all-telemetry` values configuration. 
- Can you create a example in the /examples dir so we can see the for differences this PR would add? I think it would would be worth having an example where all telemetry types are enabled for `splunkPlatform.*`. This will likely make reviews easier as well.
- You can also even add unit tests under https://github.com/signalfx/splunk-otel-collector-chart/tree/main/test for unit validation tests to help ensure added Helm template logic.

jvoravong (approved):

PR #1380, opened 2024-08-08
[chore] retry installing the pod monitor until CRDs are all deployed

atoulme:


pjanotti (approved):

functional_tests/functional_test.go
```diff
@@ -334,14 +334,18 @@ func deployChartsAndApps(t *testing.T) {
...
-	_, err = dynamicClient.Resource(g).Namespace("default").Create(context.Background(), podMonitor.(*unstructured.Unstructured), metav1.CreateOptions{})
-	if err != nil {
-		_, err2 := dynamicClient.Resource(g).Namespace("default").Update(context.Background(), podMonitor.(*unstructured.Unstructured), metav1.UpdateOptions{})
-		assert.NoError(t, err2)
-		if err2 != nil {
-			require.NoError(t, err)
+	// CRDs sometimes take time to register. We retry deploying the pod monitor until such a time all CRDs are deployed.
+	require.EventuallyWithT(t, func(tt *assert.CollectT) {
+		_, err = dynamicClient.Resource(g).Namespace("default").Create(context.Background(), podMonitor.(*unstructured.Unstructured), metav1.CreateOptions{})
+		if err != nil {
```
pjanotti, line 340: Nit for later: Kent's style :) if err == nil return, same for the update call.

Does it make a difference in the eventually having 2 asserts? I mean do we get the error messages somewhere?

PR #1387, opened 2024-08-15
Add an experimental feature gate to enable exporter batching

dmitryax:
The feature gate enables experimental exporter batching instead of batch processor. This feature ensures the backpressure is propagated to the file readers and no data is dropped. Not recommended to use with enabled gateway.


atoulme:
Can we also add infinite retries for filelog and the exporter?
```
   retry_on_failure:
     max_elapsed_time: 0s
```

dmitryax:
> Can we also add infinite retries for filelog and the exporter?

Done

atoulme (approved):

PR #1389, opened 2024-08-15
[chore] restrict Java app to be deployed on a Linux environment only

atoulme:
Change the deployment of test applications and the operator test environment to only select Linux nodes.


jinja2:
looks like operator also needs it? same selector under the `operator` field https://github.com/signalfx/splunk-otel-collector-chart/blob/main/functional_tests/testdata/values/aks_test_values.yaml.tmpl#L71

crobert-1 (approved):

PR #1403, opened 2024-08-21
[chore] open ports for metrics when testing with kind

atoulme:
**Description:**
Enable metrics ports that will be used for collecting control plane metrics.

### DIFF
--- a/.github/workflows/configs/kind-config.yaml
+++ b/.github/workflows/configs/kind-config.yaml
@@ -29,7 +29,6 @@ nodes:
       - |
         kind: KubeProxyConfiguration
         metricsBindAddress: "0.0.0.0:10249"
-
     extraPortMappings:
       - containerPort: 80
         hostPort: 80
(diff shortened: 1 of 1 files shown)


.github/workflows/configs/kind-config.yaml
```diff
@@ -11,6 +11,25 @@ nodes:
...
+          extraArgs:
+            bind-address: "0.0.0.0"
+        etcd:
+          local:
+            extraArgs:
+              listen-metrics-urls: "http://0.0.0.0:2381"
+      - |
+        kind: KubeProxyConfiguration
+        metricsBindAddress: "0.0.0.0:10249"
+
```
pjanotti, line 32: 😬 there is no such blank like before line 14... (duck for cover...)

pjanotti (approved):

.github/workflows/configs/kind-config.yaml
atoulme:
```suggestion
```

PR #1408, opened 2024-08-22
Remove envvars

atoulme:
**Description:** 
Remove the use of environment variables when configuring hostmetricsreceiver. Use `root_path` instead.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -40,6 +40,7 @@ receivers:
   {{- if (eq (include "splunk-otel-collector.metricsEnabled" .) "true") }}
   hostmetrics:
     collection_interval: 10s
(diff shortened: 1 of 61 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -40,6 +40,7 @@ receivers:
   {{- if (eq (include "splunk-otel-collector.metricsEnabled" .) "true") }}
   hostmetrics:
     collection_interval: 10s
+    root_path: {{ .Values.isWindows | ternary "C:\\hostfs" "/hostfs" }}
```
pjanotti, line 43: I suppose one can't change from the two values above.

pjanotti (approved):

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
atoulme: no they're mounted as such elsewhere.

PR #1425, opened 2024-08-28
Added fsyncEnabled option to the config

sbylica-splunk:
**Description:** Added fsyncEnabled flag, it is passed to file storage used by the agent. This is related to the issue:
https://splunk.atlassian.net/browse/ADDON-71899

With this flag enabled we can avoid issues with data corruption in the future.

**Testing:** Manual testing

**Documentation:** Flag is described in values.yaml

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -9,12 +9,18 @@ extensions:
   {{- if and (eq (include "splunk-otel-collector.logsEnabled" .) "true") (eq .Values.logsEngine "otel") }}
   file_storage:
     directory: {{ .Values.logsCollection.checkpointPath }}
+    {{- if not (eq (toString .Values.splunkPlatform.fsyncEnabled) "<nil>") }}
+    fsync: {{ .Values.splunkPlatform.fsyncEnabled }}
+    {{- end }}
   {{- end }}
 
   {{- if .Values.splunkPlatform.sendingQueue.persistentQueue.enabled }}
   file_storage/persistent_queue:
     directory: {{ .Values.splunkPlatform.sendingQueue.persistentQueue.storagePath }}/agent
     timeout: 0
+    {{- if not (eq (toString .Values.splunkPlatform.fsyncEnabled) "<nil>") }}
+    fsync: {{ .Values.splunkPlatform.fsyncEnabled }}
+    {{- end }}
   {{- end }}
 
 
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -121,6 +121,11 @@ splunkPlatform:
       enabled: false
       storagePath: "/var/addon/splunk/exporter_queue"
 
+  # Option to set fsync value for filestorage extension used by the agent. Enabling this option will ensure
+  # database integrity at the cost of performance. If not set it will use default value for this extension.
+  # Refer to: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage#file-storage
+  # fsyncEnabled: true
+
 ################################################################################
 # Splunk Observability configuration
 ################################################################################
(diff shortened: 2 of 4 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -9,12 +9,14 @@ extensions:
   {{- if and (eq (include "splunk-otel-collector.logsEnabled" .) "true") (eq .Values.logsEngine "otel") }}
   file_storage:
     directory: {{ .Values.logsCollection.checkpointPath }}
+    fsync: {{ .Values.splunkPlatform.fsyncEnabled }}
```
atoulme, line 12: Can we set this option only if it has been set explicitly by the user?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -121,6 +121,11 @@ splunkPlatform:
       enabled: false
       storagePath: "/var/addon/splunk/exporter_queue"
 
+  # Option to disable or enable fsync for filestorage extension used by the agent. Enabling this option will ensure
+  # database integrity at the cost of performance.
+  # Refer to: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage#file-storage
+  fsyncEnabled: false
```
atoulme, line 127: Can we abide by the default from upstream?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -121,6 +121,11 @@ splunkPlatform:
       enabled: false
       storagePath: "/var/addon/splunk/exporter_queue"
 
+  # Option to disable or enable fsync for filestorage extension used by the agent. Enabling this option will ensure
+  # database integrity at the cost of performance.
```
atoulme, line 125: Do you know what the performance cost is by any chance?

sbylica-splunk: Not really, performance is mentioned here:
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage#file-storage

I think that this is strange since our filestorage extension sets fsync to `false` by default (so `nosync` in bbolt is set to `true`), but if we dig more into that, the recommendation is to keep `nosync` flag as `false`. As seen here:
https://pkg.go.dev/go.etcd.io/bbolt#DB

This makes sense as we now know that this flag caused some problems for our customers.
I think that we should expose this flag, keep its value as is (so `fsync` is set to `false`), and maybe do some performance tests to see if we can change it in the future.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
sbylica-splunk: We can put this inside an "if" block, is there any upside to that? Since this flag is already set to `false` by default here:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/storage/filestorage/factory.go

helm-charts/splunk-otel-collector/values.yaml
sbylica-splunk: Yeah we could, see the comment above

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
atoulme: That can change - it will change if I have anything to do with it, and we find that performance is not that impacting.

sbylica-splunk: Ok, changed it to respect the default value. I can see that there is a test failing, but it doesn't seem to be connected to changes here, and other PRs are also failing, so is that ok?

atoulme: it's not ok, please help fix that test :) but for the purpose of this PR, it won't block merging.

helm-charts/splunk-otel-collector/values.yaml
atoulme: OK so 2 follow-ups:
* one issue in contrib to have performance test benchmarks with and without this setting
* one issue in contrib to have this setting changed to the right default per bbolt.

atoulme (approved):

PR #1436, opened 2024-09-04
`Operator`: Migrate .Values.operator.instrumentation.spec to .Values.instrumentation

jvoravong:
- Update Operator Helm values to migrate .Values.operator.instrumentation.spec to .Values.instrumentation
- Add unit tests for instrumentation.yaml Helm template code
- Add migration notes


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1225,6 +1171,87 @@ operator:
         "helm.sh/hook": post-install,post-upgrade
         "helm.sh/hook-weight": "1"
 
+# The default Splunk Instrumentation object deployed when operator.enabled=true.
+# For more details see:
+# - Splunk Documentation: https://docs.splunk.com/observability/en/gdi/opentelemetry/automatic-discovery/k8s/k8s-backend.html#optional-configure-the-instrumentation
+# - OpenTelemetry Documentation: https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api.md#instrumentation
+instrumentation:
```
dmitryax, line 1178: Let's add a note that this section is only applicable when `operator.enabled=true`?

dmitryax (approved):
LGTM

helm-charts/splunk-otel-collector/values.yaml
jvoravong: A note already exists on 1174, but we can emphasize it and improve configurability by introducing a YAML anchor too. Added `instrumentation.enabled` with an anchored value, allowing for more flexible control over the default Splunk Instrumentation deployment.

dmitryax: Sorry I missed the note. I don't like anchoring. Is it two ways or operator.enabled is the only source of truth? This This confuses me as a user. Let;s not add it

jvoravong: I'm fine with that

PR #1438, opened 2024-09-09
Add support for writing out new expected results files via env var

jvoravong:
- Make it so new expected test results are generate under the `functional_tests/results` directory and in the golden file format
- Use the `UPDATE_EXPECTED_RESULTS` env var to determine if new expected test results are generated, the env var should default to false.

### DIFF
--- a/.github/workflows/functional_test_v2.yaml
+++ b/.github/workflows/functional_test_v2.yaml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
     inputs:
       UPDATE_EXPECTED_RESULTS:
-        description: 'Set this to true to update expected results and collect new test output.'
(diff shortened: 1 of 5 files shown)


.github/workflows/functional_test_v2.yaml
```diff
@@ -62,14 +68,15 @@ jobs:
       - name: Deploy cert-manager
         run: |
           make cert-manager
-      - name: run functional tests
+      - name: Run functional tests
```
atoulme, line 71: unrelated change?

jvoravong: reverted

atoulme (approved):
LGTM

samiura (approved):

PR #1462, opened 2024-09-20
[agent] use root_path 

atoulme:
**Description:**
use root_path to configure the hostmetricsreceiver, instead of environment variables.


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -49,6 +49,7 @@ receivers:
   {{- if (eq (include "splunk-otel-collector.metricsEnabled" .) "true") }}
   hostmetrics:
     collection_interval: 10s
+    root_path: {{ .Values.isWindows | ternary "C:\\hostfs" "/hostfs" }}
```
pjanotti, line 52: Q.: is it safe to assume that is always installed on the `C:` drive? Sorry, I had to ask...

pjanotti (commented):
LGTM - with a Q

pjanotti (approved):

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
atoulme: it was set like this. I don't know any better.

pjanotti: oh, if I am understanding this right you must set it to empty on anything other than Linux, see https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/0f2fea60e36e9951a3caf4884a8264473a8204c5/receiver/hostmetricsreceiver/hostmetrics_others.go#L14

pjanotti (changes requested):
See https://github.com/signalfx/splunk-otel-collector-chart/pull/1462/files#r1777855573

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
atoulme: good catch, I'll fix.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -49,7 +49,9 @@ receivers:
   {{- if (eq (include "splunk-otel-collector.metricsEnabled" .) "true") }}
   hostmetrics:
     collection_interval: 10s
-    root_path: {{ .Values.isWindows | ternary "C:\\hostfs" "/hostfs" }}
+    {{- if not .Values.isWindows }}
```
pjanotti, line 52: This will work assumes that all images are either Linux or Windows - I *think* that is true. The ideal fix would be `if .Values.isLinux` but I don't know if that is a possibility.

pjanotti (approved):

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
atoulme: yes so far only linux and windows are supported

PR #1481, opened 2024-10-01
Ingest cluster receiver events into index defined in annotation

pszkamruk-splunk:
**Description:** 
Ingest cluster receiver events into index defined in annotation

**Link to Splunk idea:**

**Testing:** Tested manually 

**Documentation:**


functional_tests/testdata_histogram/expected/v1.29/controller_manager_metrics.yaml
```diff
@@ -4883,4 +4883,4 @@ resourceMetrics:
             name: node_ipam_controller_cidrset_allocation_tries_per_request
         scope:
           name: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver
-          version: v0.109.0
+          version: v0.110.0
```
jvoravong, line 4886: https://github.com/signalfx/splunk-otel-collector-chart/pull/1482 should make it so you don't have to updates these [controller_manager_metrics.yaml](https://github.com/signalfx/splunk-otel-collector-chart/pull/1481/files#diff-51ebe0bb5fea8cc6a7fda3c168b0c4eb19bc5f63625f5207cd88767309aa3116) and [coredns_metrics.yaml](https://github.com/signalfx/splunk-otel-collector-chart/pull/1481/files#diff-8a6dcdfda9b9238fbb81a49821b5932fc9b0398d7f2938f47e39982b2cd30e08) tests

pszkamruk-splunk: OK, got it, I added it initially as pipeline was failing here 
I removed that change from PR

jvoravong: Thank you

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -209,6 +209,34 @@ k8sattributes:
...
+  extract:
+    metadata:
+      - k8s.namespace.name
+      - k8s.node.name
+      - k8s.pod.name
+      - k8s.pod.uid
+      - container.id
+      - container.image.name
+      - container.image.tag
+    annotations:
```
jvoravong, line 230: We should put an if clause here and only add these annotations if Splunk Platform/Enterprise is the backend.

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -73,6 +73,7 @@ receivers:
   {{- end }}
 
 processors:
+  {{- include "splunk-otel-collector.k8sClusterReceiverAttributesProcessor" . | nindent 2 }}
```
jvoravong, line 76: Lets refactor this a little for a couple reasons. This configurations will always be included even if not used and we could relocate it to be closer to similar configurations like it.
- Move this content to around lines 92-112
- Add conditional clauses appropriately to make this configuration content only included when it is used. Something like this.

```
  {{- if and $clusterReceiver.eventsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
  {{- include "splunk-otel-collector.k8sClusterReceiverAttributesProcessor" . | nindent 2 }}
  {{- end }}`
```

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -209,6 +209,34 @@ k8sattributes:
       {{- include "splunk-otel-collector.addExtraLabels" . | nindent 6 }}
     {{- end }}
 {{- end }}
+{{- define "splunk-otel-collector.k8sClusterReceiverAttributesProcessor" -}}
+k8sattributes/clusterReceiver:
```
jvoravong, line 213: If we're adding this change, we need to address some cleanup to prevent code duplication related to setting Kubernetes attributes:
- Remove redundant attribute setting: We should clean up any attributes set using the previous method that will now be handled by the k8sattributes processor.
  - [Reference line in PR](https://github.com/signalfx/splunk-otel-collector-chart/pull/1481/files#diff-069710df8ca3984808067088bb09303468c24d7e97e4d8970234e703ef52ffd2R121)
- Clean up unused environment variables: The environment variables in the cluster receiver deployment that are no longer used to set attributes the old way should also be removed.
  - [Env var reference 1](https://github.com/signalfx/splunk-otel-collector-chart/blob/e394d52af4f7a065952737a3bcc9bb0b6b94757b/helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml#L94)
  - [Env var reference 2](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml#L132)

pszkamruk-splunk: I removed variables - changes are on separate branch: [here](https://github.com/pszkamruk-splunk/splunk-otel-collector-chart/commit/556fad616e152c214c1fb7f7ff2604ba7752e3fd)
I checked that, after removing env variables from configuration, attributes are still being added to events using k8sAttributesProcessor.
There is another [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35528) in contrib repo which fixes some logs which are not matched by k8sAttributesProcessor. 
In case of these missing/not matched events, env variables are causing that these events have also additional attributes like k8s.pod.name, etc

I noticed also that these env variables are also used to add attributes to metrics by `resource/add_collector_k8s`: [link](https://github.com/signalfx/splunk-otel-collector-chart/blob/7ea0cab65fd0825b61e162e405ae11bedcf06890/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl#L121) 
There are 3 metric pipelines and one of them (`metrics/collector` pipeline) is using these env variables so I am not sure if we can remove these variables.
But on the other hand all metric pipelines are also using k8sAttributesProcessor, and this also duplicates the functionality of adding attributes to metric pipeline. 

So regarding all above, my questions:
1. If the changes on separate branch are OK, I will merge them here 
2. Should I also remove `resource/add_collector_k8s` which is used by metric pipeline?

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
pszkamruk-splunk: done

helm-charts/splunk-otel-collector/templates/config/_common.tpl
pszkamruk-splunk: Let's move this change of removing environment variables to separate PR.
Before removing environment variables, we first need to merge contrinb [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35528) - adding namespace attribute to k8s events allowing k8sAttributesProcessor to match all k8s events. Currently K8sAttributesProcessor does not matching all k8s events.
We need also decide what to do with  `resource/add_collector_k8s`: [link](https://github.com/signalfx/splunk-otel-collector-chart/blob/7ea0cab65fd0825b61e162e405ae11bedcf06890/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl#L121) 
as this is using env variables which we want to remove, but the functionality of adding additional parameters is covered by `k8sattributes/metrics`, so probably it is safe to remove `resource/add_collector_k8s` from configuration

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -73,6 +73,7 @@ receivers:
   {{- end }}
 
 processors:
+{{/*  {{- include "splunk-otel-collector.k8sClusterReceiverAttributesProcessor" . | nindent 2 }}*/}}
```
jvoravong, line 76:
```suggestion
{{/* {{- include "splunk-otel-collector.k8sClusterReceiverAttributesProcessor" . | nindent 2 }} */}}
```

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -211,6 +211,34 @@ k8sattributes:
...
+  extract:
+    metadata:
+      - k8s.namespace.name
+      - k8s.node.name
+      - k8s.pod.name
+      - k8s.pod.uid
+      - container.id
+      - container.image.name
+      - container.image.tag
+    annotations:
```
jvoravong, line 232: Can you add another if statement to only include these annotations configurations if `splunk-otel-collector.splunkPlatformEnabled=true` since Splunk Observability cannot use these?

helm-charts/splunk-otel-collector/templates/config/_common.tpl
jvoravong: - Environment Variables
  - Remove the environment variables in a separate PR after the contrib fix is merged.
- resource/add_collector_k8s:
  - Replace this with a k8sattributes processor.
  - Aim to use a single instance of the k8sattributes processor if possible. If required, add a second instance (e.g., k8sattributes/add_collector) for handling collector metrics specifically.
  - Functional Tests cover internal collector metrics and attributes (e.g., [otelcol_exporter_send_failed_spans](https://github.com/signalfx/splunk-otel-collector-chart/blob/0e083695672ef3055a008964f6c33675dea236f4/functional_tests/testdata/expected_kind_values/expected_internal_metrics.yaml#L5044)). Check to see if you can get the tests to pass with using only a single k8sattributes processor instance. Add a second processor instance (named something like `k8sattributes/add_collector`) because the tests failed and it is necessary.

helm-charts/splunk-otel-collector/templates/config/_common.tpl
pszkamruk-splunk: done

jvoravong (approved):

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -117,6 +117,10 @@ processors:
           - set(resource.attributes["com.splunk.sourcetype"], Concat(["kube:object:", attributes["k8s.resource.name"]], ""))
   {{- end }}
 
+  {{- if and $clusterReceiver.eventsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
```
jvoravong, line 120: We actually need to update this if statement, there are cases where k8sattributes/clusterReceiver processor is not included but it should be because it is [used in pipelines](https://github.com/signalfx/splunk-otel-collector-chart/pull/1481/files/7345d03bcc7b760cfb512a068061e26f1111475f#diff-4e951abec9da3c5944c771fee7bd203aab089a25f1ea23d2b86fadc0590d7074R133).

jvoravong: There might be cleaner ways of doing this, but I think we will at least something here that looks like this. See what you can do.
```
{{- if or 
  (and $clusterReceiver.eventsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true")) 
  (and (eq (include "splunk-otel-collector.objectsEnabled" .) "true") (eq (include "splunk-otel-collector.logsEnabled" .) "true")) 
  (eq (include "splunk-otel-collector.o11yInfraMonEventsEnabled" .) "true") 
}}
```

pszkamruk-splunk: or we can introduce sth like this:
`{{- $splunkLogsEnabled := eq (include "splunk-otel-collector.logsEnabled" .) "true" -}}`
`{{- $splunkObjectsEnabled := eq (include "splunk-otel-collector.objectsEnabled" .) "true" -}}`
`{{- $o11yInfraMonEventsEnabled := eq (include "splunk-otel-collector.o11yInfraMonEventsEnabled" .) "true" -}}`

`{{- if or 
  (and $clusterReceiver.eventsEnabled $splunkLogsEnabled) 
  (and $splunkObjectsEnabled $splunkLogsEnabled) 
  $o11yInfraMonEventsEnabled 
}}`

but here we will need to refactor the whole file and this will introduce even more variables

jvoravong: I think the current state of this PR LGTM. Let's try to get this merged in and you can do more refactoring in separate PRs as you see fit. I'll merge this work shortly.

jvoravong (approved):

PR #1485, opened 2024-10-03
Update the chart - collector agent can use a K8s service by default

jvoravong:
This PR updates the default values.yaml to enable the agent service by default, allowing customers using Kubernetes to send data via a Kubernetes service instead of relying on host ports. This change addresses customer requests for better security and alignment with modern Kubernetes practices, while still maintaining the use of the host network for now.

Changes:
- Default Change: agent.service.enabled is now set to true to enable the agent service by default.
- The agent will continue to use the host network as before, ensuring compatibility while we plan future efforts to migrate away from the host network (not in scope for this PR).


atoulme (approved):

atoulme:
Is there a CHANGELOG entry to add for this?

jvoravong:
> Is there a CHANGELOG entry to add for this?

Changelog entry added. 

Just a heads-up: our CI/CD tests for the changelog only check if chart templates were updated and whether a corresponding changelog entry was added. Changes to default values in values.yaml won't trigger the tests.

PR #1493, opened 2024-10-08
Testing: Adding splunk matrix

pszkamruk-splunk:
**Description:** 
Adding Splunk matrix: 
- updating currently used Splunk 8.x to the latest 8.x relase
- adding Splunk 9.x - latest available release

+ parametrize of `hec` and `splunk password` in k8s-splunk.yaml

### DIFF
--- a/ci_scripts/k8s-splunk.yml
+++ b/ci_scripts/k8s-splunk.yml
@@ -60,7 +60,7 @@ spec:
     runAsGroup: 0
   containers:
     - name: splunk
-      image: docker.io/splunk/splunk:SPLUNK_VERSION
+      image: docker.io/splunk/splunk:ENV_SPLUNK_VERSION
       ports:
         - name: web-interface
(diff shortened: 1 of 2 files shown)


ci_scripts/k8s-splunk.yml
```diff
@@ -77,11 +77,11 @@ spec:
         - name: SPLUNK_USER
           value: root
         - name: SPLUNK_PASSWORD
-          value: helloworld
+          value: ENV_CI_SPLUNK_PASSWORD
```
jvoravong, line 80: Can you see if you can set a default value if these default env vars don't exist? It makes it easier to run these tests locally for already created methods.

pszkamruk-splunk: Done,
We are now matching using default values instead of parameters names which I introduced in previous commits

jvoravong (approved):

PR #1533, opened 2024-11-08
Update hostmetrics receiver config to accommodate recent changes

jvoravong:
- Fix bug where hostmetrics receiver was failing to scrape the filesystem
- Was able to test this fix with minikube


aquam8:
I think you're trying to fix that problem that i am experimenting using 0.112.0
```
error scraping metrics "failed to read usage at /hostfs/conf: no such file or directory
```

Removing the path also seems to make the problem go away, thank you!

swahba (approved):

PR #1559, opened 2024-12-05
OTL-3190 Additional fixes for operator install order issues using Helm hooks

jvoravong:
**Description**  
Helm install failures occur in rare cases due to deployment order issues:  
- **Instrumentation Object:** May fail because the Operator and its webhook are not ready.  
- **Cert-Manager Certificate:** May fail because Cert-Manager is not yet initialized.  

**Proposed Solution**  
Add Helm hooks to enforce readiness checks and ensure resources are deployed in the correct order. Cert-Manager and Operator readiness will be validated before Instrumentation deployment. We will use appropriate Hook weight values to ensure each related operation is completed sequentially.

**Installation Order**  
1. **Default Installation Phase:** Deploy Cert-Manager, Operator, and Operator CRDs.  
2. **Cert-Manager Readiness Check:** Post-install hook (weight 1) verifies Cert-Manager is operational.  
3. **Install Issuer:** Post-install hook (weight 2) deploys the Cert-Manager Issuer.  
4. **Install Certificate:** Post-install hook (weight 3) deploys the Cert-Manager Certificate.  
5. **Operator Readiness Checks:** Post-install hooks (weight 4) validate the Operator and related webhook are ready.  
6. **Install Instrumentation:** Post-install hook (weight 5) deploys Instrumentation resources.  

**Benefits**  
- Ensures dependencies are deployed in the correct order.  
- Mitigates Helm installation errors reported from the field.  

**Cons**  
- Using Helm hooks like this and this much is an anti-pattern for Helm.

**Callouts**  
- **Future Refactoring:** We may refactor Cert-Manager usage and CRD installation in the near future, potentially eliminating the need for post-install hooks. While this PR provides an effective interim solution, it may be short-lived within this project.  

**Testing**  
- Verified with local and lab Kubernetes clusters. 

**Related Alernative solution:** https://github.com/signalfx/splunk-otel-collector-chart/pull/1561

PR #1561, opened 2024-12-06
OTL-3190 Move operator CRD installation to the crds/ folder via a subchart to resolve Helm install ordering issues

jvoravong:
Helm install failures occur in rare cases due to deployment order issues:  
- **Instrumentation Object:** May fail because the Operator and its webhook are not ready.  
- **Cert-Manager Certificate:** May fail because Cert-Manager is not yet initialized.  

**Proposed Solution** 
- We move needed (Instrumentation) CRD installation into a new local subchart opentelemetry-operator-subchart which is only used for Helm chart installation. 
  - Then we can stop using a post-install Helm hook to install the [instrumentation.yaml](https://github.com/signalfx/splunk-otel-collector-chart/blob/44d8c42316ad52800a26145370d44dee4dbc178a/examples/enable-operator-and-auto-instrumentation/rendered_manifests/operator/instrumentation.yaml#L21).
  - Other OpenTelemetry charts ([opentelemetry-kube-stack chart](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-kube-stack)) have also taken this approach.

**Testing:**
TBD

**Related Alernative solution:** https://github.com/signalfx/splunk-otel-collector-chart/pull/1559


atoulme:
I have tried this fix, but the CI doesn't pass ; same problem with the operator CRDs not being installed early enough.

PR #1562, opened 2024-12-09
Make it so the default tolerations used to deploy the agent collector account for k8s distribution

jvoravong:
**Description:**  
- Added default tolerations to monitor control-plane, worker, and distribution-specific node pools with the agent collector based on the Kubernetes distribution:
  - OpenShift: infrastructure node pool
  - AKS: system node pool  
- Previously, OpenShift and AKS users needed to configure these tolerations manually. Based on field feedback, this change automates the configuration, simplifying setup and ensuring out-of-the-box support for unique node pools.

### DIFF
--- a/.chloggen/enhance-agent-daemonset-toleration.yaml
+++ b/.chloggen/enhance-agent-daemonset-toleration.yaml
@@ -1,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Make it so the default tolerations used to deploy the agent collector account for k8s distribution
+# One or more tracking issues related to the change
+issues: [1562]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: OpenShift infra nodes and Azure system nodes will now be monitored by the agent by default
--- a/.chloggen/enhance-agent-daemonset-toleration.yaml
+++ b/.chloggen/enhance-agent-daemonset-toleration.yaml
@@ -9,4 +9,4 @@ issues: [1562]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext: OpenShift infra nodes and Azure system nodes will now be monitored by the agent by default
+subtext: OpenShift infra nodes and AKS system nodes will now be monitored by the agent by default
--- a/helm-charts/splunk-otel-collector/templates/daemonset.yaml
+++ b/helm-charts/splunk-otel-collector/templates/daemonset.yaml
@@ -80,10 +80,10 @@ spec:
       {{- else }}
         - key: node-role.kubernetes.io/master
           effect: NoSchedule
-          operator: Equal
+          operator: Exists
         - key: node-role.kubernetes.io/control-plane
           effect: NoSchedule
-          operator: Equal
+          operator: Exists
         {{- if eq .Values.distribution "aks" }}
         - key: kubernetes.io/system-node
           effect: NoSchedule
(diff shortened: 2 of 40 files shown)


helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -74,9 +74,24 @@ spec:
...
+        {{ toYaml .Values.tolerations | nindent 8 }}
+      {{- else }}
+        - key: node-role.kubernetes.io/master
+          effect: NoSchedule
+        - key: node-role.kubernetes.io/control-plane
+          effect: NoSchedule
+        {{- if eq .Values.distribution "openshift" }}
+        - key: node-role.kubernetes.io/infra
+          effect: NoSchedule
+          operator: Exists
```
jinja2, line 88: Small nit - can we make the tolerations consistent? The other 2 tolerations before do not have an explicit `operator` field

jvoravong: Sure. Done.

.chloggen/enhance-agent-daemonset-toleration.yaml
```diff
@@ -0,0 +1,12 @@
...
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Make it so the default tolerations used to deploy the agent collector account for k8s distribution
+# One or more tracking issues related to the change
+issues: [1562]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: OpenShift infra nodes and Azure system nodes will now be monitored by the agent by default
```
atoulme, line 12: is this correct? It appears we explicitly set to not schedule by default on openshift and azure.
https://github.com/signalfx/splunk-otel-collector-chart/pull/1562/files#diff-a196d36724bc1a2c08c5c54b9c3efe8763e29bb0c007995455b93ab280710c8fR87

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -74,9 +74,26 @@ spec:
...
       {{- end }}
-      {{- with .Values.tolerations }}
       tolerations:
-        {{ toYaml . | nindent 8 }}
+      {{- if and (.Values.tolerations) (ne (len .Values.tolerations) 0) }}
+        {{ toYaml .Values.tolerations | nindent 8 }}
+      {{- else }}
+        - key: node-role.kubernetes.io/master
+          effect: NoSchedule
+          operator: Equal
```
jinja2, line 83:
```suggestion
          operator: Exists
```

From the docs 
```
The default value for operator is Equal.

A toleration "matches" a taint if the keys are the same and the effects are the same, and:
the operator is Exists (in which case no value should be specified), or
the operator is Equal and the values should be equal.
```
Since we are not specifying a value to match against, `Exists` seems to be the correct operator type here. Weird though that this has been working w/o the value.

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -74,9 +74,26 @@ spec:
...
+        - key: node-role.kubernetes.io/control-plane
+          effect: NoSchedule
+          operator: Equal
```
jinja2, line 86:
```suggestion
          operator: Exists
```

.chloggen/enhance-agent-daemonset-toleration.yaml
jinja2: The changelog is correct. Tolerations on pods work with taints on nodes. K8s will schedule a pod on a node with a taint like `key=value:NoSchedule` only if the pod tolerates it. That's what Josh's change does, it tells k8s that the agent pod tolerate these node taints and to schedule it anyway.

atoulme: ok thanks!

helm-charts/splunk-otel-collector/templates/daemonset.yaml
jvoravong: For deployed nodes the standard value for these default tolerations is an empty string which is why the default `Equal` operator works. I have no strong opinion here to prefer one of the `Equal` and `Exists` operators here, I'll add the requested changes.

jinja2 (approved):

PR #1571, opened 2024-12-12
Add preserveLeadingWhitespaces and preserveTrailingWhitespaces variables under logsCollection.containers in values.yaml

wojtekzyla:
**Description:**  Allow users to control whether filelog receiver should remove leading and trailing whitespaces from their logs. We noticed a case when one log message is spread across several logs and then joined using [recombine operator](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/recombine.md), the default behaviour of filelog receiver (removing leading and trailing whitespaces) resulted in malformed combined log. 

**Testing:** Manual testing

**Documentation:**

### DIFF
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -621,6 +621,11 @@ logsCollection:
     # Set to 0 to remove any size limit.
     maxRecombineLogSize: 1048576
 
+    # Whether to preserve leading whitespaces in logs.
+    preserveLeadingWhitespaces: false
+    # Whether to preserve trailing whitespaces in logs.
+    preserveTrailingWhitespaces: false
+
 # Configuration for collecting journald logs using otel collector
   journald:
     enabled: false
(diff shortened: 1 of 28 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -621,6 +621,11 @@ logsCollection:
     # Set to 0 to remove any size limit.
     maxRecombineLogSize: 1048576
 
+    # Whether to preserve leading whitespaces in logs.
+    preserveLeadingWhitespaces: false
+    # Whether to preserve trailing whitespaces in logs.
+    preserveTrailingWhitespaces: false
```
dmitryax, line 627: Why do we need extra parameters when these are already available with:
```
agent.config.receivers.filelog.preserve_leading_whitespaces
agent.config.receivers.filelog.preserve_trailing_whitespaces
```

Let's not extend the values.yaml interface when it is not necessary

dmitryax (changes requested):

helm-charts/splunk-otel-collector/values.yaml
wojtekzyla: That's right, I missed the part with that configuration. I'll close this PR

PR #1573, opened 2024-12-13
Update config for setting up internal metrics scraper

dmitryax:
Update config for scraping internal metrics to use new config interface and loopback address. This also drops redundant attributes reported with the internal metrics: `net.host.name` and `server.address`.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -847,7 +847,12 @@ exporters:
 service:
   telemetry:
     metrics:
-      address: 0.0.0.0:8889
+      readers:
+        - pull:
+            exporter:
+              prometheus:
+                host: localhost
+                port: 8889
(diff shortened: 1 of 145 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -847,7 +847,12 @@ exporters:
 service:
   telemetry:
     metrics:
-      address: 0.0.0.0:8889
+      readers:
```
jinja2, line 850: Is this new config interface also default for the version we are on? I think 8889 is our custom port so we probably have to add this anyway but just curious

jinja2 (approved):

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: Yeah, that's right. Default is the same, just different 8888 port

PR #1600, opened 2025-01-09
Do not install CRDs as part of the helm chart templates.

atoulme:
This changes the behavior of the chart, as a breaking change, asking users to install CRDs by themselves as helm does not manage CRDs directly well.


atoulme:
Closing in favor of #1604

PR #1604, opened 2025-01-10
Use a subchart to install CRDs via helm

atoulme:
**Description:**
Reprising #1561.
This PR has the following changes:
* Rebased off main
* Remove the test code that installs CRDs, use the subchart instead.


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1191,6 +1200,11 @@ operator:
     collectorImage:
       repository: quay.io/signalfx/splunk-otel-collector
 
+# Should the CRDs be installed by a chart's crd/ directory or by using templates.
+operatorcrds:
+  # Should the CRDs be installed
+  install: true
```
atoulme, line 1206: ah, maybe the default needs to be false unless operator is also enabled.

helm-charts/splunk-otel-collector/Chart.yaml
```diff
@@ -20,13 +20,18 @@ maintainers:
...
-  # - Avoid hyphen characters in names/aliases, they introduce template rendering complications (https://github.com/helm/helm/issues/2192)
+  # - Avoid uppercase letters in aliases, they cause install failure due to subchart resource naming
+  # - Avoid hyphen characters in aliases, they introduce template rendering complications (https://github.com/helm/helm/issues/2192)
   - name: cert-manager
     version: v1.14.4
     alias: certmanager
     repository: https://charts.jetstack.io
     condition: certmanager.enabled
+  - name: opentelemetry-operator-crds
+    version: "0.0.0"
```
dmitryax, line 31: Shouldn't be a valid version? 0.0.1?

helm-charts/splunk-otel-collector/values.yaml
dmitryax: Right. Please ensure that no crds are installed by default (auto-instrumentation disabled).

dmitryax (approved):

PR #1616, opened 2025-01-14
[chore] fix GCE clusterName deployment

atoulme:


atoulme:
The fix is valid, but GCE fails later because of the operator overwhelming the API server.

atoulme:
This is now ready for merge.

jinja2 (approved):

PR #1617, opened 2025-01-14
[chore] add a test checking upgrades from a previous release

atoulme:


.github/workflows/functional_test_v2.yaml
```diff
@@ -224,6 +224,78 @@ jobs:
...
+          aws eks update-kubeconfig --name rotel-eks --region us-west-1
+      - name: Update dependencies
+        run: |
+          cd base && make dep-update
+      - name: Deploy cert-manager
+        run: |
+          cd base && make cert-manager
+      - name: Deploy previous version of the chart
+        run: |
+          cd base && helm install sock helm-charts/splunk-otel-collector --set cloudProvider=aws --set distribution=eks --set splunkObservability.realm=us0 --set splunkObservability.accessToken=xxxxx
```
jinja2, line 286:
```suggestion
          cd base && helm install sock helm-charts/splunk-otel-collector --set cloudProvider=aws --set distribution=eks --set splunkObservability.realm=us0 --set splunkObservability.accessToken=xxxxx --set operator.enabled=true --set certmanager.enabled=true
```

jinja2: The default has operator disabled in last version https://github.com/signalfx/splunk-otel-collector-chart/blob/splunk-otel-collector-0.113.0/helm-charts/splunk-otel-collector/values.yaml#L1168. If the goal is to test for operator/CRD changes are no-op for customers running with operator, then we should the extra values ^

pszkamruk-splunk (approved):

jinja2 (approved):

jvoravong (approved):

PR #1601, opened 2025-01-10
[chore] fix test expectations

atoulme:
Fix expectations - functional tests were not running and fell behind on expectations from testing.

PR #1618, opened 2025-01-15
test upgrade with no change in values

jinja2:


atoulme:
The chart fails to upgrade:
```
Error: UPGRADE FAILED: Unable to continue with update: Instrumentation "sock-splunk-otel-collector" in namespace "default" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "sock"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "default"
```

jinja2:
So upgrade fails due to [this](https://github.com/signalfx/splunk-otel-collector-chart/pull/1601/files) change introduced in latest which makes instrumentation a non-hook object. On upgrade helm sees the object exists but it is not part of tracked objects in its release, hence the error about missing annotations `"meta.helm.sh/*"`

.github/workflows/functional_test_v2.yaml
```diff
@@ -488,3 +488,270 @@ jobs:
...
+      - name: check helm release status
+        run: |
+          helm status sock
+      - name: check instrumentations installed
+        run: |
+          kubectl describe instrumentations.opentelemetry.io -A
+      - name: Update dependencies
+        run: |
+          make dep-update
+      - name: try to import instrumentation in helm
```
jinja2, line 566: required cmds to get around the instrumentation hook to non-hook change on upgrade

.github/workflows/functional_test_v2.yaml
```diff
@@ -488,3 +488,270 @@ jobs:
...
+        run: |
+          NAMESPACE=default
+          RELEASE_NAME=sock
+          kubectl get -n $NAMESPACE instrumentations.opentelemetry.io -o name | xargs -I % kubectl annotate -n $NAMESPACE % meta.helm.sh/release-name=$RELEASE_NAME
+          kubectl get -n $NAMESPACE instrumentations.opentelemetry.io -o name | xargs -I % kubectl annotate -n $NAMESPACE % meta.helm.sh/release-namespace=$NAMESPACE
+      - name: upgrade helm chart
+        run: |
+          helm upgrade sock helm-charts/splunk-otel-collector --debug \
+          --set splunkObservability.realm=us0 --set splunkObservability.accessToken=xxxxx \
+          --set operator.enabled=true --set operator.crds.create=true --set environment=dev --set clusterName=dev
```
jinja2, line 665: `--set operator.crds.create=true` is the extra value to be set

atoulme: Could we also recommend to use `--reuse-values` here? https://helm.sh/docs/helm/helm_upgrade/

jinja2: no, I tried it in one of my earlier runs. The operator schema seems to have an update between the 2 versions which makes it fail validation with the option. I think it will also be a problem with the new values vars we have introduced since last version

jinja2:
I think I have tried all combinations of possible ways an existing user with operator enabled could try to helm upgrade. The only option which seems to work where upgrade does not fail and the crds are kept is where they explicitly add the var `operator.crds.create=true` and some fix for the instrumentation object. For the latter, let me see if i can conditionally introduce the hook annotations back if crds.create is true and that solves the issue for new installs with subchart and existing operator templates.

PR #1642, opened 2025-02-03
[chart] use security context under container level to allow setting additional permissions

atoulme:
**Description:**
Use security context under container level to allow setting additional permissions
Set defaults that heighten security.

### DIFF
--- a/.chloggen/movesecuritycontexttocontainers.yaml
+++ b/.chloggen/movesecuritycontexttocontainers.yaml
@@ -1,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: chart
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: use security context under container level to allow setting additional permissions
+# One or more tracking issues related to the change
+issues: [1642]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/examples/add-filter-processor/rendered_manifests/deployment-cluster-receiver.yaml
+++ b/examples/add-filter-processor/rendered_manifests/deployment-cluster-receiver.yaml
@@ -43,6 +43,15 @@ spec:
         - --config=/conf/relay.yaml
         image: quay.io/signalfx/splunk-otel-collector:0.117.0
         imagePullPolicy: IfNotPresent
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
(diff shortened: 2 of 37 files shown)


jvoravong:
We need to make a few changes before this PR is ready to merge.
- We need to be able to update the security context for the agent, cluster receiver, and gateway
- Right now the security context value changes only apply to the pod level, we need to support setting a security context at the pod level and container level. These two different levels have different options for configurations.
- For path of least resistance and ease of support, I might suggest supporting the current values and creating values like this. This is a specific example for the gateway and in the end result we need to support all 3 agent modes.
  - `gateway.securityContext`: Currently exists and sets security context on pods. This is kept like this to support previous customer usage.
  - `gateway.securityContext.pod`: Would be a new value that sets security context on the [gateway pods](https://github.com/signalfx/splunk-otel-collector-chart/blob/0198bd5e74be66ad5ad553ada9097fced66362a6/examples/collector-all-modes/rendered_manifests/deployment-gateway.yaml#L27). Similar to `gateway.securityContext`, but would error out if `gateway.securityContext` and `gateway.securityContext.pod` are both used.
  - `gateway.securityContext.containers.otel-collector`: Would be a new value that sets security context on the ["otel-collector" container deployed by the gateway pod](https://github.com/signalfx/splunk-otel-collector-chart/blob/0198bd5e74be66ad5ad553ada9097fced66362a6/examples/collector-all-modes/rendered_manifests/deployment-gateway.yaml#L40).

atoulme:
> We need to make a few changes before this PR is ready to merge.
> 
> * We need to be able to update the security context for the agent, cluster receiver, and gateway
> * Right now the security context value changes only apply to the pod level, we need to support setting a security context at the pod level and container level. These two different levels have different options for configurations.

It seemed to me that container level securityContext options are inclusive of pod options. Is that not the case?

> * For path of least resistance and ease of support, I might suggest supporting the current values and creating values like this. This is a specific example for the gateway and in the end result we need to support all 3 agent modes.
>   
>   * `gateway.securityContext`: Currently exists and sets security context on pods. This is kept like this to support previous customer usage.
>   * `gateway.securityContext.pod`: Would be a new value that sets security context on the [gateway pods](https://github.com/signalfx/splunk-otel-collector-chart/blob/0198bd5e74be66ad5ad553ada9097fced66362a6/examples/collector-all-modes/rendered_manifests/deployment-gateway.yaml#L27). Similar to `gateway.securityContext`, but would error out if `gateway.securityContext` and `gateway.securityContext.pod` are both used.

The current setup is securityContext is set to object without any validation, so this is likely to be a breaking change.

>   * `gateway.securityContext.containers.otel-collector`: Would be a new value that sets security context on the ["otel-collector" container deployed by the gateway pod](https://github.com/signalfx/splunk-otel-collector-chart/blob/0198bd5e74be66ad5ad553ada9097fced66362a6/examples/collector-all-modes/rendered_manifests/deployment-gateway.yaml#L40).

As above ; the gateway has one container as well, is it worth setting up permissions by naming the container?

atoulme:
Ah never mind, it's not inclusive. I will close this PR and offer that we apply the container security context under its own object.

PR #1647, opened 2025-02-05
Extend options for setting security context

dmitryax:
Add option to set security contexts on:
- container in the clusterRecever deployment
- container in the gateway deployment
- agent pod

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/NOTES.txt
+++ b/helm-charts/splunk-otel-collector/templates/NOTES.txt
@@ -81,3 +81,9 @@ Splunk OpenTelemetry Collector is installed and configured to send data to Splun
   - Status: Instrumentation language maturity varies. See `operator.instrumentation.spec` and documentation for utilized instrumentation details.
   - Splunk Support: We offer full support for Splunk distributions and best-effort support for native OpenTelemetry distributions of auto-instrumentation libraries.
 {{- end }}
+{{- if not (eq (toString .Values.gateway.securityContext) "<nil>") }}
+[WARNING] "gateway.securityContext" parameter is deprecated. Please use "gateway.podSecurityContext" instead.
+{{ end }}
+{{- if not (eq (toString .Values.clusterReceiver.securityContext) "<nil>") }}
+[WARNING] "clusterReceiver.securityContext" parameter is deprecated. Please use "clusterReceiver.podSecurityContext" instead.
+{{ end }}
--- a/helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
+++ b/helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
@@ -81,9 +81,10 @@ spec:
         ` }}
         {{- $clusterReceiver.affinity | mustMergeOverwrite (fromYaml $clusterReceiverPodAntiAffinity) | toYaml | nindent 8 }}
       {{- end }}
-      {{- if $clusterReceiver.securityContext }}
+      {{- $podSecurityContext := $clusterReceiver.podSecurityContext | default $clusterReceiver.securityContext }}
+      {{- if $podSecurityContext }}
       securityContext:
-        {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $clusterReceiver.securityContext) | nindent 8 }}
+        {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $podSecurityContext) | nindent 8 }}
       {{- end }}
       {{- if eq (include "splunk-otel-collector.distribution" .) "eks/fargate" }}
       initContainers:
@@ -126,6 +127,10 @@ spec:
         {{- end }}
         {{- if .Values.clusterReceiver.featureGates }}
         - --feature-gates={{ .Values.clusterReceiver.featureGates }}
+        {{- end }}
+        {{- with .Values.clusterReceiver.containerSecurityContext }}
+        securityContext:
+          {{- toYaml . | nindent 10 }}
         {{- end }}
         image: {{ template "splunk-otel-collector.image.otelcol" . }}
         imagePullPolicy: {{ .Values.image.otelcol.pullPolicy }}
--- a/helm-charts/splunk-otel-collector/templates/deployment-gateway.yaml
+++ b/helm-charts/splunk-otel-collector/templates/deployment-gateway.yaml
@@ -61,9 +61,10 @@ spec:
       affinity:
         {{- toYaml $gateway.affinity | nindent 8 }}
       {{- end }}
-      {{- if $gateway.securityContext }}
+      {{- $podSecurityContext := $gateway.podSecurityContext | default $gateway.securityContext }}
+      {{- if $podSecurityContext }}
       securityContext:
-        {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $gateway.securityContext) | nindent 8 }}
+        {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $podSecurityContext) | nindent 8 }}
       {{- end }}
       containers:
       - name: otel-collector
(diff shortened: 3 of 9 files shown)


helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -577,9 +577,36 @@
             }
           }
         },
+        "podSecurityContext": {
+          "type": "object"
+        },
         "securityContext": {
```
dmitryax, line 583: This can also be deprecated in favor of `containerSecurityContext` but I think we can defer that until fluentd is removed

helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -78,6 +78,10 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.agent.podSecurityContext }}
```
atoulme, line 81: I see `$agent.securityContext` used in this file quite a bit, please take a look?

dmitryax: `$agent` is to support the old section that was called `otelAgent`. It's not relevant for new fields

dmitryax: `podSecurityContext` is a new field

helm-charts/splunk-otel-collector/templates/NOTES.txt
```diff
@@ -81,3 +81,9 @@ Splunk OpenTelemetry Collector is installed and configured to send data to Splun
   - Status: Instrumentation language maturity varies. See `operator.instrumentation.spec` and documentation for utilized instrumentation details.
   - Splunk Support: We offer full support for Splunk distributions and best-effort support for native OpenTelemetry distributions of auto-instrumentation libraries.
 {{- end }}
+{{- if not (eq (toString .Values.gateway.securityContext) "<nil>") }}
+[WARNING] "gateway.securityContext" parameter is deprecated. Please use "gateway.podSecurityContext" instead.
+{{ end }}
+{{- if not (eq (toString .Values.clusterReceiver.securityContext) "<nil>") }}
+[WARNING] "clusterReceiver.securityContext" parameter is deprecated. Please use "clusterReceiver.podSecurityContext" instead.
+{{ end }}
```
jvoravong, line 89: Nit: Can we throw a failure or print a warning if one of the following cases happens? I just want to avoid confusion about which of the two available values (securityContext vs podSecurityContext) will be used.
- Both gateway.securityContext and gateway.podSecurityContext are used
- Both gateway.securityContext and gateway.podSecurityContext are used

jvoravong (approved):

jinja2 (commented):
We are missing updates in few places in [this block](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/daemonset.yaml#L146-L178). I think it should be updated to `podSecurityContext` instead. Actually, we might have to do some checking to see which of the 2 (pod or container) securityContext has `runAsUser` set. Both can have this field.

jinja2:
Another thing - do we know if the original ask also wants a way to set the context on initContainer? I know these only come into play when logs are enabled, but best to check if we need to include those.

jvoravong:
> Another thing - do we know if the original ask also wants a way to set the context on initContainer? I know these only come into play when logs are enabled, but best to check if we need to include those.

From my current understanding initContainers are used by the agent daemonset to update permissions for enabling collecting logs and all 3 agent modes to enable using persistent data exporter queues normally only used by Splunk Enterprise/Platform customers.

dmitryax:
> We are missing updates in few places in [this block](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/daemonset.yaml#L146-L178). I think it should be updated to podSecurityContext instead. Actually, we might have to do some checking to see which of the 2 (pod or container) securityContext has runAsUser set. Both can have this field.

I've actually started with adding this to the init containers as well, but I'm afraid of unnecessarily bloating up the config interface, see the first version in https://github.com/signalfx/splunk-otel-collector-chart/commit/74e35dd2361bb76ba38730dc87392b159220d2c9. I think we can start without it at this point, see if it's needed for users, then add it. WDYT?

> Another thing - do we know if the original ask also wants a way to set the context on initContainer? I know these only come into play when logs are enabled, but best to check if we need to include those.

No, as far as I know. That's why I removed it eventually

helm-charts/splunk-otel-collector/templates/NOTES.txt
dmitryax: I'm not sure I understand your suggestion. Please submit a suggesting diff

helm-charts/splunk-otel-collector/templates/deployment-gateway.yaml
```diff
@@ -61,9 +61,10 @@ spec:
       affinity:
         {{- toYaml $gateway.affinity | nindent 8 }}
       {{- end }}
-      {{- if $gateway.securityContext }}
+      {{- $podSecurityContext := $gateway.podSecurityContext | default $gateway.securityContext }}
+      {{- if $podSecurityContext }}
       securityContext:
-        {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $gateway.securityContext) | nindent 8 }}
+        {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $podSecurityContext) | nindent 8 }}
```
jvoravong, line 67:
```suggestion
        {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $gateway.podSecurityContext) | nindent 8 }}
```

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
```diff
@@ -81,9 +81,10 @@ spec:
         ` }}
         {{- $clusterReceiver.affinity | mustMergeOverwrite (fromYaml $clusterReceiverPodAntiAffinity) | toYaml | nindent 8 }}
       {{- end }}
-      {{- if $clusterReceiver.securityContext }}
+      {{- $podSecurityContext := $clusterReceiver.podSecurityContext | default $clusterReceiver.securityContext }}
+      {{- if $podSecurityContext }}
       securityContext:
-        {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $clusterReceiver.securityContext) | nindent 8 }}
+        {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "securityContext" $podSecurityContext) | nindent 8 }}
```
jvoravong, line 87:
```suggestion
        {{- include "splunk-otel-collector.securityContext" (dict "isWindows" .Values.isWindows "clusterReceiver.securityContext" $podSecurityContext) | nindent 8 }}
```

helm-charts/splunk-otel-collector/templates/deployment-gateway.yaml
dmitryax: Why?

helm-charts/splunk-otel-collector/templates/deployment-cluster-receiver.yaml
dmitryax: Why?

jinja2 (approved):

helm-charts/splunk-otel-collector/templates/NOTES.txt
jinja2, line 89:
```suggestion
{{- if not (eq (toString .Values. clusterReceiver.securityContext) "<nil>") }}
{{- if .Values. clusterReceiver.podSecurityContext }}
[WARNING] "clusterReceiver.securityContext" parameter is deprecated. Helm release will use values set in "clusterReceiver.podSecurityContext" instead. 
{{- else }}
[WARNING] "clusterReceiver.securityContext" parameter is deprecated. Please use "clusterReceiver.podSecurityContext" instead.
{{- end }}
{{ end }}
```
@jvoravong is this what you are suggesting? How does the PR you referenced relate to this one? Tmk this is a different ask

PR #1648, opened 2025-02-05
[Draft] Use Helm generated instead of cert-manager generated certs for the operator

jvoravong:
- Initial changes for using Helm generated instead of cert-manager generated certs for the operator by default
- TODO: Final docs updates for customer, developer, and migration content
- TODO: Edge case testings


helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -1731,12 +1731,14 @@
     "certmanager": {
       "description": "cert-manager adds certificates and certificate issuers as resource types in Kubernetes clusters, and simplifies the process of obtaining, renewing and using those certificates.",
       "type": "object",
-      "additionalProperties": true
+      "additionalProperties": true,
+      "deprecated": true
```
jinja2, line 1735: The schema docs - `The deprecated keyword is a boolean that indicates that the [instance](https://json-schema.org/learn/glossary#instance) value the keyword applies to should not be used and may be removed in the future.` Since we are removing cert-manager completely here, should we just remove the instance definitions from schema? Deprecated makes it seem like setting the value will still get certmanager installed

docs/auto-instrumentation-install.md
```diff
@@ -533,10 +529,93 @@ kubectl get certificates
...
+  admissionWebhooks:
+    certManager:
+      enabled: true
+      certificateAnnotations:
+        "helm.sh/hook": post-install,post-upgrade
+        "helm.sh/hook-weight": "1"
+      issuerAnnotations:
+        "helm.sh/hook": post-install,post-upgrade
+        "helm.sh/hook-weight": "1"
+certmanager:
```
jinja2, line 587: I am a little confused here. In the chart.yaml, certmanager has been removed from dependency. Are we targeting the removal in this PR?

helm-charts/splunk-otel-collector/values.schema.json
jvoravong: - I would not recommend hard removing (deleting the json entry) any fields in values.schema.json at this time. The values.schema.json is used in various ways by Helm and providers of Helm products where I would only recommend applying the `deprecated: true` to values we no longer use.

docs/auto-instrumentation-install.md
jvoravong: Including these changes was in flux depending a few factors. Last update we are actually going to keep using certmanager as a subchart.

helm-charts/splunk-otel-collector/values.schema.json
jinja2: Ok, if you are updating the PR to keep the cert-manager dependency, then adding the deprecated annotation sounds good. My request for removing it was contingent upon the removal of cert-manager, which, as I mentioned in the other comment, I couldn't figure out was part of this PR or not.

docs/auto-instrumentation-install.md
jinja2: sounds good

docs/auto-instrumentation-install.md
```diff
@@ -533,10 +529,93 @@ kubectl get certificates
...
+
+---
+
+##### 1. **Using a Self-Signed Certificate Generated by the Chart**
+
+This is the default and simplest method for generating a TLS certificate. It automatically creates a self-signed certificate for the webhook. It is suitable for internal environments or testing purposes but may not be trusted by clients outside your cluster.
+
+**Configuration:**
+- Set `admissionWebhooks.certManager.enabled` to `false` and `admissionWebhooks.autoGenerateCert.enabled` to `true`.
+- Helm generates a self-signed certificate, valid for 10 years, and creates a secret for the webhook.
```
jinja2, line 544:
```suggestion
- Helm generates a self-signed certificate, valid for 10 years, and creates a secret for the webhook. The validity of this self-signed certificate can be adjust with the config `operator.admissionWebhooks.autoGenerateCert.certPeriodDays`
```

docs/auto-instrumentation-install.md
```diff
@@ -533,10 +529,93 @@ kubectl get certificates
...
+- The certificate is automatically recreated on every Helm upgrade.
```
jinja2, line 545:
```suggestion
- The certificate is automatically recreated on every Helm upgrade. This behavior can be disabled by setting the config `operator.admissionWebhooks.autoGenerateCert.recreate` to `false`
```

jinja2:
I think we'll need an entry for how existing certmanager subchart users can keep their current setup in the [upgrading](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md) guide which we can add to in a changelog entry.

And I think it would be good to document if user can safely disable existing certmanager and use the autogenerate feature when they helm upgrade to a version of chart with this change.

docs/auto-instrumentation-install.md
jvoravong: I found a [sizable issue](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/be4c2659ec3005b1dbce238df7d730c5461746ba/charts/opentelemetry-operator/templates/admission-webhooks/operator-webhook.yaml#L12) related to this I'm trying to test fixes for locally and make a fix for upstream.

jinja2: Can you clarify what is the issue? secret being recreated on each upgrade?

jvoravong: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1531

jvoravong: Added this change.

UPGRADING.md
```diff
@@ -1,5 +1,101 @@
...
+### **Maintaining Previous Functionality via Helm Values Update**
+
+#### **Scenario 1: Operator and cert-manager Deployed via This Helm Chart**
+
+If you previously deployed both the Operator and cert-manager via this Helm chart (`operator.enabled=true` and `certmanager.enabled=true`), you can preserve functionality by adding the following values:
+
+```yaml
+operator:
+  enabled: true
+  admissionWebhooks:
```
jinja2, line 21: Does a user not need to explicitly disable the now true option `autoGenerateCert.enabled`?

jvoravong: The user does not. Whenever `operator.admissionWebhooks.certManager.enabled=true` the usage of `operator.admissionWebhooks.autoGenerateCert.enabled` is nullified. Adding a doc update relating to this for a little more clean up.

UPGRADING.md
```diff
@@ -1,5 +1,101 @@
...
+
+```
+warning: Upgrade "{helm_release_name}" failed: pre-upgrade hooks failed: warning: Hook pre-upgrade splunk-otel-collector/charts/operator/templates/admission-webhooks/operator-webhook.yaml failed: 1 error occurred:* secrets "splunk-otel-collector-operator-controller-manager-service-cert" already exists
+```
+
+This typically occurs because:
+- cert-manager deletes its `Certificate` resources immediately.
+- However, cert-manager does not delete the associated **secrets** instantly. It waits for its garbage collector process to remove them.
+
+You will first have to delete this chart, wait for cert-manager to do garbage collection, and then install the latest version of this chart.
```
jinja2, line 63: Is there no way for a user to switch without having to reinstall it? My understanding reading this doc and the helm chart templates is that the main issue is the use of same secret for cert-manager and autoGenerateCert.

I haven't tried it myself, but thinking out loud, would the following approach work?

- Set `operator.admissionWebhooks.certManager.enabled=false` and update the secret name to something else `operator.admissionWebhooks.secretName=new-certificate` and run helm upgrade. This should do the following if I am reading the templates correctly - 
  - create new secret object in pre install `new-certificate` the contents of which are going to be the self-signed key/cert.
  - remove the Certificate and Issuer objects
  - update the `volumes.secret.secretName` config of the operator deployment object to use the new secret
  - update the webhook configuration to add self-signed caBundle and remove the cert-manager injection annotation.
  
  We would want the certificate in operator pod to be update first so in an ideal situation we would just update the cert in the existing secret which would then cause kubelet to update the files in the pod and if operator is written to watch the cert, it would pick up the new certs w/o a bounce. This would then be followed by updating the ca in the webhook. With the helm upgrade there might be some issue with the certificate update happening in the webhook pod after the webhook config but [the value of failurePolicy](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-operator/values.yaml#L255) for the pod mutation hook config is Ignore so even if there was a weird situation where there is deadlock due to order in which this gets applied, the operator pod should still come up. We should probably check helm's installation order for webhookConfiguration objects.
- If this works, the next step would be to remove the cert-manager `certmanager.enabled=false`.  The secret `splunk-otel-collector-operator-controller-manager-service-cert` might still need to deleted if the cert-manager is not running with the config `--enable-certificate-owner-ref` as mentioned in their [docs](https://cert-manager.io/docs/usage/certificate/#cleaning-up-secrets-when-certificates-are-deleted).
  I think we should reference [cert-manager's uninstallation guide](https://cert-manager.io/docs/installation/helm/#uninstalling) for better guidance; it also includes CRD cleanup.

examples/enable-operator-and-auto-instrumentation/otel-demo-dotnet.md
```diff
@@ -24,14 +24,10 @@ curl https://raw.githubusercontent.com/signalfx/splunk-otel-collector-chart/main
 #### 2.1 Deploy the Helm Chart with the Operator enabled
 
 To install the chart with operator in an existing cluster, make sure you have cert-manager installed and available.
-Both the cert-manager and operator are subcharts of this chart and can be enabled with `--set operatorcrds.install=true,operator.enabled=true,certmanager.enabled=true`.
+Both the cert-manager and operator are subcharts of this chart and can be enabled with `--set operatorcrds.install=true,operator.enabled=true`.
```
jinja2, line 27: Did you mean to remove `cert-manager` from here?

examples/enable-operator-and-auto-instrumentation/otel-demo-nodejs.md
```diff
@@ -20,15 +20,11 @@ curl https://raw.githubusercontent.com/signalfx/splunk-otel-collector-chart/main
 #### 2.1 Deploy the Helm Chart with the Operator enabled
 
 To install the chart with operator in an existing cluster, make sure you have cert-manager installed and available.
-Both the cert-manager and operator are subcharts of this chart and can be enabled with `--set operatorcrds.install=true,operator.enabled=true,certmanager.enabled=true`.
+Both the cert-manager and operator are subcharts of this chart and can be enabled with `--set operatorcrds.install=true,operator.enabled=true`.
```
jinja2, line 23:
```suggestion
The opentelemetry-operator is a subchart of this chart and can be enabled with `--set operatorcrds.install=true,operator.enabled=true`.
```

Not sure why are we mentioning the cert-manager

examples/enable-operator-and-auto-instrumentation/otel-demo-dotnet.md
jvoravong: Yes

jvoravong:
Having to use a different PR label to get the upgrade tests to pass, a new PR label can be added after https://github.com/signalfx/splunk-otel-collector-chart/pull/1682.

examples/enable-operator-and-auto-instrumentation/otel-demo-nodejs.md
jvoravong: Yeah my grep missed these in our example demo docs for this specific section. Pushed out a commit to update these docs with your input in mind.

UPGRADING.md
jvoravong: > Is there no way for a user to switch without having to reinstall it? My understanding reading this doc and the helm chart templates is that the main issue is the use of same secret for cert-manager and autoGenerateCert.
> 
> I haven't tried it myself, but thinking out loud, would the following approach work?
> 
> * Set `operator.admissionWebhooks.certManager.enabled=false` and update the secret name to something else `operator.admissionWebhooks.secretName=new-certificate` and run helm upgrade. This should do the following if I am reading the templates correctly -
>   
>   * create new secret object in pre install `new-certificate` the contents of which are going to be the self-signed key/cert.
>   * remove the Certificate and Issuer objects
>   * update the `volumes.secret.secretName` config of the operator deployment object to use the new secret
>   * update the webhook configuration to add self-signed caBundle and remove the cert-manager injection annotation.

- I think this point includes both ideas I support and some I’m less in favor of.
  - Yes, we could rename the secret generated for a one-way upgrade migration, but this could lead to issues if the user downgrades and uses the same values. They could end up in an inconsistent state, and I don’t think we can document all potential use cases.
  - I want to promote the support of not preinstalling Helm-generated certificates because they use Helm hooks, meaning the secrets containing the Helm-generated certs aren’t deleted when the chart is deleted (potentially causing future install issues). I created a [draft PR](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1531) for this, but it hasn't received much attention.
- The best solution in my opinion would be to have the upstream OpenTelemetry operator Helm chart use different secret names depending on whether a cert-manager, Helm-generated, or user-supplied certificate is used. However, due to time constraints and possible reluctance from upstream without further refinement, I’d prefer to move forward with this PR as-is. We can add better migration steps in the future. The existing UPGRADING.md steps already supports users maintaining previous functionality.

> * If this works, the next step would be to remove the cert-manager `certmanager.enabled=false`.  The secret `splunk-otel-collector-operator-controller-manager-service-cert` might still need to deleted if the cert-manager is not running with the config `--enable-certificate-owner-ref` as mentioned in their [docs](https://cert-manager.io/docs/usage/certificate/#cleaning-up-secrets-when-certificates-are-deleted).
>   I think we should reference [cert-manager's uninstallation guide](https://cert-manager.io/docs/installation/helm/#uninstalling) for better guidance; it also includes CRD cleanup.

- In my initial work, I investigated the usage of --enable-certificate-owner-ref. However, I don’t consider it a solution for this effort, especially since `--enable-certificate-owner-ref` is not enabled by default when installing cert-manager. To delete secrets when a cert-manager certificate is deleted, the `--enable-certificate-owner-ref` startup parameter must have already been used to deploy cert-manager. Adding it to our new Helm chart version would not resolve our migrations issues. Additionally, we have customers who have already deployed cert-manager without our chart and without `--enable-certificate-owner-ref`, so it wouldn’t help them either.
- We already mention cert-manager deletion in the docs related to this change at [auto-instrumentation-install.md#tls-certificate-requirement-for-kubernetes-operator-webhooks](https://github.com/signalfx/splunk-otel-collector-chart/blob/otl-2900-migration/docs/auto-instrumentation-install.md#tls-certificate-requirement-for-kubernetes-operator-webhooks).
- I validated the installation order for webhookConfiguration objects as part oft https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1531.

jvoravong: Just created a ticket with some of my notes for improving Helm docs because I think it will help the community in the long run.
https://github.com/helm/helm/issues/30617

jinja2: > We already mention cert-manager deletion in the docs related to this change at [auto-instrumentation-install.md#tls-certificate-requirement-for-kubernetes-operator-webhooks](https://github.com/signalfx/splunk-otel-collector-chart/blob/otl-2900-migration/docs/auto-instrumentation-install.md#tls-certificate-requirement-for-kubernetes-operator-webhooks).

I don't see us mention/link to the official cert-manager guidelines for uninstalling it fully including the CRDs which helm delete is not removing.

jinja2 (approved):

PR #1652, opened 2025-02-10
[chore] rewrite python tests into golang and remove python tests

pszkamruk-splunk:
**Description:** This PR contains changes related to removing the Python dependency from the project. Python was previously used for Splunk integration tests, which have now been rewritten in Golang.

**Changes:**
1. Rewriting existing Python tests in Golang.
2. Removing Python-related test files and unnecessary configuration files.
3. Updating the pipeline to run Golang Splunk integration tests.
4. Improving test verifications.
5. Adding missing tests to fully test the pod annotations feature (the setup.yaml file contains configurations that were not fully tested).

**Next steps:**
1. Integrate Golang Splunk integration tests with existing functional tests 
- use testcontainers-go to spin Splunk
- setup environment via `deployEnv` function
- integration with existing functional test matrix

PS. Current pipeline which runs minikube with different container runtimes for docker will be removed  this will reduce the number of jobs to run. Also number of tested Splunk versions will be reduced to the latest one.


test/go.mod
```diff
@@ -0,0 +1,10 @@
+module splunk-integration-tests
+
+go 1.23.1
```
sbylica-splunk, line 3: Maybe 1.24.0 like in functional tests to keep it simple?

https://github.com/signalfx/splunk-otel-collector-chart/blob/0180b914648ab0f7a64562104558b6ce7d1ff073/functional_tests/go.mod#L6

.github/workflows/functional_test.yaml
```diff
@@ -3,6 +3,12 @@ name: Splunk Platform Functional Test
 on:
   pull_request:
 
+env:
+  GO_VERSION: 1.24.1
```
jvoravong, line 7: I like pinning the go version. We should probably keep the major and minor version in sync with the [collector](https://github.com/signalfx/splunk-otel-collector/blob/a389cf7e71ec107394be888e4d9d28f12df08afd/go.mod#L3C1-L3C10)

jvoravong (approved):
LGTM. I can think of some improvements to add later, but this work gets us closer to our end goals.

mstopa-splunk (changes requested):

.github/workflows/functional_test.yaml
```diff
@@ -97,30 +107,13 @@ jobs:
...
           sleep 60
 
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.10"
-
-      - name: Run functional tests
+      - name: Run Splunk Integration Tests
+        env:
+          K8S_VERSION: ${{ matrix.k8s-version }}
```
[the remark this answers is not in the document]
pszkamruk-splunk: actually this variable is not used here, not sure why I put it here
I have removed it

.github/workflows/functional_test.yaml
```diff
@@ -3,6 +3,12 @@ name: Splunk Platform Functional Test
 on:
   pull_request:
 
+env:
```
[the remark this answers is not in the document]
pszkamruk-splunk: I would keep it as it is, the same approach is used in `functional_test_v2` pipeline
this global variable defines what Golang version should be used by GitHub runner

test/go.mod
pszkamruk-splunk: updated to `go 1.24.1`

PR #1654, opened 2025-02-12
Option change the default internal traffic policy for the agent service

icicimov:
**Description:** 

During our Helm deployments in EKS it came up that the hard coded `internalTrafficPolicy: Local` causes a communication issues with the service returning 404 error which was fixed by manually changing this to `Cluster`. This PR adds the option to override the value of the `internalTrafficPolicy` in the chart.

### DIFF
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -408,6 +408,9 @@ agent:
     # so that agent pods can be discovered via dns etc
     enabled: true
 
+    # (optional) Override the default service internal traffic policy
+    internalTrafficPolicy: "Local"
+
   # hostNetwork schedules the pod with the host's network namespace.
   # Disabling this value will affect monitoring of some control plane
   # components. Enabling the agent service is recommended (see above).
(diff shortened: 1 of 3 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -408,6 +408,9 @@ agent:
     # so that agent pods can be discovered via dns etc
     enabled: true
 
+    # (optional) Override the default service internal traffic policy
+    internalTrafficPolicy: "Local"
```
pszkamruk-splunk, line 412: I think it would be good to add information about available options like it is done [here](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/values.yaml#L182) for `cloudProvider`

jvoravong, line 412: Can we make this more optional to not bloat the set values and be more portable? This would require updating the values.schema.json file.
```suggestion
    # internalTrafficPolicy: "Local"
```

jinja2:
@icicimov can you describe your k8s setup and error in more details? With `Cluster` type policy, traffic to the agent can be from any node, which can lead to issues with pod metadata association in k8sattribute processor. The k8sattribute processor is configured to watch on pods on the local node only. The traffic policy `Local` ensures that a collector agent only receives traffic from the local node.

jinja2:
Since we haven't heard back on this PR, I'm going to close it for now. Thanks for your contribution and feel free to open a new one if you'd like to continue.

PR #1665, opened 2025-02-18
[chore] Add optional label to ignore functional tests

jvoravong:
Add optional PR label to ignore functional test results. This allows us to check in known breaking changes that will cause functional tests to fail, useful for Helm chart version upgrade tests.

### DIFF
--- a/.github/workflows/functional_test_v2.yaml
+++ b/.github/workflows/functional_test_v2.yaml
@@ -46,7 +46,7 @@ jobs:
           - configuration_switching
           - istio
     runs-on: ubuntu-latest
-    continue-on-error: ${{ !contains(github.event.pull_request.labels.*.name, 'Ignore Tests') }}
+    continue-on-error: ${{ contains(github.event.pull_request.labels.*.name, 'Ignore Tests') }}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
@@ -108,7 +108,7 @@ jobs:
       KUBE_TEST_ENV: eks
       SKIP_TESTS: "true" # we need to skip functional tests as we have not set probes to listen to the traffic.
     runs-on: ubuntu-latest
-    continue-on-error: ${{ !contains(github.event.pull_request.labels.*.name, 'Ignore Tests') }}
+    continue-on-error: ${{ contains(github.event.pull_request.labels.*.name, 'Ignore Tests') }}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
@@ -147,7 +147,7 @@ jobs:
       KUBE_TEST_ENV: eks
       SKIP_TESTS: "true" # we need to skip functional tests as we have not set probes to listen to the traffic.
     runs-on: ubuntu-latest
-    continue-on-error: ${{ !contains(github.event.pull_request.labels.*.name, 'Ignore Tests') }}
+    continue-on-error: ${{ contains(github.event.pull_request.labels.*.name, 'Ignore Tests') }}
     steps:
       - uses: actions/checkout@v4
       - name: Checkout main if in a PR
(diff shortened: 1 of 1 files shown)


jvoravong:
This blocks https://github.com/signalfx/splunk-otel-collector-chart/pull/1648 from merging

.github/workflows/functional_test_v2.yaml
crobert-1, line 1: Can we change the label name to be a bit more specific, something like "Ignore functional test results", or "Allow functional test failure"? Happy to hear different suggestions, but `Ignore Tests` is a little too broad, and would cause me to be concerned if I forgot the purpose of it.

(Maybe also as a team we should include comments on PRs when adding the label to explain why we're adding it.)

.github/workflows/functional_test_v2.yaml
```diff
@@ -46,6 +46,7 @@ jobs:
           - configuration_switching
           - istio
     runs-on: ubuntu-latest
+    continue-on-error: ${{ !contains(github.event.pull_request.labels.*.name, 'Ignore Tests') }}
```
crobert-1, line 49: Shouldn't the condition be flipped here? We want to continue on error if the PR contains the label.

.github/workflows/functional_test_v2.yaml
crobert-1: Follow up from team sync: The label name can be left as is, we'll just try to add a description to the label in GitHub that can accurately describe the intention.

.github/workflows/functional_test_v2.yaml
jinja2: Yeah, looks like it should be flipped.

Can non org members add labels on PRs btw?

jvoravong: Github workflow parameters like this use passed values to functions in a syntax ordering that seems backwards.
Other Examples:
- https://github.com/signalfx/splunk-otel-collector-chart/blob/2486516e3785ff4a9821c9307e07edc2de951668/.github/workflows/validate-changelog.yaml#L12
- https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/workflows/changelog.yml#L49
- Non-org members cannot at labels.

jvoravong: Going to add some Github label descriptions for this new label and previous labels that are missing descriptions.

jinja2: The `continue-on-error` config determines whether the job should continue running even if a step fails. My understanding is that your intention with this PR is to let the action continue on error if the `Ignore Tests` label is set. The condition should be reversed by removing the `!` operator in that case (we want the config value to be true when the label exists).

In the quoted validate-changelog.yaml file `if` condition, it ensures that the job runs only when the `Skip Changelog` label is not present and the pull request title does not contain `[chore]`. I am not sure I understand what you mean by backwards ordering.

crobert-1: > Github workflow function like this use parameter ordering that seems backwards, but this is correct.

Just to clarify, is this in reference to the condition comment? The examples you shared are correct because the logic is "If the label is not applied, validate the changelog exists and is valid." In this case, we're saying "If the label is applied, we want to continue on error". So the goal of the label is to have a check of "If true, return true", at least the way it's worded.

Feel free to clarify further, I may be misunderstanding something here.

jvoravong: Okay now I see what you folks are talking about. My apologies, between my github workflow and bash code related work lately I can interpret the initial comments in several ways.
- I can make updates for this because you folks are right. I might ask for a explicit/specific code suggestions if possible in the future, granted, this one was mostly on me.
- I need to test changes with several other non-trivial separate code branches involved at the same time so I'm lagging on testing sometimes.

jinja2 (approved):

PR #1667, opened 2025-02-19
[chore] Fix for the github workflow that updates the operator subchart

jvoravong:
- Fix for the github workflow that updates the operator subchart, was failing due to a search token missing and indentation issues.
- Related workflow run that failed: https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/13369805246/job/37335625315

### DIFF
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -1263,6 +1263,7 @@ instrumentation:
   #   - repository: Specifies the Docker image repository.
   #   - tag: Indicates the Docker image tag.
   #   - env: (Optional) Allows you to add any additional environment variables.
+  # Auto-instrumentation Libraries (Start)
   java:
     repository: ghcr.io/signalfx/splunk-otel-java/splunk-otel-java
     tag: v2.12.0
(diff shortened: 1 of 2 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1263,6 +1263,7 @@ instrumentation:
   #   - repository: Specifies the Docker image repository.
   #   - tag: Indicates the Docker image tag.
   #   - env: (Optional) Allows you to add any additional environment variables.
+  # Auto-instrumentation Libraries (Start)
```
crobert-1, line 1266: What's the added value of this comment? I see the comment block's first line is almost identical, so I'm wondering why we need this line as well.

crobert-1 (approved):
LGTM, just a tiny question on the added comment.

helm-charts/splunk-otel-collector/values.yaml
jvoravong: Good question. 

We use the text tokens `# Auto-instrumentation Libraries (Start)` and `# Auto-instrumentation Libraries (End)` to identify the exact section of the values.yaml file that should be updated.

With our current CI/CD tools, if we didn’t specify a section to modify, the entire file would be rewritten (including formatting and comments) resulting in undesirable and systematic changes.

That said, I believe we could improve/replace this process with Renovate CI/CD tools and some additional developer work. This isn’t a priority right now, but it’s something we may consider in the future. For now, we’re using a minimal solution that ensures only the necessary updates are made while preserving the rest of the file.

PR #1677, opened 2025-02-26
[chore] Extend functional test k8s version support matrix

jvoravong:


jvoravong:
New expected test data entries for Kind 1.31 and 1.32 need to be added for the functional historgram tests.

PR #1682, opened 2025-02-28
[chore] migrate from using the "Ignore Tests" PR label to "Skip Upgrade Tests"

jvoravong:
- This repository's settings won't allow us to use the github worfklow "continue-on-error" clause and "Ignore Tests" PR label as originally intended. When using "continue-on-error" it would still cause our required checks to fail if a step failed, granted downstream workflows from that step would still run.
- We'll have to migrate to using a github worfklow "if" clause and a label like "Skip Upgrade Tests". I tested this approach in my forked repository, I don't for see any issues with this approach in this repository.

### DIFF
--- a/.github/workflows/functional_test_v2.yaml
+++ b/.github/workflows/functional_test_v2.yaml
@@ -47,7 +47,6 @@ jobs:
           - k8sevents
           - istio
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jvoravong:
Looks like our EKS and AKS test clusters need some manual maintenance. The EKS cluster somehow got into a bad state. The AKS cluster credentials expired or are no longer valid. I'll see what I can do.

jvoravong:
This might not be needed anymore. Closing it for now.

PR #1683, opened 2025-03-01
[chore] Update collector image to 0.120.0

dmitryax:
This also updates a test validating expected internal collector metrics. The upgrade to Prometheus 3.0 caused that `.` in metric names are not replaced to `_` by the Prometheus receiver. As the result, there is one metric affected in the test `otelcol_processor_filter_logs_filtered` needs to be updated to `otelcol_processor_filter_logs.filtered`. There may be couple more of collector internal metrics affected. Not a big problem. I'll find them and update the changelog upstream with more details.

### DIFF
--- a/examples/add-filter-processor/rendered_manifests/clusterRole.yaml
+++ b/examples/add-filter-processor/rendered_manifests/clusterRole.yaml
@@ -9,7 +9,7 @@ metadata:
     helm.sh/chart: splunk-otel-collector-0.119.0
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: default
-    app.kubernetes.io/version: "0.119.0"
+    app.kubernetes.io/version: "0.120.0"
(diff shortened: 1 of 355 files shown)


jvoravong:
Looks like our EKS and AKS test clusters need some manual maintenance. The EKS cluster somehow got into a bad state. The AKS cluster credentials expired or are no longer valid. I'll see what I can do.

dmitryax:
> Looks like our EKS and AKS test clusters need some manual maintenance. The EKS cluster somehow got into a bad state. The AKS cluster credentials expired or are no longer valid. I'll see what I can do.

I believe they has been failing regardless of the upgrade

jvoravong:
Requesting we merge https://github.com/signalfx/splunk-otel-collector-chart/pull/1648 before this PR.

jinja2 (approved):

jvoravong (approved):

PR #1686, opened 2025-03-04
[chore] Add FIPS image configuration doc snippet

jvoravong:


jinja2 (changes requested):
Are we adding this due to product ask? I don't think this is the right terminology. We can state that a user can leverage our FIPS-compliant image with the chart. But FedRAMP is a much broader framework and just running an application in FIPs mode is not enough. I would recommend getting this verified with product.

Also, I think there might be some more chart configs which might not be usable with the fips image. Example, any chart config which enables a smartagent receiver.

atoulme:
No, product didn't ask for this, we will update docs over time, and not in the chart repository. Thanks!

PR #1687, opened 2025-03-04
fix: enable configuring fipsEnabled otelcollector image

egkelly:

Bugfix: When using the fips otelcollector image, the migratecheckpoint init container (and subsequently, the rest of the deployment) fails to start, as the [binary has been excluded from the fips image](https://github.com/signalfx/splunk-otel-collector/blob/e586a6f0f2f11b7e7686b9a7ac708d5149dd19bb/packaging/docker-otelcol.sh#L73). This change also edits the image.otelcol.repository such that `-fips` is appended to the repository.

None, but I have opened a case #3672212, and wanted to help push this fix effort along.

Ran `helm template mytestchart . --values=values.test` with the following values:
```
clusterName: my-cluster
splunkPlatform:
  endpoint: "https://test.platform:8080/services/collector/event"
  index: test
  # A real certificate is not required in caFile as long as it exists in the splunk_hec_token secret
  # caFile requires a value so the Splunk OTeL Pods read the certificate from the splunk_hec_token secret
  caFile: enabled
  logsEnabled: true
  metricsEnabled: false
  tracesEnabled: false
secret:
  create: false
  name: ${secret_name}
cloudProvider: aws
distribution: eks
image:
  otelcol:
    fipsEnabled: true
```
The deployment started and ran as expected. When `fipsEnabled: false` as is default, the deployment fails with 
```
failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/migratecheckpoint": stat /migratecheckpoint: no such file or directory: unknown
```

Documented in the [values](https://github.com/signalfx/splunk-otel-collector-chart/blob/bc6f0934eb9014a35169bb8e4b7e626b135d5746/helm-charts/splunk-otel-collector/values.yaml#L972):
```
    # To use the FIPS-140 enabled version, set this value to true
    # As a result, the repository will become quay.io/signalfx/splunk-otel-collector-fips
    fipsEnabled: false
```

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/_helpers.tpl
+++ b/helm-charts/splunk-otel-collector/templates/_helpers.tpl
@@ -210,9 +210,12 @@ Create the fluentd image name.
 Create the opentelemetry collector image name.
 */}}
 {{- define "splunk-otel-collector.image.otelcol" -}}
-{{- printf "%s:%s" .Values.image.otelcol.repository (.Values.image.otelcol.tag | default .Chart.AppVersion) -}}
+{{- $repository := .Values.image.otelcol.repository -}}
+{{- if .Values.image.otelcol.fipsEnabled -}}
+{{- $repository = printf "%s-fips" $repository -}}
 {{- end -}}
-
+{{- printf "%s:%s" $repository (.Values.image.otelcol.tag | default .Chart.AppVersion) -}}
+{{- end -}}
 {{/*
 Create the patch-log-dirs image name.
 */}}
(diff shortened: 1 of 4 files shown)


atoulme (changes requested):
Thank you for your contribution. We appreciate it.
At this time, a workaround is to set `agent.skipInitContainers=true` to avoid this issue.
We will also stop running migrate-checkpoint by default moving forward, as this only applies to fluentd migrations.

PR #1690, opened 2025-03-05
fixing missing logs while pod log file being rolled over

pszkamruk-splunk:
**Description:** 
This PR contains changes related to the bug raised by one of our customer. They mentioned that not all logs were being collected from the log files due to the files being rotated.
The changes to fix above issue have been added by using `featureGates.fixMissedLogsDuringLogRotation` flag. Setting that flag to `true` will enable the fix and will allow connector to collect logs also from files which are being rolled over. 

image

**Testing:** checked manually, no duplicated logs, no missed logs

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -410,14 +410,31 @@ receivers:
   {{- if and (eq (include "splunk-otel-collector.logsEnabled" .) "true") (eq .Values.logsEngine "otel") }}
   {{- if .Values.logsCollection.containers.enabled }}
   filelog:
+    {{- if not .Values.featureGates.fixMissedLogsDuringLogRotation }}
     {{- if .Values.isWindows }}
     include: ["C:\\var\\log\\pods\\*\\*\\*.log"]
     {{- else }}
     include: ["/var/log/pods/*/*/*.log"]
     {{- end }}
+    {{- else }}
+    {{- if .Values.isWindows }}
+    include: ["C:\\var\\log\\pods\\*\\*\\*.log*"]
+    {{- else }}
+    include: ["/var/log/pods/*/*/*.log*"]
+    {{- end }}
+    {{- end }}
     # Exclude logs. The file format is
     # /var/log/pods/<namespace_name>_<pod_name>_<pod_uid>/<container_name>/<restart_count>.log
     exclude:
+      {{- if .Values.featureGates.fixMissedLogsDuringLogRotation }}
+      {{- if .Values.isWindows }}
+      - "C:\\var\\log\\pods\\*\\*\\*.log*.gz"
+      - "C:\\var\\log\\pods\\*\\*\\*.log*.tmp"
+      {{- else }}
+      - "/var/log/pods/*/*/*.log*.gz"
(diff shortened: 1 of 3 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -410,14 +410,29 @@ receivers:
...
+    {{- else }}
+    include: ["/var/log/pods/*/*/*.log*"]
+    {{- end }}
+    {{- end }}
     # Exclude logs. The file format is
     # /var/log/pods/<namespace_name>_<pod_name>_<pod_uid>/<container_name>/<restart_count>.log
     exclude:
+      {{- if .Values.featureGates.fixMissedLogsDuringLogRotation }}
+      {{- if .Values.isWindows }}
+      - "C:\\var\\log\\pods\\*\\*\\*.log*.gz"
```
dmitryax, line 431: We should probably add .tmp files here as well

pszkamruk-splunk:
> @pszkamruk-splunk, have you verified that this actually solves the problem? Do you know what's the container runtime and k8s cluster the customer has? AFAIK containerd should be doing copy/truncate, in which case this change isn't needed, see: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/fileconsumer/design.md#rotated-files-that-end-up-out-of-the-matching-pattern. Also, there is [relevant issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35777) where it's ended up to be another problem

I tested it locally with loggen pods, and as shown in the attached picture we are collecting events from rotated files. If the fix is not enabled (`featureGates.fixMissedLogsDuringLogRotation: false`) then from time to time I was observing some dropped logs. 
I saw docs regarding file rotation but I was not aware of that issue.  
I asked support to confirm what container runtime is being used by customer. 

My proposal is to proceed with this change and see if it will help on customer environment, and then we can decide if we want to keep it or remove it.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
pszkamruk-splunk: Done

pszkamruk-splunk:
@dmitryax 
The agency runs OpenShift 4 - specifically v4.16 and at the time this ticket was opened v4.12. All versions of OpenShift 4.x leverage cri-o as the container runtime

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: It's not windows only. Please add it in the else section as well

dmitryax:
> My proposal is to proceed with this change and see if it will help on customer environment, and then we can decide if we want to keep it or remove it.

Sounds good to me. Please address my other comment above, and we will merge it

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
pszkamruk-splunk: Added in `else` section 
Thank you :)

mstopa-splunk (approved):

dmitryax (approved):

PR #1691, opened 2025-03-05
Add k8sevent transform

jinja2:

For the option, `clusterReceiver.eventsEnabled`, the logs pipeline for k8s_events now adds attributes of the type `k8s.<objectkind>.name` and `k8s.<objectkind>.uid`.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
@@ -98,6 +98,27 @@ processors:
   {{- end }}
 
   {{- if and $clusterReceiver.eventsEnabled (eq (include "splunk-otel-collector.logsEnabled" .) "true") }}
+
+  # Add k8s event attributes - k8s.<kind>.name and k8s.<kind>.uid
+  transform/k8sevents:
+    error_mode: ignore
+    log_statements:
+      - conditions:
+          - resource.attributes["k8s.object.kind"] == "HorizontalPodAutoscaler"
+        statements:
+          - set(resource.attributes["k8s.hpa.name"], resource.attributes["k8s.object.name"])
+          - set(resource.attributes["k8s.hpa.uid"], resource.attributes["k8s.object.uid"])
+      - conditions:
+          - resource.attributes["k8s.object.kind"] != "HorizontalPodAutoscaler"
+        statements:
+          - set(resource.attributes[Concat(["k8s", ConvertCase(resource.attributes["k8s.object.kind"], "lower"), "name"], ".")], resource.attributes["k8s.object.name"])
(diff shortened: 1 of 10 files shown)


functional_tests/testdata_k8sevents/expected_k8sevents.yaml
```diff
@@ -127,8 +139,25 @@ resourceLogs:
...
+            stringValue: sock-splunk-otel-collector-k8s-cluster-receiver-859c547cf4gxb2v
+        - key: com.splunk.source
+          value:
+            stringValue: kubernetes
+        - key: com.splunk.sourcetype
+          value:
+            stringValue: kube:events
+        - key: com.splunk.index
+          value:
+            stringValue: index_from_pod
```
jinja2, line 158: Events in the resource before this ( e.g. `Successfully assigned k8sevents-test/k8sevents-test-0 to kind-control-plane`) do not have their index overridden with the annotation value. I think there'll be some ambiguity about the index with the existing config.

For example, an event `create Pod in StatefulSet k8sevents-test successful` which is associated with the object.kind = statefulset. It does not have a `k8s.node.name` value (this is set in the receiver to match the source.host so if event is from kubelet about a local pod activity this is set, but other controllers might not add this and even if it did this is meant to be the ID/host of the controller instance). It has the `k8s.namespace.name` set which is used as one of the association rule, but in our test we have no existing pod at the time this event was generated in the test ns. So I think the processor isn't be able to add the index value from the namespace and it remains `main`.

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -98,6 +98,27 @@ processors:
...
+          - set(resource.attributes["k8s.hpa.uid"], resource.attributes["k8s.object.uid"])
+      - conditions:
+          - resource.attributes["k8s.object.kind"] != "HorizontalPodAutoscaler"
+        statements:
+          - set(resource.attributes[Concat(["k8s", ConvertCase(resource.attributes["k8s.object.kind"], "lower"), "name"], ".")], resource.attributes["k8s.object.name"])
+          - set(resource.attributes[Concat(["k8s", ConvertCase(resource.attributes["k8s.object.kind"], "lower"), "uid"], ".")], resource.attributes["k8s.object.uid"])
+      - conditions:
+          - resource.attributes["k8s.object.kind"] == "Pod" and IsMatch(resource.attributes["k8s.object.fieldpath"], "spec\\.containers.*")
+        statements:
+          - merge_maps(resource.cache, ExtractPatterns(resource.attributes["k8s.object.fieldpath"], "spec.containers\\{(?P<k8s_container_name>[^\\}]+)\\}"), "insert")
```
jinja2, line 119: named capture group errors when `.` are used, so I am writing to cache and then setting the attribute in next step

dmitryax (approved):
LGTM

PR #1694, opened 2025-03-05
[chore] Move functional test to separate packages

dmitryax:
Move functional test to separate packages instead of using build tags. The build tags are not expected to use this way and hard to work with


dmitryax:
The k8s events has not been run apparently because of [the typo in the build tag](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/functional_tests/k8sevents_test.go#L4). They are [failing](https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/13687828736/job/38275124241?pr=1694). I'm going to skip them for now

jinja2 (approved):

jinja2:
Going to merge this and push a fix for the disabled test.

PR #1699, opened 2025-03-07
Do not setup the entities pipeline if Splunk Observability isn't enabled

dmitryax:
Resolves https://github.com/signalfx/splunk-otel-collector-chart/discussions/1693

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -40,8 +40,10 @@ receivers:
     endpoint: 0.0.0.0:8006
   {{- end }}
 
+  {{- if eq (include "splunk-otel-collector.splunkO11yEnabled" .) "true" }}
   # Placeholder receiver needed for discovery mode
   nop:
+  {{- end }}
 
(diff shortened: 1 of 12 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -1081,7 +1083,18 @@ service:
         {{- end }}
     {{- end }}
 
-    {{- if and (eq (include "splunk-otel-collector.splunkO11yEnabled" .) "true") .Values.featureGates.useControlPlaneMetricsHistogramData }}
+    {{- if eq (include "splunk-otel-collector.splunkO11yEnabled" .) "true" }}
```
jinja2, line 1086:
```suggestion
{{- if eq (include "splunk-otel-collector.splunkO11yEnabled" .) "true" }}
```
nit: aligns with the spacing of the corresponding {{- end }}

jinja2 (approved):

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
jinja2: nvm, i don't think that's the end

PR #1703, opened 2025-03-07
[job] Add a job to block release if main branch is broken

dmitryax:


.github/workflows/check-main-branch-ci.yaml
```diff
@@ -0,0 +1,58 @@
...
+            echo "Version changed!"
+          fi
+
+      - name: Check main branch build status
+        # Only run this step if the version actually changed
+        if: steps.compare.outputs.version_changed == 'true'
+        run: |
+          response=$(curl -s \
+            -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            "https://api.github.com/repos/${{ github.repository }}/commits/main/status")
```
jinja2, line 49: Do you know if this gets the status of the last completed run? or of runs in progress too? The workflow run endpoint gives a status and conclusion to differentiate this https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#check-statuses-and-conclusions. Not sure if this is a concern here.

dmitryax: Good question. I don't know. Looks like it should be _any_ of the last commit including in_progress. But let me print the fetched status so we can easy debug if needed

PR #1704, opened 2025-03-08
[chore] Use a k8s lease to avoid functional tests concurrency

dmitryax:
The Github actions concurrency feature allows only one queued job at a time, see [the docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#using-concurrency-in-different-scenarios). It means that the third EKS job is always cancelled, see https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/13731622293/job/38416102735. Even if we had 2 eks jobs, there would be race condition between several executions (PRs+main). But we need to ensure that jobs on the latest main are never cancelled https://github.com/signalfx/splunk-otel-collector-chart/pull/1703.

This change introduces another approach to handle concurrency - by using the k8s leases. It allows queueing many jobs on the same cluster. It also makes it easier to run test suites locally.


functional_tests/go.mod
```diff
@@ -3,21 +3,19 @@
 
 module github.com/signalfx/splunk-otel-collector-chart/functional_tests
 
-go 1.23.0
+go 1.24.0
```
dmitryax, line 6: New version is needed to use https://pkg.go.dev/testing#B.Context

functional_tests/functional/functional_test.go
```diff
@@ -61,6 +61,7 @@ const (
 	manifestsDir                           = "manifests"
 	eksValuesDir                           = "expected_eks_values"
 	kindValuesDir                          = "expected_kind_values"
+	helmUninstallTimeout                   = 5 * time.Minute
```
jinja2, line 64: I have no concerns adding this. But I think the default is 5mins already (at least when invoked from cli).

If you want it to wait for pods to be fully removed, try the DeletionPropagation option set to foreground instead.

dmitryax: I see. The problem with `Wait` https://github.com/helm/helm/blob/99c065789ef8c45bade24d4bc2d33432595de956/pkg/action/uninstall.go#L133-L139. We need to delete it or apply timeout. Otherwise, it's useless and confusing. Maybe we can rely on k8s client, but I've seen a lot of errors like [this](https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/13734405876/job/38416507267?pr=1704). So maybe we should set more than 5 minutes. BTW I've submitted a separate PR for this https://github.com/signalfx/splunk-otel-collector-chart/pull/1706

dmitryax: Ok, the error that I mentioned is because of a bug in [this line](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/.github/workflows/functional_test_v2.yaml#L186). `helm list | grep -q "^sock$"` is always false because `helm list` output is a table. The upgrade tests are pretty broken

dmitryax: I removed the timeout for now and solved the failures by moving installation of the initial (pre-upgrade) release in the go tests so the lease is acquired prior to that

dmitryax: @jinja2 looks like the timeout is actually needed. [This failure](https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/13796823642/job/38590881913?pr=1704) sounds like particularly caused by this

jinja2 (approved):

PR #1706, opened 2025-03-09
[chore] Set timeout for helm uninstall in tests

dmitryax:
Otherwise, the Wait option isn't applied. I've set it on couple the most important tests for now. Later, I'm going to consolidate the spinup/teardown code to avoid the duplication


dmitryax:
merged as part of another PR

PR #1710, opened 2025-03-11
signalfxexporter: send otlp histograms by default

asuresh4:
Set `send_otlp_histograms:true` on `signalfx` exporter in the agent and gateway config to export OTLP histograms without translations.


asuresh4:

dmitryax:
I've been asking @aunshc about it but haven't got the green light

atoulme:
I am going to close this PR for now as this change needs to be accompanied with a breaking change disclaimer, and requires alignment with our infra monitoring and ingest processing. We will do this in due time. In the mean time, the customer can turn this feature gate to true in their chart to get the functionality with native histograms.

PR #1720, opened 2025-03-13
[chore] fix formatting in values.schema.json that causes linter test failures for eks-addon releases

jvoravong:


dmitryax (approved):

dmitryax:
I belive there should be a linter that we can utilize here

jvoravong:
Yeah I was looking into an automated solution and testing them out. I think the pre-commit hooks added in these latest commits should do the trick. Let me know if you want something else, if not I'll merge this in tomorrow.

dmitryax:
SGTM as long as it does the job 👍

mstopa-splunk (approved):

PR #1734, opened 2025-03-19
[chore] Fix chart waiting for deployments in tests

dmitryax:
Use chart "wait" feature instead of custom logic

### DIFF
--- a/functional_tests/internal/chart.go
+++ b/functional_tests/internal/chart.go
@@ -46,6 +46,7 @@ func ChartInstallOrUpgrade(t *testing.T, testKubeConfig string, valuesFile strin
 	install := action.NewInstall(actionConfig)
 	install.Namespace = Namespace
 	install.ReleaseName = chartReleaseName
+	install.Wait = true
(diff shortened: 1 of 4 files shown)


jvoravong (approved):

functional_tests/internal/chart.go
```diff
@@ -46,6 +46,7 @@ func ChartInstallOrUpgrade(t *testing.T, testKubeConfig string, valuesFile strin
 	install := action.NewInstall(actionConfig)
 	install.Namespace = Namespace
 	install.ReleaseName = chartReleaseName
+	install.Wait = true
```
pszkamruk-splunk, line 49: just thinking loudly ;) 
do we want to set also `Timeout` here? in case of failures with installation 
`install.Timeout = time.Duration(120) * time.Second`

dmitryax: `Timeout` is already set on the next line :)

pszkamruk-splunk: ohh, did not notice that 
👍

PR #1737, opened 2025-03-25
[chore] configuration switching tests stabilisation

pszkamruk-splunk:
**Description:** tests stabilisation changes, recent improvements shown some flakiness in configuration switching tests, this PR adds additional logic which makes tests more stable.
Changes are addressing recent failures observed for following tests: 

- Test_Functions/logs_and_metrics_index_switch/non_default_source_type
- Test_Functions/cluster_receiver_enabled_or_disabled/check_cluster_receiver_disabled

### DIFF
--- a/functional_tests/configuration_switching/configuration_switching_test.go
+++ b/functional_tests/configuration_switching/configuration_switching_test.go
@@ -239,13 +239,13 @@ func testClusterReceiverEnabledOrDisabled(t *testing.T) {
 	logsObjectsHecEndpoint := fmt.Sprintf("http://%s:%d/services/collector", hostEp, internal.HECObjectsReceiverPort)
 
 	t.Run("check cluster receiver disabled", func(t *testing.T) {
-		internal.ResetLogsSink(t, logsObjectsConsumer)
 		replacements := map[string]interface{}{
 			"ClusterReceiverEnabled": false,
 			"LogObjectsHecEndpoint":  logsObjectsHecEndpoint,
 		}
 		deployChartsAndApps(t, valuesFileName, replacements)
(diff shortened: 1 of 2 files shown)


functional_tests/configuration_switching/configuration_switching_test.go
```diff
@@ -236,6 +245,7 @@ func testClusterReceiverEnabledOrDisabled(t *testing.T) {
 			"LogObjectsHecEndpoint":  logsObjectsHecEndpoint,
 		}
 		deployChartsAndApps(t, valuesFileName, replacements)
+		internal.WaitForTerminatingPods(t, clientset, internal.Namespace)
```
jinja2, line 248: Is this step for waiting on collector pod? Just wondering which pod could be terminating after the collector is installed (or is this an upgrade?) in the deployChartsAndApps.

pszkamruk-splunk: Yes, correct.
There was `ResetLogsSink` at the beginning but the better place for this is just after `WaitForTerminatingPods` - I moved it there
The second test has this (`ResetLogsSink`) in correct place

jinja2 (approved):

mstopa-splunk (approved):
lgtm with two optional enhancements to consider: 
- you can possibly extract clientset creation into a helper function, so we could do:
```
clientset := getKubeClientset(t)
```
both for `listPodsInNamespace` and `testClusterReceiverEnabledOrDisabled`

- you can add a comment to explain why we need `internal.WaitForTerminatingPods` because it was also unclear for me before I read your thread with jinja2

PR #1738, opened 2025-03-25
[chore] Add a troubleshooting tip for kubelet API TLS issues on local clusters

dmitryax:


jvoravong (approved):

jinja2 (approved):
Jfyi - another option would be to sign the certs with a cmd like [this](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/.github/workflows/functional_test_v2.yaml#L66). But updating chart config is probably more easier. 
I think AKS clusters also require this setting because they use a different CA to sign kubelet certs, but I see this is only for locally-running clusters.

pszkamruk-splunk:
What do you think about creating a `Troubleshooting` section on main page? Just like we have [advanced configuration](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/advanced-configuration.md)
Probably there will be more things which we can put there in the future.

pszkamruk-splunk (approved):

dmitryax:
> What do you think about creating a Troubleshooting section on main page? Just like we have [advanced configuration](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/advanced-configuration.md)
Probably there will be more things which we can put there in the future.

Added "Troubleshooting" header for this. (also removed somewhat confusing part which is unrelated to "Running locally" section)

PR #1756, opened 2025-04-08
Introduce image.otelcol.fips which set the appropriate repo/image and skips migrate-checkpoint if needed

dloucasfx:
**Description:** 
FIPS images do not support [migratecheckpoint](https://github.com/signalfx/splunk-otel-collector/blob/5bac0670a4cfe1c7e6f361b028721c719aa378d7/packaging/docker-otelcol.sh#L73) to migrate checkpoints from fluentd when `logsEnabled` is set , this causes the pod to get into a crashloop.
```
create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/migratecheckpoint": stat /migratecheckpoint: no such file or directory: unknown
```

The fix introduces a new ` image.otelcol.fips` option, which will be used to set the default repo/image and skips `migratecheckpoint` init container if needed.

The default repo logic is moved into `splunk-otel-collector.image.otelcol` helper function so user can still set a custom repo if needed.

### DIFF
--- a/.chloggen/fipsswitch.yaml
+++ b/.chloggen/fipsswitch.yaml
@@ -1,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: all
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Introduce a new option `image.otelcol.fips`, when set to true, it sets the appropriate repo/image and skips `migrate-checkpoint` when log is enabled
+# One or more tracking issues related to the change
+issues: [1756]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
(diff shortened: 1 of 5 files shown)


.chloggen/fipsswitch.yaml
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: all
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Introduce a new option `image.otelcol.fips`, when set to true, it sets the appropriate repo/image and skips `migrate-checkpoint` when log is enabled
```
jinja2, line 6: Are we officially rolling out our FIPs image? I am asking cause we have previously closed [similarly PRs](https://github.com/signalfx/splunk-otel-collector-chart/pull/1687) and have been waiting for green light from product.

dloucasfx: I was not aware of this PR, I wonder if this is the same user who reported it internally.
The workaround provided in the linked PR is not ideal as you will loose some functionalities by skipping all init containers.
As for the fips image support, this implies https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/values.yaml#L976 that it's supported.

dloucasfx:
We decided another approach, see https://github.com/signalfx/splunk-otel-collector-chart/pull/1757

PR #1757, opened 2025-04-08
Disable migrateCheckpoint and provide an option to enable it

dloucasfx:
**Description:** 
This PR disables migrateCheckpoint inticontainer and provides an option to enable if needed by setting `migrateCheckpoint` to true

### DIFF
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -186,6 +186,16 @@ splunkObservability:
 logsEngine: otel
 
 ################################################################################
+# Migrating Fluentd's position file to Otel checkpoints
+# Aimed for users who have been using fluentd as a logging agent but wants to
+# use OpenTelemetry Collector.
+# Enabling this option during the switch to the `logsEngine: otel` prevents double ingestion.
+# When set to true, an initContainer is added to the daemonset and runs the migration
+################################################################################
+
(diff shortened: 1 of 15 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -185,6 +185,15 @@ splunkObservability:
 
 logsEngine: otel
 
+################################################################################
+# Migrating Fluentd's position file to Otel checkpoints
+# Aimed for users who have been using fluentd as a logging agent but wants to
+# use OpenTelemetry Collector.
```
jinja2, line 191:
```suggestion
# use OpenTelemetry Collector.
# Enabling this option during the switch to the `logsEngine: otel` prevents double ingestion.
```
would it be useful or even correct to add a line as such?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -185,6 +185,15 @@ splunkObservability:
...
+# When set to true, an initContainer is added to the daemonset and runs the migration
+################################################################################
+
+migrateCheckpoint: false
```
dmitryax, line 195: Maybe pick a more descriptive name given that it's on the root level?
```suggestion
migrateLogsCheckpoints: false
```

dmitryax (approved):

PR #1759, opened 2025-04-09
Remove smartagent/signalfx-forwarder from the agent default config

dmitryax:
The old SignalFx instrumentation libraries are not supported anymore. So we don't need to keep the forwarder in the default config. Use Splunk OpenTelemetry instrumentation libraries to send traces in OTLP format instead.

If you need to keep the receiver for longer, add the following section to your values.yaml file:
```yaml
agent:
  ports:
    sfx-forwarder:
      containerPort: 9080
      hostPort: 9080
      protocol: TCP
      enabled_for: [traces]
  config:
    receivers:
      smartagent/signalfx-forwarder:
        listenAddress: 0.0.0.0:9080
        type: signalfx-forwarder
    service:
      pipelines:
        traces:
          receivers:
            - otlp
            - jaeger
            - smartagent/signalfx-forwarder
            - zipkin
```

### DIFF
--- a/.chloggen/remove-signalfx-forwarder.yaml
+++ b/.chloggen/remove-signalfx-forwarder.yaml
@@ -1,0 +1,8 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove smartagent/signalfx-forwarder from the agent default config.
+# One or more tracking issues related to the change
+issues: [1759]
--- a/examples/add-filter-processor/rendered_manifests/configmap-agent.yaml
+++ b/examples/add-filter-processor/rendered_manifests/configmap-agent.yaml
@@ -413,9 +413,6 @@ data:
         - k8s_observer
       signalfx:
         endpoint: 0.0.0.0:9943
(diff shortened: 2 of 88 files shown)


.chloggen/remove-signalfx-forwarder.yaml
```diff
@@ -0,0 +1,8 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove smartagent/signalfx-forwarder from the agent default config.
```
jinja2, line 6: Can we add the guidance in PR description to the subtext of changelog so the info is more visible for users?

dmitryax: I was thinking it isn't worth bloating the changelog given that it already has a link to PR, but I can add it if you think otherwise.

jinja2: Ah, that works too. I always want more info everywhere 😅

jinja2 (approved):

dloucasfx (commented):
PR looks good!
My only question is communicating this change and making the user aware of it.
Do we have customers pointing their instrumented apps to 9080?

dmitryax:
> Do we have customers pointing their instrumented apps to 9080?

Only if they use the old SignalFx instrumentation libraries which are not supported anymore

PR #1760, opened 2025-04-09
allow control plane metrics from all distro

jinja2:

This PR introduces support for receiver_creator configuration for control plane components across any distribution of Kubernetes clusters. This functionality is applicable when both featureGates.useControlPlaneMetricsHistogramData and the relevant components in agent.controlPlaneMetrics are set to true.

I tested these configs in EKS, AKS and GKE clusters and have a few findings -

1. EKS - coredns and kube-proxy, both receiver's rules match, and so if `featureGates.useControlPlaneMetricsHistogramData=true` the agent will collect metrics for both these components.
2. AKS - coredns rule matches AKS' coredns addon in the cluster but kube-proxy rule does not. The rule can be updated to match these by matching for `labels["component"] == "kube-proxy"`.
3. GKE - the default dns addon in GKE is not coredns. For `distribution = gke`, the receiver_creator config will include a `prometheus/kubedns` receiver which matches GKE's default kubedns addon pods. Another thing to note, kube-proxy rule which matches AKS also matches GKE but, I didn't find an exposed metric port for these.

Given the above findings and inconsistencies across managed clusters, I wonder if we want to proceed with enabling all control plane receivers? Should we include only the DNS receivers (coredns and kubedns) instead? 

### DIFF
--- a/.chloggen/controlplaneMetrics.yaml
+++ b/.chloggen/controlplaneMetrics.yaml
@@ -1,0 +1,23 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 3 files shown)


dmitryax (approved):
LGTM

PR #1761, opened 2025-04-09
Add useMemoryLimitPercentage feature gate

dmitryax:
Add `useMemoryLimitPercentage` feature gate that enables the use of the memory_limiter processor based on the percentage. It is required if the helm chart is used with another collector image from upstream.


jinja2 (approved):
I think we should re-evaluate the usage of container's memory limit as reference for these limit values since it can be significantly higher than what's guaranteed to be available (memory request).

PR #1780, opened 2025-04-21
[chore] Update workflow to include CRDs

jinja2:

The current workflow which updates the opentelemetry-operator version in our helm chart does not check for updates to CRDs. This PR introduces steps to check for CRD updates corresponding the operator app version we are pulling in during the opentelemetry-operator version update. This makes certain that our opentelemetry-operator-crds subchart is kept in sync.
The workflow will also bump up the patch version of the crds subchart when changes are made. The subchart CRD changes are NOT applied by helm on an upgrade. The explicit version bump is to improve visibility for users that CRDs have changed and they'd want to upgrade the same in their clusters outside of helm.  

### DIFF
--- a/.github/workflows/update_chart_dependencies.yaml
+++ b/.github/workflows/update_chart_dependencies.yaml
@@ -43,6 +43,11 @@ jobs:
 
           make update-chart-dep CHART_PATH=${{ matrix.yaml_file_path }} SUBCHART_NAME='${{ matrix.dependency_name }}' DEBUG_MODE=$DEBUG_MODE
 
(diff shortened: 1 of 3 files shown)


jvoravong (approved):
LGTM, this is a great start.

For future improvements we should keep these in mind:
- Consider refactoring some of the new Bash commands added in this PR to leverage using base_utils.sh more.
- We should check if the operator repo has tests for proper CRD versioning. I'll check on this to make sure we prevent cases where required CRD version bumps were missed upstream.

PR #1792, opened 2025-04-22
[chore] add logic to update our tests k8s versions

dloucasfx:
Added a utility to get a list of supported vanilla k8s clusters, compare it to our test k8s versions and update when necessary.

I added a make target, which we can keep or remove, but the end goal is to add a gh action that uses this logic to create a PR with the changes if detected.

We currently have a bunch of EOL k8s versions that we use, it's a way to check compatibility with cloud providers that extend the support. 
This is not ideal as we keep increasing our test matrix and the vanilla cluster is not identical to the one used by the cloud provider. To test for schema compatibility, we should start using `kubeconform`

To Do (sperate PRs):
- Add a gh action
- track EKS, AKS and GKE

unit test

### DIFF
--- a/.github/workflows/update_k8s_versions.yaml
+++ b/.github/workflows/update_k8s_versions.yaml
@@ -1,0 +1,47 @@
+name: Check for new matrix versions and update tests if needed
+
+on:
+  schedule:
+    # Run every Monday at noon.
+    - cron: "0 12 * * 1"
+  workflow_dispatch:
+    inputs:
+      DEBUG_ARGUMENT:
+        description: 'Enable debug by setting -debug to true'
+        required: false
+        default: '-debug=false'
+
+jobs:
+  check_and_update:
+    runs-on: ubuntu-latest
+    env:
+      DEBUG: ${{ github.event.inputs.DEBUG_ARGUMENT }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Check for new matrix versions
+        id: check_for_update
+        run: |
+          echo "Checking for new matrix versions"
+          make update-matrix-versions DEBUG=$DEBUG
+
+      - name: check for changes
+        id: git-check
+        run: |
+          if git diff --quiet; then
+            echo "No changes detected, exiting workflow successfully"
+            exit 0
+          fi
+          echo "changes=true" >> $GITHUB_OUTPUT
+
+      - name: Open PR for matrix version update
+        if: steps.git-check.outputs.changes == 'true'
+        uses: peter-evans/create-pull-request@v7
+        with:
+          commit-message: Update matrix test versions
+          title: Update matrix versions used for testing
+          body: Use latest supported matrix versions
+          branch: update-matrix-test-versions
+          base: main
+          delete-branch: true
--- a/tools/k8s_versions/update_k8s_versions.go
+++ b/tools/k8s_versions/update_k8s_versions.go
@@ -1,0 +1,288 @@
+// Copyright Splunk Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+)
+
+var debug bool
+
+const (
+	EndOfLifeURL        string = "https://endoflife.date/api/kubernetes.json"
+	KindDockerHubURL    string = "https://hub.docker.com/v2/repositories/kindest/node/tags?page_size=1&page=1&ordering=last_updated&name="
+	MiniKubeURL         string = "https://raw.githubusercontent.com/kubernetes/minikube/master/pkg/minikube/constants/constants_kubernetes_versions.go"
+	KubeKindVersion     string = "k8s-kind-version"
+	KubeMinikubeVersion string = "k8s-minikube-version"
+)
+
+type KubernetesVersion struct {
+	Cycle       string `json:"cycle"`
+	ReleaseDate string `json:"releaseDate"`
+	EOLDate     string `json:"eol"`
+	Latest      string `json:"latest"`
+}
+
+type DockerImage struct {
+	Count int `json:"count"`
+}
+
+// getSupportedKubernetesVersions returns the supported Kubernetes versions
+// by checking the EOL date of the collected versions.
+func getSupportedKubernetesVersions(url string) ([]KubernetesVersion, error) {
+	body, err := getRequestBody(url)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get k8s versions: %w", err)
+	}
+	var kubernetesVersions, supportedKubernetesVersions []KubernetesVersion
+	if err = json.Unmarshal(body, &kubernetesVersions); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal JSON: %w", err)
+	}
+
+	now := time.Now()
+	for _, kubernetesVersion := range kubernetesVersions {
+		eolDate, parseErr := time.Parse(time.DateOnly, kubernetesVersion.EOLDate)
+		if parseErr != nil {
+			return nil, fmt.Errorf("error parsing date: %w", parseErr)
+		}
+		if eolDate.After(now) {
+			supportedKubernetesVersions = append(supportedKubernetesVersions, kubernetesVersion)
+		} else {
+			logDebug("Skipping version %s, EOL date %s", kubernetesVersion.Cycle, kubernetesVersion.EOLDate)
+		}
+	}
+	return supportedKubernetesVersions, nil
+}
+
+// getLatestSupportedMinikubeVersions iterates through the K8s supported versions and find the latest minikube after parsing
+// the sorted ValidKubernetesVersions slice from constants_kubernetes_versions.go
+func getLatestSupportedMinikubeVersions(url string, k8sVersions []KubernetesVersion) ([]string, error) {
+	body, err := getRequestBody(url)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get minikube versions: %w", err)
+	}
+
+	// Extract the slice using a regular expression
+	re := regexp.MustCompile(`ValidKubernetesVersions = \[\]string{([^}]*)}`)
+	matches := re.FindStringSubmatch(string(body))
+	if len(matches) < 2 {
+		return nil, errors.New("minikube, failed to find the Kubernetes versions slice")
+	}
+
+	// Parse and cleanup the slice values
+	minikubeVersions := strings.Split(strings.NewReplacer("\n", "", `"`, "", "\t", "", " ", "").Replace(matches[1]), ",")
+
+	logDebug("Found minikube versions: %s", minikubeVersions)
+
+	var latestMinikubeVersions []string
+	// the minikube version slice is sorted, break when first cycle match is found
+	for _, k8sVersion := range k8sVersions {
+		for _, minikubeVersion := range minikubeVersions {
+			if strings.Contains(minikubeVersion, k8sVersion.Cycle) {
+				latestMinikubeVersions = append(latestMinikubeVersions, minikubeVersion)
+				break
+			}
+		}
+	}
+
+	return latestMinikubeVersions, nil
+}
+
+// getLatestSupportedKindImages iterates through the K8s supported versions and find the latest kind
+// tag that supports that version
+func getLatestSupportedKindImages(url string, k8sVersions []KubernetesVersion) ([]string, error) {
+	var supportedKindVersions []string
+	for _, k8sVersion := range k8sVersions {
+		tag := k8sVersion.Latest
+		for {
+			exists, err := imageTagExists(url, tag)
+			if err != nil {
+				return supportedKindVersions, fmt.Errorf("failed to check image tag existence: %w", err)
+			}
+			if exists {
+				supportedKindVersions = append(supportedKindVersions, "v"+tag)
+				break
+			}
+			tag, err = decrementMinorMinorVersion(tag)
+			if err != nil {
+				// It's possible that kind still does not have a tag for new versions, break the loop and
+				// process other k8s versions
+				if strings.Contains(err.Error(), "minor version cannot be decremented below 0") {
+					logDebug("No kind image found for k8s version %s", k8sVersion.Cycle)
+					break
+				}
+				return supportedKindVersions, fmt.Errorf("failed to decrement k8sVersion: %w", err)
+			}
+		}
+	}
+	return supportedKindVersions, nil
+}
+
+func imageTagExists(url string, tag string) (bool, error) {
+	body, err := getRequestBody(url + tag)
+	if err != nil {
+		return false, fmt.Errorf("failed to get image tag: %w", err)
+	}
+
+	var kindImage DockerImage
+	if err = json.Unmarshal(body, &kindImage); err != nil {
+		return false, fmt.Errorf("failed to unmarshal JSON: %w", err)
+	}
+
+	if kindImage.Count > 0 {
+		return true, nil
+	}
+	return false, nil
+}
+
+func decrementMinorMinorVersion(version string) (string, error) {
+	parts := strings.Split(version, ".")
+	if len(parts) < 3 {
+		return "", fmt.Errorf("version does not have a minor version: %s", version)
+	}
+
+	minor, err := strconv.Atoi(parts[2])
+	if err != nil {
+		return "", fmt.Errorf("invalid minor version: %s", parts[1])
+	}
+
+	if minor == 0 {
+		return "", errors.New("minor version cannot be decremented below 0")
+	}
+
+	parts[2] = strconv.Itoa(minor - 1)
+	return strings.Join(parts, "."), nil
+}
+
+func updateMatrixFile(filePath string, kindVersions []string, minikubeVersions []string) error {
+	content, err := os.ReadFile(filePath)
+	if err != nil {
+		return fmt.Errorf("failed to read file: %w", err)
+	}
+
+	var testMatrix map[string]map[string][]string
+	if err = json.Unmarshal(content, &testMatrix); err != nil {
+		return fmt.Errorf("failed to unmarshal JSON: %w", err)
+	}
+
+	for _, value := range testMatrix {
+		if len(kindVersions) > 0 && value[KubeKindVersion] != nil {
+			value[KubeKindVersion] = kindVersions
+		} else if len(minikubeVersions) > 0 && value[KubeMinikubeVersion] != nil {
+			value[KubeMinikubeVersion] = minikubeVersions
+		}
+	}
+	// Marshal the updated test matrix back to JSON
+	updatedContent, err := json.MarshalIndent(testMatrix, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal updated JSON: %w", err)
+	}
+
+	// Ensure the file ends with a new line to make the pre-commit check happy
+	updatedContent = append(updatedContent, '\n')
+
+	if err = os.WriteFile(filePath, updatedContent, 0o644); err != nil {
+		return fmt.Errorf("failed to write updated file: %w", err)
+	}
+	return nil
+}
+
+func sortVersions(versions []string) {
+	sort.Slice(versions, func(i, j int) bool {
+		vi := strings.Split(versions[i][1:], ".") // Remove "v" and split by "."
+		vj := strings.Split(versions[j][1:], ".")
+
+		for k := 0; k < len(vi) && k < len(vj); k++ {
+			if vi[k] != vj[k] {
+				return vi[k] > vj[k] // Sort in descending order
+			}
+		}
+		return len(vi) > len(vj)
+	})
+}
+
+func getRequestBody(url string) ([]byte, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch URL: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code: %d", resp.StatusCode)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response body: %w", err)
+	}
+	return body, nil
+}
+
+func logDebug(format string, v ...any) {
+	if debug {
+		log.Printf(format, v...)
+	}
+}
+
+func main() {
+	// setup logging
+	flag.BoolVar(&debug, "debug", false, "Enable debug logging")
+	flag.Parse()
+	log.SetOutput(os.Stdout)
+	log.SetFlags(log.LstdFlags | log.Lshortfile)
+
+	k8sVersions, err := getSupportedKubernetesVersions(EndOfLifeURL)
+	if err != nil || len(k8sVersions) == 0 {
+		log.Fatalf("Failed to get k8s versions: %v", err)
+	}
+	logDebug("Found supported k8s versions %v", k8sVersions)
+
+	kindVersions, err := getLatestSupportedKindImages(KindDockerHubURL, k8sVersions)
+	if err != nil {
+		log.Printf("failed to get all kind versions: %v", err)
+	}
+	if len(kindVersions) > 0 {
+		// needs to be sorted so we don't end up with false positive diff in the json matrix file
+		sortVersions(kindVersions)
+		logDebug("Found supported kind images: %v", kindVersions)
+	}
+
+	minikubeVersions, err := getLatestSupportedMinikubeVersions(MiniKubeURL, k8sVersions)
+	if err != nil {
+		log.Printf("failed to get minikube versions: %v", err)
+	}
+	if len(minikubeVersions) > 0 {
+		logDebug("Found supported minikube versions: %v", minikubeVersions)
+	}
+
+	if len(kindVersions) == 0 && len(minikubeVersions) == 0 {
+		log.Fatalf("No supported versions found. Run with -debug=true for more info.")
+	}
+
+	path := "ci-matrix.json"
+	currentDir, err := os.Getwd()
+	if err != nil {
+		log.Fatalf("Failed to get current directory: %v ", err)
+	}
+	path = filepath.Join(currentDir, filepath.Clean(path))
+	err = updateMatrixFile(path, kindVersions, minikubeVersions)
+	if err != nil {
+		log.Fatalf("Failed to update matrix file: %v", err)
+	}
+	os.Exit(0)
+}
--- a/tools/k8s_versions/update_k8s_versions.go
+++ b/tools/k8s_versions/update_k8s_versions.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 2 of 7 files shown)


tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,142 @@
...
+		for i, line := range versionBlock {
+			trimmed := strings.TrimSpace(line)
+			if strings.HasPrefix(trimmed, fmt.Sprintf("- v%s", version.Cycle)) {
+				if eolDate.After(now) {
+					output = append(output, line)
+				}
+				break
+			} else if i == len(versionBlock)-1 && eolDate.After(now) {
+				// Add the new version at the end of the list
+				output = append(output, fmt.Sprintf("%s- v%s # EOL %s", strings.Repeat(" ", len(line)-len(trimmed)), version.Latest, version.EOLDate))
```
jinja2, line 118: We are appending to the list of kind cluster versions we run functional tests against, right? If that's the case we should confirm that kind has an image out for the new version we are adding here or we'll have errors when we try to start in the kind cluster.

dloucasfx: Good point @jinja2 . I will add some checks and if not available go down to previous minor version

dloucasfx: PR updated; we now check docker hub if the image exists and decrement the minor minor until we find one

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,222 @@
+package main
```
pjanotti, line 1: nit - I'm not sure about this but other go sources in the repo have a header. That said there is no check enforcing that....

pjanotti:
```suggestion
// Copyright Splunk Inc.
// SPDX-License-Identifier: Apache-2.0

package main
```

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,222 @@
...
+	parts[2] = strconv.Itoa(minor - 1)
+	return strings.Join(parts, "."), nil
+}
+
+func main() {
+	url := "https://endoflife.date/api/kubernetes.json"
+	k8sVersions, err := fetchKubernetesVersions(url)
+	if err != nil {
+		fmt.Println("Failed to get k8s versions:", err)
+		return
```
pjanotti, line 211: It will be good to set the exit code in this case so make and other tools could be notified of the failure.

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	filesToUpdate := []string{".github/workflows/functional_test.yaml", ".github/workflows/functional_test_v2.yaml"}
```
pjanotti, line 214: nit: nice to have these as parameters, but, it can be as it is too

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,222 @@
...
+		// Locate the k8s-version block needed to update
+		if strings.HasPrefix(trimmed, "k8s-version:") || strings.HasPrefix(trimmed, "kubernetes_version:") {
+			inK8sVersionBlock = true
+			output = append(output, line)
+			continue
+		}
+
+		if inK8sVersionBlock {
+			if strings.HasPrefix(trimmed, "-") || strings.HasPrefix(trimmed, "#") {
+				k8sVersionBlock = append(k8sVersionBlock, line)
```
pjanotti, line 79: nit: a comment describing that here the code is collecting all entries in the list of version or some to that effect

pjanotti (commented):
Nice!

tools/k8s_versions/update_k8s_versions.go
dloucasfx: makes sense!

tools/k8s_versions/update_k8s_versions.go
dloucasfx: Makes sense, I will be using
0 // success
1 // failed
2 // partial success. ie some files got update

tools/k8s_versions/update_k8s_versions.go
dloucasfx: I am planning on expanding this to include EKS, AKS and GKE with an option to choose what cloud provider list to update and I will also add the option to pass the list of yaml file to update and the name of the array element where the versions exist. 
But this will be in the next version

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,238 @@
...
+					return nil, fmt.Errorf("failed to get available kind image: %w", err)
+				}
+				output = append(output, fmt.Sprintf("%s- v%s # EOL %s", strings.Repeat(" ", len(line)-len(trimmed)), tag, version.EOLDate))
+			}
+		}
+	}
+	return output, nil
+}
+
+func getAvailableKindImage(tag string) (string, error) {
```
crobert-1, line 145: This logic is pretty hard for me to follow. It looks like for every k8s version we're getting an available kind image. Is there some way to iterate over the kind images supported, rather than manually doing string operations to deprecate versions and check if they're valid? If we [ordered results simply by name](https://hub.docker.com/v2/repositories/kindest/node/tags?=page=1&ordering=name), we'd be able to just iterate directly checking if the k8s version exists as a kind version, and if not, keep iterating. We could also use [go-version](https://pkg.go.dev/github.com/hashicorp/go-version) to handle some of the versioning string/regex functionality comparing values if it's useful.

crobert-1 (commented):
Looks good! With all of my comments feel free to drop or clarify as you see fit. I'm generally trying to reduce regexes and string parsing logic as much as possible. I'd be happy to discuss any options here.

.github/workflows/update_k8s_versions.yaml
```diff
@@ -0,0 +1,41 @@
+name: Check for new k8s versions and update tests if needed
+
+on:
+  pull_request:
+#schedule:
+    # Run every Monday at noon.
+  #  - cron: "0 12 * * 1"
+  #workflow_dispatch:
```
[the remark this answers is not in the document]
dloucasfx: correct, I was just testing it on pull_request, I'll update it soon

.github/workflows/update_k8s_versions.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Check for new k8s versions
+        id: check_for_update
+        run: |
+          echo "Checking for new k8s versions"
+          make update-k8s-test-versions
+
+      - name: check for changes
+        id: git-check
+        run: |
+          if git diff --quiet; then
```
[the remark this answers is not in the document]
dloucasfx: `--exit-code` will show the diff , which is not what we need

`--quiet` prevents the diff from showing and exits with 1 if there was a diff, 0 if not
`--exit-code` will exit with 1 if there’s a diff and 0 if not but it’ll show the diff

.github/workflows/update_k8s_versions.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Open PR for K8S Version Update
+        if: steps.git-check.outputs.changes == 'true'
+        uses: peter-evans/create-pull-request@v7
+        with:
+          commit-message: Update K8S test versions
+          title: Update K8S versions used for testing
+          body: Use latest supported k8s kind versions
+          branch: update-k8s-test-versions
+          base: main
+          delete-branch: true
```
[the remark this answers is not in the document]
dloucasfx: I don't want to deal with tokens/secret and all the extra steps and logic that `create-pull-request` takes care of for you. In addition to the token , the PR branch management https://github.com/marketplace/actions/create-pull-request
We also widely use this action in other workflows.

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,238 @@
...
+	var k8sVersionBlock []string
+	var inK8sVersionBlock bool
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+		trimmed := strings.TrimSpace(line)
+
+		// Locate the k8s-version block needed to update
+		if strings.HasPrefix(trimmed, "k8s-version:") || strings.HasPrefix(trimmed, "kubernetes_version:") {
```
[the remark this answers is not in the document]
dloucasfx: In the coming iteration, I had to move the the matrix out of the workflow.
I will update the PR shortly and please review again, would love to hear your feedback

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,238 @@
...
+		return fmt.Errorf("failed to read file: %w", err)
+	}
+
+	if err := os.WriteFile(filePath, []byte(strings.Join(output, "\n")+"\n"), 0644); err != nil {
+		return fmt.Errorf("failed to write updated file: %w", err)
+	}
+	return nil
+}
+
+func updateVersionsBlock(versions []KubernetesVersion, versionBlock []string) ([]string, error) {
```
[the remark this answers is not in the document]
dloucasfx: this was drastically reduce in the next iteration,  I will update the PR shortly and please review again, would love to hear your feedback

tools/k8s_versions/update_k8s_versions.go
dloucasfx: This was also drastically simplified in the next iteration.
But the idea of getting the kind images first is not ideal, as you will also have to do with paging and large payloads as there are many images/tags for different platform etc...
The idea here and still in the next iteration, that we have a list of "currently" supported k8s versions, with latest and cycle (those are normally 3 or 4 max versions), we iterate through the latest and decrement it until we find an existing image on docker hub and use that tag.

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,238 @@
...
+		fmt.Println("Failed to get current directory:", err)
+		os.Exit(1)
+	}
+
+	var partialFailure bool
+	for _, filePath := range filesToUpdate {
+		filePath = filepath.Join(currentDir, filepath.Clean(filePath))
+		if err := updateK8sVersion(filePath, k8sVersions); err != nil {
+			fmt.Println("Error:", err)
+			partialFailure = true
```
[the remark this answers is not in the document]
dloucasfx: I will add this! Thanks!

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,238 @@
...
+		return "", fmt.Errorf("minor version cannot be decremented below 0")
+	}
+
+	parts[2] = strconv.Itoa(minor - 1)
+	return strings.Join(parts, "."), nil
+}
+
+func main() {
+	url := "https://endoflife.date/api/kubernetes.json"
+	k8sVersions, err := fetchKubernetesVersions(url)
```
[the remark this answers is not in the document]
dloucasfx: good point

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,238 @@
...
+		tag, err = decrementMinorMinorVersion(tag)
+		if err != nil {
+			return "", fmt.Errorf("failed to decrement version: %w", err)
+		}
+	}
+}
+
+func imageTagExists(tag string) (bool, error) {
+	url := "https://hub.docker.com/v2/repositories/kindest/node/tags?page_size=1&page=1&ordering=last_updated&name="
+	resp, err := http.Get(url + tag)
```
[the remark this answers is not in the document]
dloucasfx: fixed in the next iteration

tools/k8s_versions/update_k8s_versions.go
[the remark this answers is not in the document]
dloucasfx: On a second thought, I need to pass the URL for unit tests when using a mock server

.github/workflows/update_k8s_versions.yaml
```diff
@@ -0,0 +1,41 @@
...
+          make update-k8s-test-versions
+
+      - name: check for changes
+        id: git-check
+        run: |
+          if git diff --quiet; then
+            echo "No changes detected, exiting workflow successfully"
+            exit 0
+          fi
+          echo "changes=true" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
dloucasfx: I am not sure I follow, but I am also a beginner when it comes to gh actions

tools/k8s_versions/update_k8s_versions.go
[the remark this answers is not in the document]
dloucasfx: not needed with new iteration

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+func updateMatrixFile(filePath string, kindVersions []string, minikubeVersions []string) error {
+	content, err := os.ReadFile(filePath)
+	if err != nil {
+		return fmt.Errorf("failed to read file: %w", err)
+	}
+
+	var testMatrix map[string]map[string][]string
+	if err := json.Unmarshal(content, &testMatrix); err != nil {
```
crobert-1, line 175: Same as other comment `err` ends up getting overshadowed here, we should use `=` instead through the rest of the method.

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,285 @@
...
+		return fmt.Errorf("failed to unmarshal JSON: %w", err)
+	}
+
+	for _, value := range testMatrix {
+		if len(kindVersions) > 0 && value["k8s-kind-version"] != nil {
+			value["k8s-kind-version"] = kindVersions
```
crobert-1, line 181: I'd prefer to have the json keys as constants that are reused rather than plain strings, just in case of typos or future changes.

.github/workflows/update_k8s_versions.yaml
```diff
@@ -0,0 +1,47 @@
...
+on:
+  schedule:
+    # Run every Monday at noon.
+    - cron: "0 12 * * 1"
+  workflow_dispatch:
+    inputs:
+      DEBUG_ARGUMENT:
+        description: 'Enable debug by setting -debug to true'
+        required: false
+        default: '-debug=false'
```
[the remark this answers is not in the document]
dloucasfx: Right, you can run the debug locally. But this is just to give the user an option to debug the gh action in case (I can't think of one now, maybe when extra logic is added later on), you can't reproduce locally.

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response body: %w", err)
+	}
+	return body, nil
+}
+
+func logDebug(format string, v ...interface{}) {
```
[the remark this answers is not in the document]
dloucasfx: Yeah, I was considering zap package , but for a small tool like this, the standard package log would the job .
If it's ok with you, I would like to stick with log.

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,285 @@
...
+		return nil, fmt.Errorf("failed to get k8s versions: %w", err)
+	}
+	var kubernetesVersions, supportedKubernetesVersions []KubernetesVersion
+	if err := json.Unmarshal(body, &kubernetesVersions); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal JSON: %w", err)
+	}
+
+	now := time.Now()
+	for _, kubernetesVersion := range kubernetesVersions {
+		eolDate, err := time.Parse("2006-01-02", kubernetesVersion.EOLDate)
```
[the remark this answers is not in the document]
dloucasfx: sounds good, I can do that!

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,285 @@
...
+
+	// Extract the slice using a regular expression
+	re := regexp.MustCompile(`ValidKubernetesVersions = \[\]string{([^}]*)}`)
+	matches := re.FindStringSubmatch(string(body))
+	if len(matches) < 2 {
+		return nil, fmt.Errorf("minikube, failed to find the Kubernetes versions slice")
+	}
+
+	// Parse and cleanup the slice values
+	minikuveVersions := strings.Split(strings.ReplaceAll(strings.ReplaceAll(matches[1], "\n", ""), `"`, ""), ",")
```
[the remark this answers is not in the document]
dloucasfx: yuck :) , good catch!

.github/workflows/update_k8s_versions.yaml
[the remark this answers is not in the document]
dloucasfx: @crobert-1 
Thanks, I see now, you want me to use the outcome property. 
The reason I have not followed this approach, because exiting with non-zero will mark the step as failed, and of course, still need to add `continue-on-error: true`

example:
 ````
       - name: check for changes
        id: git-check
        continue-on-error: true
        run: |
          if git diff --quiet; then
            echo "No changes detected, exiting workflow successfully"
            exit 0
          fi
          exit 1

      - name: Open PR for matrix version update
        if: steps.git-check.outcome == 'failure'
        run: |
          echo "create PR logic"
````

The exit 1 is needed to differentiate the states (change vs no-change).  (or simply replace the run with `git diff --quiet` , but still exit 1 applies)

If someone not familiar with how those steps work in gh action, looking at the log, you would think that the step has failed and this can be confusing
````
[Check for new matrix versions and update tests if needed/check_and_update] ⭐ Run Main check for changes
[Check for new matrix versions and update tests if needed/check_and_update]   🐳  docker exec cmd=[bash -e /var/run/act/workflow/git-check] user= workdir=
[Check for new matrix versions and update tests if needed/check_and_update] Failed but continue next step
[Check for new matrix versions and update tests if needed/check_and_update]   ❌  Failure - Main check for changes [127.233375ms]
[Check for new matrix versions and update tests if needed/check_and_update] ⭐ Run Main Open PR for matrix version update
[Check for new matrix versions and update tests if needed/check_and_update]   🐳  docker exec cmd=[bash -e /var/run/act/workflow/3] user= workdir=
| create PR logic
[Check for new matrix versions and update tests if needed/check_and_update]   ✅  Success - Main Open PR for matrix version update [69.318834ms]
[Check for new matrix versions and update tests if needed/check_and_update] ⭐ Run Complete job
[Check for new matrix versions and update tests if needed/check_and_update] Cleaning up container for job check_and_update
[Check for new matrix versions and update tests if needed/check_and_update]   ✅  Success - Complete job
[Check for new matrix versions and update tests if needed/check_and_update] 🏁  Job succeeded
````

In short, `steps.git-check.outcome` can create confusion as the step will be marked as failed and the current logic avoid this behavior

File: tools/k8s_versions/update_k8s_versions.go [hunk 22]
```diff
@@ -0,0 +1,285 @@
...
+
+// getSupportedKubernetesVersions returns the supported Kubernetes versions
+// by checking the EOL date of the collected versions.
+func getSupportedKubernetesVersions(url string) ([]KubernetesVersion, error) {
+	body, err := getRequestBody(url)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get k8s versions: %w", err)
+	}
+	var kubernetesVersions, supportedKubernetesVersions []KubernetesVersion
+	if err := json.Unmarshal(body, &kubernetesVersions); err != nil {
```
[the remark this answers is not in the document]
dloucasfx: good catch!

.github/workflows/update_k8s_versions.yaml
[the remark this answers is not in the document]
crobert-1: Okay, thanks for clarifying! Agreed that's not what we want, and I appreciate the job output too 👍

tools/k8s_versions/update_k8s_versions.go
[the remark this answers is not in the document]
dloucasfx: I run `strings.TrimSpace(minikubeVersion)` when I build the returned slice.
But I guess I cam move it to this code as we print the output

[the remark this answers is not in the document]
dloucasfx: Ended up using replacer , which is cleaner to read

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,287 @@
...
+		return nil, fmt.Errorf("failed to get k8s versions: %w", err)
+	}
+	var kubernetesVersions, supportedKubernetesVersions []KubernetesVersion
+	if err = json.Unmarshal(body, &kubernetesVersions); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal JSON: %w", err)
+	}
+
+	now := time.Now()
+	for _, kubernetesVersion := range kubernetesVersions {
+		eolDate, err := time.Parse(time.DateOnly, kubernetesVersion.EOLDate)
```
crobert-1, line 57: This is another shadowing error

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+func imageTagExists(url string, tag string) (bool, error) {
+	body, err := getRequestBody(url + tag)
+	if err != nil {
+		return false, fmt.Errorf("failed to get image tag: %w", err)
+	}
+
+	var kindImage DockerImage
+	if err := json.Unmarshal(body, &kindImage); err != nil {
```
crobert-1, line 141: This is another shadowing error

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,287 @@
...
+
+	for _, value := range testMatrix {
+		if len(kindVersions) > 0 && value[KubeKindVersion] != nil {
+			value[KubeKindVersion] = kindVersions
+		} else if len(minikubeVersions) > 0 && value[KubeMinikubeVersion] != nil {
+			value[KubeMinikubeVersion] = minikubeVersions
+		}
+	}
+	// Marshal the updated test matrix back to JSON
+	updatedContent, err := json.MarshalIndent(testMatrix, "", "  ")
```
crobert-1, line 189: This is a shadowing error

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch URL: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code: %d", resp.StatusCode)
+	}
+
+	body, err := io.ReadAll(resp.Body)
```
crobert-1, line 228: This is a shadowing error

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,287 @@
...
+	log.SetOutput(os.Stdout)
+	log.SetFlags(log.LstdFlags | log.Lshortfile)
+
+	k8sVersions, err := getSupportedKubernetesVersions(EndOfLifeURL)
+	if err != nil || len(k8sVersions) == 0 {
+		log.Fatalf("Failed to get k8s versions: %v", err)
+	}
+	logDebug("Found supported k8s versions %v", k8sVersions)
+
+	kindVersions, err := getLatestSupportedKindImages(KindDockerHubURL, k8sVersions)
```
crobert-1, line 254: This is a shadowing error

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		log.Printf("failed to get all kind versions: %v", err)
+	}
+	if len(kindVersions) > 0 {
+		// needs to be sorted so we don't end up with false positive diff in the json matrix file
+		sortVersions(kindVersions)
+		logDebug("Found supported kind images: %v", kindVersions)
+	}
+
+	minikubeVersions, err := getLatestSupportedMinikubeVersions(MiniKubeURL, k8sVersions)
```
crobert-1, line 264: This is a shadowing error

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,287 @@
...
+	if len(minikubeVersions) > 0 {
+		logDebug("Found supported minikube versions: %v", minikubeVersions)
+	}
+
+	if len(kindVersions) == 0 && len(minikubeVersions) == 0 {
+		log.Fatalf("No supported versions found. Run with -debug=true for more info.")
+	}
+
+	path := "ci-matrix.json"
+	currentDir, err := os.Getwd()
```
crobert-1, line 277: This is a shadowing error

crobert-1 (commented):
Sorry about all of the nits, once the formatting tools are introduced to the repo it'll catch earlier on in the process.

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,287 @@
...
+
+	minikubeVersions, err := getLatestSupportedMinikubeVersions(MiniKubeURL, k8sVersions)
+	if err != nil {
+		log.Printf("failed to get minikube versions: %v", err)
+	}
+	if len(minikubeVersions) > 0 {
+		logDebug("Found supported minikube versions: %v", minikubeVersions)
+	}
+
+	if len(kindVersions) == 0 && len(minikubeVersions) == 0 {
```
dmitryax, line 272: Shouldn't we fail if any of them is empty not both?

dloucasfx: I figured if one provider failed (ex: the endpoint was not reachable), we should not sacrifice the other

dloucasfx:
New push:
- Addressed most of the review entries. For var shadowing, I only addressed the linter errors as the others are not necessary
- Updated the golangci config to force strict shadow detection and excluded this tool from gosec to prevent unnecessary detections 
```
tools/k8s_versions/update_k8s_versions.go:197:11: G306: Expect WriteFile permissions to be 0600 or less (gosec)
        if err = os.WriteFile(filePath, updatedContent, 0644); err != nil {
                 ^
tools/k8s_versions/update_k8s_versions.go:218:15: G107: Potential HTTP request made with variable url (gosec)
        resp, err := http.Get(url)
      
```
-  Rebased and resolved conflicts

crobert-1:
Once this PR has been rebased on `main` we can drop all of my `shadowing err` issues. `main` now has lint checks running on PRs so once lint passes on this PR my concerns are resolved 👍

dloucasfx:
> Once this PR has been rebased on `main` we can drop all of my `shadowing err` issues. `main` now has lint checks running on PRs so once lint passes on this PR my concerns are resolved 👍

@crobert-1 
Pushed the following changes:
- rebased
- Added Makefile in the directory to reference the common makefile
- Addressed @dmitryax comment https://github.com/signalfx/splunk-otel-collector-chart/pull/1792#discussion_r2072150118
- Fixed gosec G107 by using url parse
```
tools/k8s_versions/update_k8s_versions.go:220:15: G107: Potential HTTP request made with variable url (gosec)
        resp, err := http.Get(url)
```
- The only lint that is failing now is 
```
tools/k8s_versions/update_k8s_versions.go:199:11: G306: Expect WriteFile permissions to be 0640 or less (gosec)
        if err = os.WriteFile(filePath, updatedContent, 0o644); err != nil {
```
This is very restrictive, not if we want to no lint it or change the G306 gosec config to 
```
    config:
      # Maximum allowed permissions mode for os.OpenFile and os.Chmod
      # Default: "0600"
      G306: "0644"
```

crobert-1:
> This is very restrictive, not if we want to no lint it or change the G306 gosec config to

I'd say for now my preference is to just add a nolint comment. I generally prefer getting warnings up front on new changes to let us make an educated decision on a case by case basis. Happy to hear from others though 👍

crobert-1 (approved):
LGTM once CI is passing. Thanks for all of your work here @dloucasfx!

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,292 @@
...
+	logDebug("Found supported kind images: %v", kindVersions)
+
+	minikubeVersions, err := getLatestSupportedMinikubeVersions(miniKubeURL, k8sVersions)
+	if err != nil {
+		log.Fatalf("failed to get minikube versions: %v", err)
+	}
+
+	logDebug("Found supported minikube versions: %v", minikubeVersions)
+
+	if len(kindVersions) == 0 || len(minikubeVersions) == 0 {
```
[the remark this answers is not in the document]
dloucasfx: done

tools/k8s_versions/update_k8s_versions.go
```diff
@@ -0,0 +1,294 @@
...
+		return false, fmt.Errorf("failed to unmarshal JSON: %w", err)
+	}
+
+	if kindImage.Count > 0 {
+		return true, nil
+	}
+	return false, nil
+}
+
+func decrementMinorMinorVersion(version string) (string, error) {
```
dmitryax, line 154: It's called the patch version in semver :) 
```suggestion
func decrementPatchVersion(version string) (string, error) {
```

dloucasfx: roger that!

jinja2:
@dloucasfx  We have some tests which use the K8S_VERSION var to look up for expected test data since the metrics change with k8s version for some receiver. These tests (i can only remember histograms from the top of my head) will probably start failing once a new minor version is introduced by this workflow. I took a look at the code and don't see this being addressed. Is this known? we can fix the tests once the new version is added or we could probably just copy the data from `new minor version - 1` when the workflow runs.

PR #1800, opened 2025-04-25
Upgrade NodeJS instrumentation to v3.1.2

pjanotti:
Upgrade NodeJS instrumentation to v3.1.2 - supersedes #1744


helm-charts/splunk-otel-collector/templates/operator/_helpers.tpl
```diff
@@ -153,7 +153,7 @@ Helper for generating environment variables for each instrumentation library.
 
   {{- /* Handle custom or default exporter endpoint */ -}}
   {{- $customOtelExporterEndpoint := "" }}
-  {{- if or (eq .instLibName "dotnet") (eq .instLibName "python") (eq .instLibName "java") }}
+  {{- if or (eq .instLibName "dotnet") (eq .instLibName "java") (eq .instLibName "nodejs") (eq .instLibName "python") }}
     {{- $customOtelExporterEndpoint = .endpoint | replace ":4317" ":4318" }}
```
pjanotti, line 157: We should consider if we still have cases that need this conditional. In principle we could just default to 4318.

.github/workflows/functional_test_v2.yaml
```diff
@@ -68,6 +68,12 @@ jobs:
       - name: Update dependencies
         run: |
           make dep-update
+      - name: Temporary generate new nodejs image - do not upload it
```
pjanotti, line 71: Currently these images a pushed to quay.io in a separate step. This specific image is not that expensive to build, but, to facilitate upgrades that require changes to the test images we may want to consider saving the needed images as workflow artifacts instead of publishing them to quay.io

atoulme: That might help. That can be tracked separately.

atoulme (approved):
LGTM

PR #1808, opened 2025-04-30
refactor functional and migration workflows and read the matrix from a file

dloucasfx:
This PR extracts the k8s related test matrix from related workflows and read it from a localized json file
This is needed to workaround the token issue for creating PRs in https://github.com/signalfx/splunk-otel-collector-chart/pull/1792

The code in https://github.com/signalfx/splunk-otel-collector-chart/pull/1792 will be updated and be much more simplified.

Note: `test-matrix.json` cannot be in the .github folder, otherwise it requires a PAT 
I am not sure the best path where this file should live, so for now I will add it in the same folder where the tool that updates will live

I also changed some of the matrix variable names so the update tool can easily identify and distinguish minikube from kind

### DIFF
--- a/.github/workflows/functional_test.yaml
+++ b/.github/workflows/functional_test.yaml
@@ -10,29 +10,28 @@ env:
   SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
 
 jobs:
-  functional-test:
-    name: K8s ${{ matrix.kubernetes_version }} ${{ matrix.container_runtime }}, Splunk ${{ matrix.splunk_version }}
+  get-test-matrix:
     runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


jvoravong (approved):
Just fix the pre-commit check and LGTM

tools/k8s_versions/test-matrix.json
dmitryax, line 1: Let's maybe put in in the root directory as `ci-matrix.json`. Later, we would need to move all the test under one directory with this file

dmitryax (approved):

tools/k8s_versions/test-matrix.json
dloucasfx: done

dmitryax (approved):

PR #1815, opened 2025-05-01
[chore][misspell] Add linter to help catch typos

crobert-1:

Adds the misspell tool to help catch typos in this repo. `Makefile` changes taken from the contrib and Splunk distributions of the Collector.

A future work item will be to add a workflow step that checks to make sure misspell is applied on all PRs.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -1,0 +1,29 @@
+SHELL=/bin/bash
+.SHELLFLAGS = -o pipefail -c
+
+
+# SRC_ROOT is the top of the source tree.
+ALL_SRC_AND_DOC_CMD := find . -type f \( \( -name "*.md" -o -name "*.go" -o -name "*.yaml" \) \) | sort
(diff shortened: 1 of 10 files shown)


Makefile.common
dmitryax, line 1: Do we need a common makefile if we have only one of them in this repo?

dmitryax (approved):

Makefile.common
crobert-1: I added it here as it will be needed in the next PR adding `golangci-lint`, where the `lint` target will be in `Makefile.Common`, and then more makefiles will be added importing it. In other words, if we add it here the next PR will only need to add to it and then import it as needed, vs. moving it and making the next PR seem more complicated than it is.

Let me know if you'd prefer to remove it for this PR.

dmitryax: Ok. I see. Makes sense

PR #1821, opened 2025-05-02
[chore] Fix linting errors

crobert-1:

Errors fixed:

1. `unused-parameter`
2. `var-naming: don't use an underscore in package name`
3. `var-declaration: should drop = 0 from declaration of var ...; it is the zero value`
4. `indent-error-flow: if block ends with a return statement, so drop this else and outdent its block`
5. `func `createObjectFromURL` is unused`
6. `ifElseChain: rewrite if-else to switch statement`
7. `shadow: declaration of "err" shadows declaration at line`
8. `should merge variable declaration with assignment on next line`
9. ` ineffectual assignment to err` (Means `err` wasn't used after being assigned, I added checks to tests to ensure it was empty for these cases)
10. `SA4010: this result of append is never used, except maybe in other appends`
11. `crdsInstallEnabled - result 1 (error) is always nil (unparam)`
12. `use-any: since Go 1.18 'interface{}' can be replaced by 'any'`
13. `comparing with == will fail on wrapped errors. Use errors.Is to check for a specific error`
14. `Error return value of `k8stest.DeleteObject` is not checked`
15. Some `gosec` warnings, resolved some and ignored others.
16. `could remove embedded field "ChartPathOptions" from selector`
17. `ST1017: don't use Yoda conditions (staticcheck)`
18. ` compares: use assert.GreaterOrEqual`
19. `regexp: remove unnecessary regexp.MustCompile`
20. `compares: use require.Equal`
21. `"200" can be replaced by http.StatusOK (usestdlibvars)`
22. `"GET" can be replaced by http.MethodGet (usestdlibvars)`
23. `context.Background() could be replaced by t.Context() in listPodsInNamespace (usetesting)`
24. ` unnecessary trailing newline (whitespace)`
25. `commentFormatting: put a space between `//` and comment text (gocritic)`

Note: There are still a couple linting errors left, but they may require a bit more logic to resolve so I didn't want to include them here.

### DIFF
--- a/functional_tests/functional/functional_test.go
+++ b/functional_tests/functional/functional_test.go
@@ -4,7 +4,6 @@
 package functional
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -74,7 +73,7 @@ type sinks struct {
 
 func setupSinks(t *testing.T) {
 	// create an API server
-	internal.SetupSignalFxApiServer(t)
+	internal.SetupSignalFxAPIServer(t)
 	globalSinks = &sinks{
 		logsConsumer:         internal.SetupHECLogsSink(t),
 		hecMetricsConsumer:   internal.SetupHECMetricsSink(t),
@@ -109,7 +108,9 @@ func deployChartsAndApps(t *testing.T, testKubeConfig string) {
 			continue
 		}
 
-		obj, groupVersionKind, err := decode(
+		var groupVersionKind *schema.GroupVersionKind
+		var obj k8sruntime.Object
+		obj, groupVersionKind, err = decode(
 			[]byte(resourceYAML),
 			nil,
 			nil)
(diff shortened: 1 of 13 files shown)


functional_tests/functional/functional_test.go
```diff
@@ -344,11 +343,11 @@ func deployChartsAndApps(t *testing.T, testKubeConfig string) {
 			return
 		}
 		t.Log("Cleaning up cluster")
-		teardown(t, testKubeConfig)
+		teardown(t, context.Background(), testKubeConfig)
```
crobert-1, line 346: This will end up needing a `//nolint` directive. The current output of linting is:
```
context.Background() could be replaced by t.Context() in ... (usetesting)
```
I originally tried to replace `context.Background()` in this instance, but since this is called in test cleanup, the test's context is already cancelled. This means that following usages of `t.Context()` always return context cancelled, so cleanup actions fail, causing functional tests to fail.

dloucasfx (approved):

PR #1826, opened 2025-05-05
[chore] Fix remaining linting errors

crobert-1:

Linting errors being fixed (I've only included a single example of each kind of linting error):

1. ```functional_tests/istio/istio_test.go:263:22: Error return value of `k8stest.DeleteObject` is not checked (errcheck)```
2. ```functional_tests/functional/functional_test.go:300:26: shadow: declaration of "err" shadows declaration at line 92 (govet)```
3. ```functional_tests/functional/functional_test.go:350:29: context-as-argument: context.Context should be the first parameter of a function (revive)```
4. ```functional_tests/internal/api_server.go:38:4: go-require: require must only be used in the goroutine running the test function (testifylint)```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -54,10 +54,6 @@ linters:
 
     # Excluding configuration per-path, per-linter, per-text and per-source
     rules:
-      - linters:
-          - unused
-        path: pagefile.go # This exclusion is required for Windows only
-        text: cachedBytes
 
     # Log a warning if an exclusion rule is unused.
     warn-unused: true
--- a/Makefile.common
+++ b/Makefile.common
@@ -54,6 +54,7 @@ misspell-correction: $(TOOLS_BIN_DIR)/misspell
 gofmt: $(GOFUMPT) $(GOIMPORTS) misspell-correction
 	gofmt -w -s .
 	$(GOFUMPT) -l -w .
+	$(MAKE) gogci
 	$(GOIMPORTS) -w -local github.com/signalfx/splunk-otel-collector-chart ./
 
 .PHONY: golint
--- a/functional_tests/internal/api_server.go
+++ b/functional_tests/internal/api_server.go
@@ -4,7 +4,6 @@
 package internal
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"net/http"
@@ -22,7 +21,6 @@ func SetupSignalFxAPIServer(t *testing.T) {
 		writer.WriteHeader(http.StatusOK)
 	})
 
-	_, cancelCtx := context.WithCancel(t.Context())
 	s := &http.Server{
 		Addr:              fmt.Sprintf("0.0.0.0:%d", SignalFxAPIPort),
 		Handler:           mux,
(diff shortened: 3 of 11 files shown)


functional_tests/internal/api_server.go
```diff
@@ -29,13 +29,17 @@ func SetupSignalFxAPIServer(t *testing.T) {
 		ReadHeaderTimeout: 60 * time.Minute,
 	}
 
+	errCh := make(chan error)
```
crobert-1, line 32: Use a channel for sending error back to the test's goroutine, as require must be used by the goroutine running the test.

functional_tests/internal/common.go
```diff
@@ -195,3 +201,12 @@ func WaitForTerminatingPods(t *testing.T, clientset *kubernetes.Clientset, names
 		return terminatingPods == 0
 	}, 2*time.Minute, 5*time.Second, "there are still terminating pods after 2 minutes")
 }
+
+func DeleteObject(t *testing.T, k8sClient *k8stest.K8sClient, objYAML string) {
```
crobert-1, line 205: There were two identical `deleteObject` methods in `istio_test.go` and `k8sevent_test.go`, so I've consolidated them here.

.golangci.yml
```diff
@@ -54,10 +54,6 @@ linters:
 
     # Excluding configuration per-path, per-linter, per-text and per-source
     rules:
-      - linters:
-          - unused
-        path: pagefile.go # This exclusion is required for Windows only
```
crobert-1, line 59: This exclusion was brought from contrib, irrelevant for this repo. It's also giving a `WARNING` message for every directory causing noise.

Makefile.common
```diff
@@ -54,6 +54,7 @@ misspell-correction: $(TOOLS_BIN_DIR)/misspell
 gofmt: $(GOFUMPT) $(GOIMPORTS) misspell-correction
 	gofmt -w -s .
 	$(GOFUMPT) -l -w .
+	$(MAKE) gogci
```
crobert-1, line 57: This writes changes for import ordering when they're not properly formatted. This makes it so that if `golint` fails,`gofmt` can be run to fix this kind of failure.

tools/empty_test.go
crobert-1, line 1: Without this file `golint` complains:
```
ERRO Running error: context loading failed: no go files to analyze: running `go mod tidy` may solve the problem 
```
This is because the package doesn't have any go files being built directly (`tools.go` has the `tools` build directive, so `golangci-lint` isn't picking it up properly).

dmitryax: why do we need build tools directive?

crobert-1: It looks like it's to hide it from build/linting tools:
```
tools/tools.go:19:2: import "github.com/client9/misspell/cmd/misspell" is a program, not an importable package (typecheck)
	_ "github.com/client9/misspell/cmd/misspell"
	^
```
(same error for the rest of the imports as well)

dmitryax (approved):

tools/empty_test.go
crobert-1: Unfortunately that doesn't work either. Since the failure is a `typecheck`, it means that compilation is failing for the linter. Exclusions are applied after the compilation step so the failure would still occur.

Also, once Dani's tool is introduced to the `tools` directory we'll want to run lint against it.

crobert-1: Relevant comment about this: https://github.com/golangci/golangci-lint/issues/2912#issuecomment-1149995159

PR #1866, opened 2025-06-12
Fix default config for enabled noDropLogsPipeline feature gate

dmitryax:
Update the config to adapt to the latest changes in the collector exporter batcher interface


crobert-1 (approved):
_Vuln failures look like the go version used in GitHub workflows needs updated, unrelated to this PR._

PR #1870, opened 2025-06-18
Add OTLP token passthrough support

dloucasfx:


.github/workflows/configs/kind-config.yaml
```diff
@@ -36,3 +36,6 @@ nodes:
       - containerPort: 443
         hostPort: 443
         protocol: TCP
+      - containerPort: 4318
+        hostPort: 43180
```
pjanotti, line 41: It seems worth a comment that the test expects this port (43180)

examples/add-filter-processor/rendered_manifests/configmap-agent.yaml
```diff
@@ -19,11 +19,15 @@ data:
   relay: |
     exporters:
       otlphttp:
+        auth:
+          authenticator: headers_setter
         headers:
           X-SF-Token: ${SPLUNK_OBSERVABILITY_ACCESS_TOKEN}
```
pjanotti: Why this one is kept?

pjanotti (approved):

examples/add-filter-processor/rendered_manifests/configmap-agent.yaml
dloucasfx:
```
headers:
          X-SF-Token: ${SPLUNK_OBSERVABILITY_ACCESS_TOKEN}
```
I went through all components that makes up the passthrough, checkeds previous PRs (standalone config), tested multiple scenarios and I can't see or any reason why we left this.
The only explanation is that the original author for passthrough documented it and left the headers in place and we carried it from there . 
In short, this is redundant and no longer needed when using headers_setter , I will update the PR.
Thanks !

.github/workflows/configs/kind-config.yaml
dloucasfx: I will add a comment in the test and the kind config.
Thanks

dmitryax (approved):
LGTM

PR #1772, opened 2025-04-15
fix(collector): add support for service trafficDistribution

sjoukedv:
Equivalent of https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1630

PR #1874, opened 2025-06-25
Add trafficDistribution support to the gateway and agent services

dmitryax:
Replaces https://github.com/signalfx/splunk-otel-collector-chart/pull/1772

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/service.yaml
+++ b/helm-charts/splunk-otel-collector/templates/service.yaml
@@ -20,6 +20,9 @@ metadata:
 {{- end }}
 spec:
   type: {{ .Values.service.type }}
+  {{- if .Values.service.trafficDistribution }}
+  trafficDistribution: {{ .Values.service.trafficDistribution }}
+  {{- end }}
   ports:
   {{- range $key, $port := $gateway.ports }}
   {{- $metricsEnabled   := and (eq (include "splunk-otel-collector.metricsEnabled"   $) "true") (has "metrics"   $port.enabled_for) }}
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -1195,6 +1195,8 @@ gateway:
 service:
   # Service type
   type: ClusterIP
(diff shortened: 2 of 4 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -422,6 +422,7 @@ agent:
     # create a service for the agents with a local internalTrafficPolicy
     # so that agent pods can be discovered via dns etc
     enabled: true
+    trafficDistribution: ""
```
jinja2: I think this config would be a no-op for our agent k8s service since it has like strict traffic distribution guarantee already with the hardcoded `internalTrafficPolicy: Local` in there

jinja2 (commented):
Can you add enable this config in one of our examples? Tmk, this config is only introduced in a recent version of k8s and i want to see what kubeconform output. I think this might require a note or check for min k8s version

helm-charts/splunk-otel-collector/values.yaml
dmitryax: You're right. Let me remove this from the agent

dmitryax:
> Can you add this config in one of our examples? Tmk, this config is only introduced in a recent version of k8s and i think kubeconform output would shed some light on version support for this. I think this might require a note or check for min k8s version

Apparently it's only available since 1.33 and our tests are still at 1.31. So adding it to example whould be failing. I added a note in the comments

helm-charts/splunk-otel-collector/templates/service.yaml
```diff
@@ -20,6 +20,9 @@ metadata:
 {{- end }}
 spec:
   type: {{ .Values.service.type }}
+  {{- if .Values.service.trafficDistribution }}
```
jinja2, line 23:
```suggestion
  {{- if and .Values.service.trafficDistribution (semverCompare ">=1.33.0-0" .Capabilities.KubeVersion.Version) }}
```
We could additionally verify in template but the clarification in comment is also good with me.

dmitryax: I think it's fine to fail rather then suppressing the user error.

PR #1887, opened 2025-07-02
Update k8s control plane prometheus receiver configs

jinja2:

* Updates config for receivers which collect prometheus metrics from k8s control plane components. The updates config keeps only metrics which are used in bundled control plane dashboards.
* Updates histogram tests to reflect the new limited set of histogram and non-histogram metrics being kept in prom receiver
* Adds a helper script which can be used to generate test data and confirm test results for the various k8s version on dev's local machine.


.chloggen/update-controlplane-metrics.yaml
```diff
@@ -0,0 +1,13 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
jinja2: Maybe this should be breaking given it is dropping metrics

dmitryax: Yeah, probably

dmitryax (approved):

PR #1888, opened 2025-07-02
Enable discovery on Windows

pjanotti:
There is no need for special paths or similar when enabling discovery on Windows. Obs.: Windows images have only a `C:` drive* so any root path without drive letter it is just fine.

Note: in principle one could create another driver letter either with `VOLUME` on the Dockerfile or using the `subst` command. However, I didn't find any successful example of using `VOLUME` and using `subst` is just to map a letter to the existing volume. In the end we can consider that there is just the `C:` volume and all root paths are fine.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/daemonset.yaml
+++ b/helm-charts/splunk-otel-collector/templates/daemonset.yaml
@@ -253,7 +253,7 @@ spec:
         {{- if .Values.agent.featureGates }}
         - --feature-gates={{ .Values.agent.featureGates }}
         {{- end }}
-        {{- if and .Values.agent.discovery.enabled (not .Values.isWindows) }}
+        {{- if and .Values.agent.discovery.enabled }}
         - --discovery
         {{- end }}
         ports:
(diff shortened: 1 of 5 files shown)


helm-charts/splunk-otel-collector/templates/daemonset.yaml
```diff
@@ -516,7 +512,7 @@ spec:
       {{- if .Values.isWindows }}
       - name: hostfs
         hostPath:
-          path: "C:\\"
```
jinja2, line 519: Q - was the 2nd backslash meant for escaping purposes?

pjanotti: Yes, inside double-quoted strings in YAML one needs to escape the `\`, remove the double-quoting and no escape is needed.

pjanotti: Rule of thumb for Windows file paths in YAML files:
```YAML
exporters:
  file/path-backslash:
    path: c:\Users\pjanotti\src\spikes\dir with spaces\logs\backslash.log
  file/path-forwardslash:
    path: c:/Users/pjanotti/src/spikes/dir with spaces/logs/forwardslash.log
  file/path-backslash-nodrive:
    path: \Users\pjanotti\src\spikes\dir with spaces\logs\backslash-nodrive.log
  file/path-forwardslash-nodrive:
    path: /Users/pjanotti/src/spikes/dir with spaces/logs/forwardslash-nodrive.log
  file/path-with-quotes: # needs to escape '\'
    path: "c:\\Users\\pjanotti\\src\\spikes\\dir with spaces\\logs\\with quotes.log"
```

PR #1891, opened 2025-07-07
Remove EKS and EC2 detectors for EKS/Fargate distro

dloucasfx:


pszkamruk-splunk (approved):

.chloggen/fargatedetector.yaml
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: clusterReceiver
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove EKS and EC2 resource detectors for EKS/Fargate distribution as it does not support IMDS
```
jinja2:
```suggestion
note: Remove EKS and EC2 resource detectors for EKS/Fargate distribution as the collectors do not have access to IMDS
```

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -477,7 +477,7 @@ Build the securityContext for Linux and Windows
 Whether the clusterName configuration option is optional
 */}}
 {{- define "splunk-otel-collector.clusterNameOptional" -}}
-{{- or (hasPrefix "gke" (include "splunk-otel-collector.distribution" .)) (hasPrefix "eks" (include "splunk-otel-collector.distribution" .)) }}
+{{- or (hasPrefix "gke" (include "splunk-otel-collector.distribution" .)) (eq (include "splunk-otel-collector.isNonFargateEKS" .) "true") }}
```
dmitryax, line 480: Do we really need `splunk-otel-collector.isNonFargateEKS` given that there is only one option left?
```suggestion
{{- or (hasPrefix "gke" (include "splunk-otel-collector.distribution" .)) (eq (include "splunk-otel-collector.distribution" .) "eks") }}
```

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -523,3 +523,19 @@ Create the name of the target allocator cluster role binding to use
...
 {{- end -}}
+
+{{/*
+Checks if the distribution is eks/fargate and fails if clusterName is not set.
+*/}}
+{{- define "splunk-otel-collector.validateEKSFargateConfig" -}}
+{{- if and (eq (include "splunk-otel-collector.distribution" .) "eks/fargate") (not .Values.clusterName) -}}
+  {{- fail "When distribution is set to 'eks/fargate', the 'clusterName' value must be specified" -}}
+{{- end -}}
+{{- end -}}
```
dmitryax: We don't have this for other distributions. I'd say we shouldn't require this for fargate either for now. The data management UI won't work without `cloud.account.id` anyway

helm-charts/splunk-otel-collector/templates/_helpers.tpl
dloucasfx: @dmitryax `eks/auto-mode` distribution will be added soon

helm-charts/splunk-otel-collector/templates/_helpers.tpl
dloucasfx: We recommend not setting it for EKS https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/values.yaml#L14-L26
If we don't force it, wouldn't it be risking sending data without the `k8s.cluster.name`?

dloucasfx: I see what you mean now, ie: we don't force it for other distro, why would we force it here?
In addition to my previous reply, I am forcing it because fargate would not try to auto discover it.

jinja2: If auto discovery of cluster name is an important enough feature for customers, should we introduce a var to add the detector which a customer willing to configure pod iam access can enable? Iirc we are already planning to doc this stuff for auto mode, right?

dmitryax: I believe we need consistent rules/guedelines to for users to setup `k8s.cluster.name` and `cloud.account.id`. Otherwise it's inconsistent and semi-broken anyway

dloucasfx: @jinja2 
> should we introduce a var to add the detector which a customer willing to configure pod iam access can enable?

If you are referring to pod identity, fro my research , it is not supported for Fargate
````
You can’t use EKS Pod Identities with:

Pods that run anywhere except Linux Amazon EC2 instances. Linux and Windows pods that run on AWS Fargate (Fargate) aren’t supported. Pods that run on Windows Amazon EC2 instances aren’t supported.
````

Comment (dloucasfx) [reply helm-charts/splunk-otel-collector/templates/_helpers.tpl]: > I believe we need consistent rules/guedelines to for users to setup k8s.cluster.name and cloud.account.id. Otherwise it's inconsistent and semi-broken anyway

I am surprised that this issue, ie: missing `cloud.account.id` and `k8s.cluster.name` have not been raised before.
`k8s.cluster.name` might have been set as our UI/Wizard requires the cluster name to be defined, but `cloud.account.id` is definitely, always missing

Comment (jinja2) [reply helm-charts/splunk-otel-collector/templates/_helpers.tpl]: I believe account id can be fetched from the k8s api with 0.128.0 detector. We should rethink removing the detector.

@dloucasfx sorry, i think you mentioned the reason why this fails but I didn't fully catch that. Tmk, detector only logs an error msg when it can't fetch the attr value. But I think you mentioned 2 conflicting changes in upstream which broke this?

My understanding of the current state of cluster name discovery in aws is as follows -

- EKS vanilla - Should work OOTB unless a k8s admin has blocked access to IMDS
- EKS auto mode - Does not work, IMDS can only be access from hostNetwork so a user can either set hostNetwork true or provide the pod required iam access.
- EKS fargate - Does not work and we can not set hostNetwork to true in fargate. So the only option is user does the pod irsa setup to get this working.

Adding a clusterName to values seems less work for the user at this point given the amount of documentation we are expecting them to go through for the different cases here.

The other option could be we take the clusterName var, and update to the [resource processor](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl#L205) to use `insert` strategy? We can document that the value's clusterName will the back up used if the detector discovery does not work. Assuming we can restore the og behaviour of the detector to only log errors.

Comment (dloucasfx) [reply helm-charts/splunk-otel-collector/templates/_helpers.tpl]: had a chat with @dmitryax 
The fact that:
- This has been broken for a while and the cluster name issue not been raised
- The UI forces that user to set the cluster name
- And that I will go ahead update the comment here https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/values.yaml#L14-L26

I will removed this validation as requested

Comment (jinja2) [reply helm-charts/splunk-otel-collector/templates/_helpers.tpl]: ok, but I think we should update the clusterName comment to call out fargate requires it to be set. 

Jfyi pods in fargate can use the older pod iam access stuff

```
To give the containers in your Fargate Pod permissions to access other AWS services, you must use [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).
```

Comment (dloucasfx) [reply helm-charts/splunk-otel-collector/templates/_helpers.tpl]: @jinja2 
> I believe account id can be fetched from the k8s api with 0.128.0 detector. We should rethink removing the detector.

The account ID is fetched from the `DescribeInstances` https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c457108e50c13c06c2a50f7535656d2c305d5d45/internal/metadataproviders/aws/eks/metadata.go#L129 which requires metadata server access for auth or Pod Identity which does not "seem" to be supported on fargate.

>  sorry, i think you mentioned the reason why this fails but I didn't fully catch that. Tmk, detector only logs an error msg when it can't fetch the attr value. But I think you mentioned 2 conflicting changes in upstream which broke this?

EKS 0.127.0 silently fails https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.127.0/processor/resourcedetectionprocessor/internal/aws/eks/detector.go#L164 and log this message, it does not return an error to the caller , so the retry logic is not kicked in

````
2025-07-06T22:33:17.705Z        warn    eks/detector.go:155     Unable to get EKS cluster name  {"resource": {"service.name": "otel-k8s-cluster-receiver"}, "otelcol.component.id": "resourcedetection/               k8s_cluster_name", "otelcol.component.kind": "processor", "otelcol.pipeline.id": "metrics", "otelcol.signal": "metrics", "error": "operation error ec2imds: GetRegion, exceeded maximum number of attempts, 3,        request send failed, Get \"http://169.254.169.254/latest/dynamic/instance-identity/document\": dial tcp 169.254.169.254:80: i/o timeout"}  
````

As for the conflicting errors, I will document them in a separate issue, but it won't solve our issue here.
Also note, that the accounID prior to 0.127.0 was "supposed" to come from EC2 detector, which also fails.

> EKS fargate - Does not work and we can not set hostNetwork to true in fargate. So the only option is user does the pod irsa setup to get this working.

I don't believe any of these options are available for Fargate

Comment (dloucasfx) [reply helm-charts/splunk-otel-collector/templates/_helpers.tpl]: > ok, but I think we should update the clusterName comment to call out fargate requires it to be set.
I updates the values.yaml

Comment (jinja2) [reply helm-charts/splunk-otel-collector/templates/_helpers.tpl]: Ah, I thought we had the account id from the k8s node in 0.128. But we don't, thanks for confirming.

> > EKS fargate - Does not work and we can not set hostNetwork to true in fargate. So the only option is user does the pod irsa setup to get this working.

> I don't believe any of these options are available for Fargate

I have not tried this in fargate myself but AWS document [here](https://docs.aws.amazon.com/eks/latest/userguide/pod-execution-role.html) does mentions that IRSA works with fargate pods.
```
Important
The containers running in the Fargate Pod can’t assume the IAM permissions associated with a Pod execution role. To give the containers in your Fargate Pod permissions to access other AWS services, you must use [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).
```

IRSA is similar to pod identities; just an older method to grant pods iam access. Maybe we can do some work for documenting fargate compatibility in some followup ticket.

Anyways, I am ok with the PR and just wanted to write down my thoughts :)

dmitryax (approved):

PR #1903, opened 2025-07-10
EKS Auto Mode - Remove EC2 detector and expand EKS detector which supports fetching attributes through API (Part 2 - POD Identity) .

dloucasfx:
Remove EC2 detector and expand EKS detector which supports fetching attributes through API (Part 2 - POD Identity) .

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -72,14 +72,12 @@ resourcedetection:
     - env
     {{- if or (hasPrefix "gke" (include "splunk-otel-collector.distribution" .)) (eq (include "splunk-otel-collector.cloudProvider" .) "gcp") }}
     - gcp
-    {{- else if eq (include "splunk-otel-collector.isNonFargateEKS" .) "true" }}
+    {{- else if or (eq (include "splunk-otel-collector.isNonEKSonAWS" .) "true") (eq (include "splunk-otel-collector.isNonFargateEKS" .) "true") }}
     - eks
     {{- else if eq (include "splunk-otel-collector.distribution" .) "aks" }}
     - aks
     {{- end }}
-    {{- if and (eq (include "splunk-otel-collector.cloudProvider" .) "aws") (ne (include "splunk-otel-collector.distribution" .) "eks/fargate") }}
-    - ec2
-    {{- else if eq (include "splunk-otel-collector.cloudProvider" .) "azure" }}
+    {{- if eq (include "splunk-otel-collector.cloudProvider" .) "azure" }}
     - azure
     {{- end }}
     # The `system` detector goes last so it can't preclude cloud detectors from setting host/os info.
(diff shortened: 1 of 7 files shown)


helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -90,11 +88,28 @@ resourcedetection:
     resource_attributes:
       k8s.cluster.name:
         enabled: true
-  {{- else if and (eq (include "splunk-otel-collector.isNonFargateEKS" .) "true") (not .Values.clusterName) }}
+  {{- else if or ( and (eq (include "splunk-otel-collector.cloudProvider" .) "aws") (eq (include "splunk-otel-collector.distribution" .) "") ) (eq (include "splunk-otel-collector.isNonFargateEKS" .) "true") }}
```
dmitryax: Should we put `( and (eq (include "splunk-otel-collector.cloudProvider" .) "aws") (eq (include "splunk-otel-collector.distribution" .) "")` in `splunk-otel-collector.isNonFargateEKS`? It's confusing why we do that her anyway, so maybe we put it inside the func and add a comment?

dloucasfx: agree, it is a bit confusing; 
but adding it to `isNonFargateEKS` breaks the semantic of this function, which identifies EKS clusters that are NOT fargate; with the suggested change, it will return a cluster that is not EKS (example: Vanilla K8S on EC2)

2 options:
1- create a function "isNonEKSonAWS" or ("isK8sonAWS") that returns true if it's a non EKS K8S cluster on AWS 
2- Leave it as is and add a comment to what we are checking for.

dloucasfx: would something like this be clear?
```
{{/*
Identifies K8s clutser running on AWS but they are not EKS.
Returns true if the cloud provider is aws and distribution is not set.
example: Vanilla K8s on AWS EC2
*/}}
{{- define "splunk-otel-collector.isNonEKSonAWS" -}}
{{- and (eq (include "splunk-otel-collector.cloudProvider" .) "aws") (eq (include "splunk-otel-collector.distribution" .) "") -}}
{{- end -}}
```

dmitryax: Something like `isK8sClusterOnAWSWithNodes` ? :)

dmitryax: Actually the extra `isNonEKSonAWS` function as you suggested works as well

dloucasfx: PR is updated with the above change

dmitryax: Can we use it in this line as well?

dloucasfx: arghhh... how did I miss that, sorry, updating now...

PR #1906, opened 2025-07-11
[chore] add operator crds to kubeconform

jinja2:

* add opentelemetry operator's custom resource schema to kubeconform. This will verify that our instrumentation object is in-line with crd changes for the same. We use a [helper script](https://github.com/yannh/kubeconform/blob/master/scripts/openapi2jsonschema.py) provided in kubeconform to convert the schema from the operator CRD objects to a valid JSON schema. The repo contains the schemas in the folder generate-crd-schemas. The operator upgrade workflow which picks up crd changes will be updating these jsonschemas as necessary when crds subchart is updated.
* updates render_examples.sh scripts to redact data such that the final yamls are valid - the current method is making yaml invalid and kubeconform fails to parse the admission-webhook.yaml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,3 +269,7 @@ gomoddownload:
 .PHONY: update-matrix-versions
 update-matrix-versions: ## Update matrix, ex: K8s cluster versions used for testing. Set DEBUG=-debug to enable debug logs.
 	go run ./tools/k8s_versions/update_k8s_versions.go $(DEBUG)
+
+.PHONY: kubeconform
+kubeconform: ## Run kubeconform validation on all rendered manifests
+	./ci_scripts/kubeconform-all.sh
--- a/Makefile
+++ b/Makefile
@@ -272,7 +272,7 @@ update-matrix-versions: ## Update matrix, ex: K8s cluster versions used for test
 
 .PHONY: kubeconform
 kubeconform: ## Run kubeconform validation on all rendered manifests
-	./ci_scripts/kubeconform-all.sh
+	./ci_scripts/kubeconform-all.sh $(K8S_VERSION)
 
 .PHONY: generate-crd-schemas
 generate-crd-schemas: ## Generate JSON schemas from operator CRDs for kubeconform
--- a/Makefile
+++ b/Makefile
@@ -273,3 +273,7 @@ update-matrix-versions: ## Update matrix, ex: K8s cluster versions used for test
 .PHONY: kubeconform
 kubeconform: ## Run kubeconform validation on all rendered manifests
 	./ci_scripts/kubeconform-all.sh
+
+.PHONY: generate-crd-schemas
+generate-crd-schemas: ## Generate JSON schemas from operator CRDs for kubeconform
+	ci_scripts/generate-crd-schemas.sh
--- a/ci_scripts/kubeconform-all.sh
+++ b/ci_scripts/kubeconform-all.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+# kubeconform-all.sh
+# Runs kubeconform on all rendered manifests in examples, using both default and operator CRDs.
+# Usage: ./ci_scripts/kubeconform-all.sh [k8s_version]
+
+set -euo pipefail
+
+K8S_VERSION="${1:-1.33.0}"
+EXAMPLES_DIR="examples"
+CRDS_DIR="helm-charts/splunk-otel-collector/charts/opentelemetry-operator-crds/crds"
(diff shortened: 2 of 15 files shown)


ci_scripts/kubeconform-all.sh
```diff
@@ -0,0 +1,55 @@
...
+source "$VENV_DIR/bin/activate"
+pip install --quiet pyyaml
+
+# Generate JSON schemas from operator CRDs
+(
+  cd "$SCHEMA_DIR"
+  for crd in "$OLDPWD/$CRDS_DIR"/*.yaml; do
+    FILENAME_FORMAT="{fullgroup}_{kind}_{version}" python "openapi2jsonschema.py" "$crd"
+  done
+)
```
dloucasfx: The JSON schemas is independent of the K8s Version we are validating, right?
If it's the case and the schema does not change, unless CRDs are updated, can we generate and store the schema to be reused, instead of regenerating it for each K8sversion ?

Makefile
```diff
@@ -269,3 +269,7 @@ gomoddownload:
 .PHONY: update-matrix-versions
 update-matrix-versions: ## Update matrix, ex: K8s cluster versions used for testing. Set DEBUG=-debug to enable debug logs.
 	go run ./tools/k8s_versions/update_k8s_versions.go $(DEBUG)
+
+.PHONY: kubeconform
```
jvoravong, line 273: Is it easy to add kubeconform to `make install-tools`?

ci_scripts/kubeconform-all.sh
jinja2: yeah, the crds should be independent. This was just easier but I don't mind having them in. Let me add the CRDs in this PR and my other wip PR which is for fixing operator upgrades will have the logic to keep these schemas updated.

examples/enable-operator-and-auto-instrumentation/rendered_manifests/operator/admission-webhooks/operator-webhook.yaml
```diff
@@ -31,22 +31,22 @@ webhooks:
...
-        - opentelemetry.io
-      apiVersions:
-        - v1alpha1
-      operations:
-        - CREATE
-        - UPDATE
-      resources:
-        - instrumentations
-      scope: Namespaced
+      - apiGroups:
```
jvoravong, line 34: Not blocking, but I’m concerned that our rendered version diverges from upstream. Redacted fields are fine, but the other changes seem like they should be upstreamed. Happy to review and approve that if we pursue it.

jinja2: The content is the same; just indentation has changed due to passing it through yq

Makefile
jinja2: not sure, will have to look into it

examples/enable-operator-and-auto-instrumentation/rendered_manifests/operator/admission-webhooks/operator-webhook.yaml
jvoravong: EKS add-ons enforce strict validation rules, and changes like this can sometimes cause issues. I won’t block this, but we may need to revisit it later, especially since the operator isn’t currently supported in the add-on, though we’re working to unblock that.

If possible, please try removing this pre-commit exclusion as part of this PR:
https://github.com/signalfx/splunk-otel-collector-chart/blob/fd4270019a7ad114ed2ca7ed5a78340f4e501ab8/.pre-commit-config.yaml#L26

jinja2: will do, but I am still not sure how a change to the yaml in example folder affects anything? All this is doing is making the file an actually parseable yaml. Additionally, the indentation change are due to the helm template output going through an in-place yq format change

Makefile
jinja2: done

ci_scripts/kubeconform-all.sh
jinja2: @dloucasfx added the crds in this pr

examples/enable-operator-and-auto-instrumentation/rendered_manifests/operator/admission-webhooks/operator-webhook.yaml
jinja2: removed the exclusion filter

jvoravong (approved):

PR #1928, opened 2025-07-17
[chore] do not run collect splunk-pod-logs always in the Splunk integration test pipeline

pszkamruk-splunk:
Description:
Fix for a situation where the Splunk setup failed and we attempted to collect logs, but it was not possible because the Splunk pod did not start due to a Minikube setup issue.
Now, we will trigger the log collection step if there is a failure in the functional test run step.

### DIFF
--- a/.github/workflows/functional_test.yaml
+++ b/.github/workflows/functional_test.yaml
@@ -141,7 +141,7 @@ jobs:
           retention-days: 5
 
       - name: Print splunk-otel-collector logs
-        if: always()
+        if: always() && steps.run-functional-tests.outcome == 'failure'
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/functional_test.yaml
```diff
@@ -141,7 +141,7 @@ jobs:
           retention-days: 5
 
       - name: Print splunk-otel-collector logs
-        if: always()
+        if: always() && steps.run-functional-tests.outcome == 'failure'
```
jinja2:
```suggestion
        if: always() && (steps.run-functional-tests.outcome == 'failure' || env.KUBERNETES_DEBUG_INFO == 'true')
```
to keep it same as the other steps which print/collect debug info

jinja2 (approved):

PR #1941, opened 2025-07-29
add option to set metrics sourcetype directly from values.yaml

omrozowicz-splunk:
**Description:** 
This PR introduces support for the `sourcetypeMetrics` field in the Splunk Platform configuration within the Helm chart. The new field allows users to specify a custom sourcetype value for metrics ingested into Splunk Platform, providing greater flexibility and alignment with Splunk data models.

**Purpose**: 
Introduce `sourcetypeMetrics` for granular control over metrics sourcetypes, while ensuring backward compatibility.

**Current Behavior**: 
`sourcetype` applies to all data types (logs, traces, and metrics).

**New Behavior**:
`sourcetype` will now primarily define the sourcetype for logs and traces.
For metrics, the sourcetype is determined by the following hierarchy:
`sourcetypeMetrics`: If specified, it takes precedence.
`sourcetype`: If `sourcetypeMetrics` is not specified, the value from `sourcetype` is used (to avoid a breaking change for existing configurations).
`httpevent`: If neither `sourcetypeMetrics` nor s`ourcetype` is specified, `httpevent` is used as the default.

**Testing:** functional tests, manual tests

### DIFF
--- a/.chloggen/make-metrics-sourcetype-configurable.yaml
+++ b/.chloggen/make-metrics-sourcetype-configurable.yaml
@@ -1,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: chart
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add configuration option for Splunk Platform metrics sourcetype
+# One or more tracking issues related to the change
+issues: [1941]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/.chloggen/make-metrics-sourcetype-configurable.yaml
+++ b/.chloggen/make-metrics-sourcetype-configurable.yaml
@@ -3,10 +3,12 @@ change_type: enhancement
 # The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
 component: chart
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add configuration option for Splunk Platform metrics sourcetype
+note: Adds configuration option for Splunk Platform metrics sourcetype
 # One or more tracking issues related to the change
 issues: [1941]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext:
+subtext: |
+  - Users can override the metrics sourcetype for Splunk Platform using the `splunk.com/sourcetypeMetrics` annotation or the `.Values.splunkPlatform.sourcetypeMetrics` Helm value.
+  - See `advanced-configuration.md` for detailed precedence and usage notes.
--- a/.chloggen/make-metrics-sourcetype-configurable.yaml
+++ b/.chloggen/make-metrics-sourcetype-configurable.yaml
@@ -10,5 +10,5 @@ issues: [1941]
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
-  - Users can override the metrics sourcetype for Splunk Platform using the `splunk.com/sourcetypeMetrics` annotation or the `.Values.splunkPlatform.sourcetypeMetrics` Helm value.
+  - Users can override the metrics sourcetype for Splunk Platform using the `splunk.com/metricsSourcetype` annotation or the `.Values.splunkPlatform.metricsSourcetype` Helm value.
   - See `advanced-configuration.md` for detailed precedence and usage notes.
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -373,7 +373,7 @@ resource/metrics:
       {{- if .Values.splunkPlatform.metricsSourcetype }}
       value: {{.Values.splunkPlatform.metricsSourcetype | quote }}
       {{- else }}
-      value: "{{.Values.splunkPlatform.sourcetype }}"
+      value: {{.Values.splunkPlatform.sourcetype | quote }}
       {{- end }}
       action: insert
 {{- end }}
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -294,6 +294,12 @@ k8sattributes/metrics:
       - key: splunk.com/sourcetype
         tag_name: com.splunk.sourcetype
         from: pod
+      - key: splunk.com/sourcetypeMetrics
+        tag_name: com.splunk.sourcetype
+        from: namespace
+      - key: splunk.com/sourcetypeMetrics
+        tag_name: com.splunk.sourcetype
+        from: pod
       - key: splunk.com/metricsIndex
         tag_name: com.splunk.index
         from: namespace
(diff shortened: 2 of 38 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -54,8 +54,10 @@ splunkPlatform:
   # Optional. Default value for `source` field.
   source: "kubernetes"
   # Optional. Default value for `sourcetype` field. For container logs, it will
-  # be container name. For metrics and traces it will default to "httpevent".
+  # be container name. For traces it will default to "httpevent".
   sourcetype: ""
+  # Optional. Default value for metrics `sourcetype` field. If not set it will default to "httpevent".
+  sourcetypeMetrics: ""
```
sbylica-splunk: Do you think it would make sense to create something like "sourcetypeTraces" to make it consistent? We already do this for indexes for example, and it would be weird if we leave traces as "default only" and metrics/logs as configurable.

omrozowicz-splunk: I didn't think about it in scope of this task as for Splunk Platform we don't have such type as trace, only logs and metrics — traces when sent to Splunk are basically logs. I think typically traces are sent to Splunk Observability Platform, and if Splunk Platform user has it enabled it is rather treated as a log. 

> and it would be weird if we leave traces as "default only"

`sourcetype` is set for all types of data, so traces are not default only — when set the traces are configured with `sourcetype` from `values.yaml`, the same as logs

But setting `sourcetypeTraces` wouldn't hurt, if you believe that would be useful I can add it ☺️

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -54,8 +54,10 @@ splunkPlatform:
   # Optional. Default value for `source` field.
   source: "kubernetes"
   # Optional. Default value for `sourcetype` field. For container logs, it will
-  # be container name. For metrics and traces it will default to "httpevent".
+  # be container name. For traces it will default to "httpevent".
   sourcetype: ""
+  # Optional. Default value for metrics `sourcetype` field. If not set it will default to "httpevent".
```
pszkamruk-splunk: nit:

```suggestion
  # Optional. Default value for metrics `sourcetype` field. If not set it will follow the logic of "sourcetype" field.
```

omrozowicz-splunk: added, please take a look

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
```diff
@@ -40,7 +40,7 @@ processors:
   {{- include "splunk-otel-collector.k8sAttributesSplunkPlatformMetrics" . | nindent 2 }}
     filter:
       node_from_env_var: K8S_NODE_NAME
-  {{- if .Values.splunkPlatform.sourcetype }}
+  {{- if or .Values.splunkPlatform.metricsSourcetype .Values.splunkPlatform.sourcetype }}
```
jinja2, line 43:
```suggestion
  {{- if or .Values.splunkPlatform.sourcetypeMetrics .Values.splunkPlatform.sourcetype }}
```

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -377,7 +377,11 @@ resource/metrics:
   attributes:
     # Insert the sourcetype value from values.yaml if it has not already been set through annotations.
     - key: com.splunk.sourcetype
+      {{- if .Values.splunkPlatform.sourcetypeMetrics }}
+      value: "{{.Values.splunkPlatform.sourcetypeMetrics }}"
```
jinja2:
```suggestion
      value: {{ .Values.splunkPlatform.sourcetypeMetrics | quote }}
```
better to use the quote fn

.chloggen/make-metrics-sourcetype-configurable.yaml
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: chart
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add configuration option for Splunk Platform metrics sourcetype
```
jinja2: Iirc, there is also a way to set sourcetype using an annotation on namespace (or pod, maybe?). Do we need a separate annotation key for metrics to fully support this?

I would also recommend adding a note in the advanced config doc clarifying the different ways the source type can be set and the order of evaluation. For e.g. something like below (i am not sure if this is correct) -

```
For metrics, the sourcetype is determined in the following order of precedence:

1. If the splunk.com/sourcetype annotation is set on the namespace, its value is used.
2. If .Values.splunkPlatform.sourcetypeMetrics is set in your Helm values, it is used for metrics.
3. If only .Values.splunkPlatform.sourcetype is set, it is used for all data types.
4. If none are set, the default sourcetype is httpevent.
```

omrozowicz-splunk: Ok, I added `sourcetypeMetrics` annotation and updated the docs. Using `sourcetype` for all data types is a bit complicated, but it’s necessary for backward compatibility.

jinja2:
```suggestion
note: Adds configuration option (`sourcetypeMetrics`) for setting the Splunk Platform metrics `sourcetype` via Helm values and annotation.
```

.chloggen/make-metrics-sourcetype-configurable.yaml
```diff
@@ -0,0 +1,12 @@
...
+# One or more tracking issues related to the change
+issues: [1941]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
jinja2:
```suggestion
subtext: |
  - Users can override the metrics sourcetype for Splunk Platform using the `splunk.com/sourcetypeMetrics` annotation or the `.Values.splunkPlatform.sourcetypeMetrics` Helm value.
  - See `advanced-configuration.md` for detailed precedence and usage notes.
```

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -297,6 +297,12 @@ k8sattributes/metrics:
       - key: splunk.com/sourcetype
         tag_name: com.splunk.sourcetype
         from: pod
+      - key: splunk.com/sourcetypeMetrics
+        tag_name: com.splunk.sourcetype
+        from: namespace
+      - key: splunk.com/sourcetypeMetrics
```
jinja2: Feel free to ignore this nit - since we already have `splunk.com/metricsIndex`, it would've been more consistent to go with `splunk.com/metricsSourcetype`

test/splunk_integration_test.go
```diff
@@ -157,22 +157,25 @@ func testVerifyMetricPodAnnotations(t *testing.T) {
...
 	}{
-		{"default index and default sourcetype", "", ""},
-		{"annotation index and default sourcetype", annotationIndex, ""},
-		{"default index and annotation sourcetype", "", annotationSourcetype},
-		{"annotation index and annotation sourcetype", annotationIndex, annotationSourcetype},
+		{"default index and default sourcetype", "", "", ""},
+		{"annotation index and default sourcetype", annotationIndex, "", ""},
+		{"default index and annotation sourcetype", "", annotationSourcetype, ""},
+		{"default index, annotation sourcetype and annotation metric sourcetype", "", annotationSourcetype, annotationMetricSourcetype},
+		{"annotation index and annotation sourcetype", annotationIndex, annotationSourcetype, ""},
```
jinja2, line 175: Should we complete the test matrix by including only 1. metric sourcetype 2. combo of ns/pod annotations to verify precedence? You might be testing it already, i am only going off these test descriptions :)

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
omrozowicz-splunk: ok, ignored as suggested here https://github.com/signalfx/splunk-otel-collector-chart/pull/1941#discussion_r2301148894

helm-charts/splunk-otel-collector/templates/config/_common.tpl
omrozowicz-splunk: also added this to existing code for `.Values.splunkPlatform.sourcetype`

test/splunk_integration_test.go
omrozowicz-splunk: Hey, I think you are right I was missing only metric sourcetype 😅 I added it now

PR #1943, opened 2025-07-30
deprecate Values.service config

jinja2:

Deprecate the `service` config in favor of `.agent.service` and `.gateway.service`
We don't set the `service` map in default values.yaml anymore. When `service` config is detected, it is used to set the service config for gateway/agent like original behavior (it is not merged).


.chloggen/deprecate-service.yaml
```diff
@@ -0,0 +1,10 @@
+
+change_type: deprecation
+component: chart
+note: "Deprecated root-level 'service' config in favor of 'agent.service' and 'gateway.service'."
+issues: [1943]
+subtext: |
+  - The root-level 'service' parameter is now deprecated and will be removed in a future release.
+  - Please use 'agent.service' and 'gateway.service' for configuring services for the agent and gateway, respectively.
+  - Helm chart gives preference to the old config for backward compatibility for now. Users are encouraged to migrate to the new config.
```
jinja2: On 2nd thought, I think it would be better to mergeOverwrite the 2 configs with the gateway/agent service being preferred. This would make migration easier i think

jinja2: keeping the original

PR #1944, opened 2025-07-30
enable histograms from control plane

jinja2:

This PR changes the default value of useControlPlaneMetricsHistogramData and enableEKSApiServerMetrics feature gates to true.


.chloggen/histograms-cp.yaml
```diff
@@ -0,0 +1,13 @@
+change_type: enhancement
```
jinja2: should i make this a breaking change instead? there is a way for users to revert so enhancement feels appropriate

atoulme: it's a breaking change imo.

.chloggen/histograms-cp.yaml
```diff
@@ -0,0 +1,12 @@
...
```
jinja2:
```suggestion
change_type: breaking
```

PR #1973, opened 2025-08-14
[operator] update subchart to version 0.92.4

atoulme:


dmitryax:
Should be 0.92.5 to pick up the schema file fix

dmitryax (approved):

PR #1977, opened 2025-08-18
Fix metrics annotations

omrozowicz-splunk:
**Description:** 
Currently setting `splunk.com/metricsIndex` and `splunk.com/sourcetype` in `k8sattributes/metrics` processor doesn't work at all on a pod level.

This PR fixes this problem and adds integration tests for setting annotations on a pod level.

### DIFF
--- a/test/splunk_integration_test.go
+++ b/test/splunk_integration_test.go
@@ -169,7 +169,6 @@ func testVerifyMetricPodAnnotations(t *testing.T) {
 		{"annotation index and default sourcetype", annotationIndex, ""},
 		{"default index and annotation sourcetype", "", annotationSourcetype},
 		{"annotation index and annotation sourcetype", annotationIndex, annotationSourcetype},
-		{"annotation index and annotation sourcetype in both namespace and pod", annotationIndex, annotationSourcetype},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {


test/splunk_integration_test.go
```diff
@@ -137,6 +138,51 @@ func testVerifyMetricNamespaceAnnotations(t *testing.T) {
...
+				sourcetype = tt.annotationSourcetypeValue
+			}
+			searchQuery := METRIC_SEARCH_QUERY_STRING + "index=" + index + " filter=\"sourcetype=" + sourcetype + "\" | search \"k8s.pod.name\"=" + podName
+			fmt.Println("Search Query: ", searchQuery)
+			startTime := "-15s@s"
+			events := CheckEventsFromSplunk(searchQuery, startTime)
+			fmt.Println(" =========>  Events received: ", len(events))
+			assert.Greater(t, len(events), 1)
+
+			removeAllPodsAnnotations(t, client, podName, namespace)
```
pszkamruk-splunk, line 192: I think we should also remove namespace annotations here, as we are changing them during tests

test/splunk_integration_test.go
```diff
@@ -137,6 +138,51 @@ func testVerifyMetricNamespaceAnnotations(t *testing.T) {
 	}
 }
 
+func testVerifyMetricPodAnnotations(t *testing.T) {
+	podName := "pod-metrics-index"
```
pszkamruk-splunk: nit:
could we change this to something like `pod-for-metric-annotation`

test/splunk_integration_test.go
omrozowicz-splunk: but we change namespace annotations in testVerifyMetricNamespaceAnnotations, in testVerifyMetricPodAnnotations we change only pod annotations

examples/disable-persistence-queue-traces/rendered_manifests/configmap-agent.yaml
```diff
@@ -183,20 +183,15 @@ data:
...
         - sources:
           - from: resource_attribute
-            name: k8s.pod.ip
+            name: k8s.namespace.name
         - sources:
           - from: resource_attribute
-            name: ip
-        - sources:
-          - from: connection
+            name: k8s.node.name
```
sbylica-splunk: Would it be ok to change this pod_association to match the one in 'k8sattributes'? Or would that break something? I think it would be best to keep it consistent if possible.

omrozowicz-splunk: Probably you're right, but I wouldn't like to provide such changes in scope of a fix PR. For some reason it works as it is for logs 😅 Maybe we can introduce another task to check how best we can proceed with consistency

test/splunk_integration_test.go
sbylica-splunk: I don't think it would be that much different to change it completely tbh.
But if we wan't to change it just a little then I think that we should keep
```    
  - sources:
    - from: resource_attribute
      name: k8s.pod.ip
```
as it is, since we have an association like this in logs already (that is if it still works with this kind of configuration).

omrozowicz-splunk: @sbylica-splunk yeah I still don't know why it is there as there's nothing like that in log's attributes

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -277,10 +277,13 @@ k8sattributes/metrics:
   pod_association:
     - sources:
       - from: resource_attribute
-        name: k8s.node.name
+        name: k8s.pod.uid
     - sources:
       - from: resource_attribute
-        name: k8s.pod.uid
+        name: k8s.namespace.name
```
jinja2, line 283: I don't think we should have namespace and node by themself in any of our k8sattributes processors. Just by themself, the processor can associate any pod which has same namespace or node.

These should have a 2nd condition like the `k8s.pod.name` + `k8s.node.name`.

Is below still relevant? I have not seen a resource_attr `ip` so wonder if this is some old attr replaced by k8s.pod.ip?
```
    - sources:
      - from: resource_attribute
        name: ip
 ```

I believe we'll have to make similarly fixes for the k8sattributes processor in logs pipeline.

Comment (omrozowicz-splunk) [reply helm-charts/splunk-otel-collector/templates/config/_common.tpl:283]: Yes, you're right, I was confused on how it works exactly, probably `pod.uid` was matched and that's why it worked. I changed it to what you suggested and deleted `ip` as I'm sure nothing like that exists.

Comment (omrozowicz-splunk) [reply helm-charts/splunk-otel-collector/templates/config/_common.tpl:283]: Also — `k8s.pod.ip` is not something that I've seen in any of our data samples. I deleted it at first but then but I saw it is in `expected_*_traces.yaml` so decided to leave it as it is. Do you know if it's necessary?

Comment (jinja2) [reply helm-charts/splunk-otel-collector/templates/config/_common.tpl:283]: I would suggest to leave it in. It is easy for an app to set it with the k8s downward api. Can you confirm it is usable in pod_association from the k8sattributes docs also?

File: .chloggen/fix-metrics-annotations.yaml [hunk 5]
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: chart
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix metrics annotations in the chart to work on a  pod level
```
jinja2:
```suggestion
note: Fix per-pod configuration of 'metricsIndex' and 'sourcetype' config for Splunk Platform metrics via annotations
```

omrozowicz-splunk: done

pszkamruk-splunk (approved):

PR #2006, opened 2025-08-27
[OTL-3712] Remove unneeded Java and .Net resource attributes

dloucasfx:


.chloggen/InstrumentationResourceAtt.yaml
```diff
@@ -0,0 +1,15 @@
...
+note: Limit Java and .Net auto-instrumentation resource attributes to only `Container` and `Environment` resource providers
+# One or more tracking issues related to the change
+issues: [2006]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This change limits the resource attributes added to Java and .Net spans and metrics to the only attributes generated by `Container` and `Environment` resource providers.
+  To enable additional resource providers, update `values.yaml` environment variable `OTEL_JAVA_ENABLED_RESOURCE_PROVIDERS` for Java, or set `OTEL_DOTNET_AUTO_{DETECTOR}_RESOURCE_DETECTOR_ENABLED` for .Net.
+  For more information on configuring these environment variables, see: https://help.splunk.com/en/splunk-observability-cloud/manage-data/available-data-sources/supported-integrations-in-splunk-observability-cloud/apm-instrumentation
```
dmitryax: We should probably mention that we are doing it to avoid running into the limits

dmitryax (approved):
Nice!

functional_tests/functional/testdata/expected_kind_values/expected_dotnet_traces.yaml
```diff
@@ -3,49 +3,16 @@ resourceSpans:
...
-        - key: os.build_id
-          value:
-            stringValue: 6.10.14-linuxkit
-        - key: os.name
-          value:
-            stringValue: Alpine Linux
-        - key: os.version
-          value:
-            stringValue: 3.19.1
-        - key: host.name
```
jinja2, line 28: Q - shouldn't host name still be there? Or maybe not, is our test setup with agent or gateway as the instrumentation endpoint?

Can we document this? I think this change will have different attrs going missing depending on the endpoint config of users. If you could document which resource detectors will not be available by default for endpoint combo, it would be easier for users to add back anything important - for e.g. folks sending to gateway most likely should enable the host resource detector (in addition to charts/detectors being impacted, it could also impact billing since w/o any host attr it won't be host categorized).

jinja2 (dismissed):

.chloggen/InstrumentationResourceAtt.yaml
dloucasfx: agree, I am thinking of adding a section in auto-instrumentation docs talking about this in details with examples on how to add remove

functional_tests/functional/testdata/expected_kind_values/expected_dotnet_traces.yaml
dloucasfx: so host.name from the instrumented lib, is not needed as it will be added by the collector resourcedetection, which is more accurate then the one added by the span.

> Can we document this? I think this change will have different attrs going missing depending on the endpoint config of users. If you could document which resource detectors will not be available by default for endpoint combo, it would be easier for users to add back anything important - for e.g. folks sending to gateway most likely should enable the host resource detector (in addition to charts/detectors being impacted, it could also impact billing since w/o any host attr it won't be host categorized).

Yes, I am thinking of adding a section in auto-instrumentation docs talking about this in details with examples on how to add remove

dmitryax: Good catch. This test result misses attributes that are supposed to be set by [the resource detection processor](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl#L1142) in the agent. Not sure what's up. We need to figure that out

dloucasfx: host.name has NOT been deleted , it was just moved to a different part of the file

jinja2: Ah, that makes sense. The docs lgtm except for the issue that host will be missing (?) when sent to gateway/ingest. Is this something we should explicitly instruct user to enable for billing purposes? I am actual not sure if these are host or bundled cat.

PR #2009, opened 2025-08-28
update coredns scrape config for openshift

jinja2:

Noticed openshift runs dns container with kube-rbac-proxy. The PR updates the scraper config for prometheus/coredns for OpenShift to use https and authenticate correctly. The `insecure_skip_verify` is added due to the cert not having the pod IP in CN/SAN configs. 

Ran in a test openshift cluster. With the new changes, coredns metrics are exported correctly.

### DIFF
--- a/.chloggen/openshift-coredns.yaml
+++ b/.chloggen/openshift-coredns.yaml
@@ -1,0 +1,13 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update prometheus receiver configuration for CoreDNS in OpenShift to use secure endpoint and service account token
+# One or more tracking issues related to the change
+issues: []
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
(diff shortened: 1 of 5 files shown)


.chloggen/openshift-coredns.yaml
```diff
@@ -0,0 +1,13 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update prometheus receiver configuration for CoreDNS in OpenShift to use secure endpoint and service account token
+# One or more tracking issues related to the change
+issues: []
```
jinja2:
```suggestion
issues: [2009]
```

jinja2:
For reference, this was tested in a ROSA cluster with below version
```
❯ oc get clusterversion
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.19.7    True        False         26h     Cluster version is 4.19.7
```

.chloggen/openshift-coredns.yaml
```diff
@@ -0,0 +1,13 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
```
jinja2, line 2: Made it an enhancement instead of bug_fix cause i don't know when this change was introduced in openshift and i tested against the very latest. Lmk if bug_fix would be better

pjanotti (approved):
LGTM

PR #2012, opened 2025-08-29
[agent] fix for situation where multiline logs are split into multiline events 

pszkamruk-splunk:
**Description:**
In some cases, customers observe that multiline logs are split into multiple events.
This occurs when attempting to recombine more than 1000 log entries into a single event. 
To address this, we are introducing the maxNumOfLinesToCombine parameter under the multilineConfigs section. This parameter allows specifying the maximum number of lines that should be combined into one log entry.


.chloggen/fix-combine-multiline-logs-which-have-more-than-1000-lines.yaml
```diff
@@ -0,0 +1,14 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: |
+  Allow to specify max number of lines which will be combined in log entry for multiline log configuration.
```
jinja2: Could you include the config option (full field path) in the changelog?

pszkamruk-splunk: done

PR #2020, opened 2025-09-03
Add k8sobjects logs under watch mode

tgtanyagoel:
Add k8sobjects [cronjobs, daemonsets, deployments, jobs, namespaces, nodes, pods, replicasets, statefulsets] under watch mode

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,3 +16,6 @@ updates:
       otel:
         patterns:
           - "go.opentelemetry.io/otel*"
+      kubernetes-ecosystem:
+        patterns:
+          - "k8s.io/*"


helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -386,6 +386,7 @@ service:
     logs:
       receivers:
         - k8s_events
+        - k8sobjects
```
jinja2: why are we adding this receiver in this pipelines? Given this is not a new config in values.yaml, i expect there to already be a pipeline using this receiver conditionally.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -566,7 +566,25 @@ clusterReceiver:
   #
   # The configuration format in details is described here:
   # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver
-  k8sObjects: []
+  k8sObjects:
+    - name: cronjobs
+      mode: watch
```
jinja2: Can you provide some info in the PR description on expected content? That would help reviewers to give better input on config. Like why watch vs pull?

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
pszkamruk-splunk: +1 to @jinja2 comment 

we have separate pipeline for objects: [link](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl#L410)

I think we should not add the `k8sobjects` receiver here, as this might impact existing customer data pipelines and cause some confusion. I think event and object pipelines should remain separate, as they are configured now.

.chloggen/k8sobjects.yaml
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: clusterReceiver
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add k8sobjects [cronjobs, daemonsets, deployments, jobs, namespaces, nodes, pods, replicasets, statefulsets] logs under watch mode
+# One or more tracking issues related to the change
+issues: [2013]
```
pszkamruk-splunk: nit:
```suggestion
issues: [2020]
```

.chloggen/k8sobjects.yaml
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: clusterReceiver
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add k8sobjects [cronjobs, daemonsets, deployments, jobs, namespaces, nodes, pods, replicasets, statefulsets] logs under watch mode
```
pszkamruk-splunk: I would also add information that collecting these objects will be enabled by default.

PR #2027, opened 2025-09-08
Remove usage of Values.Release.Service in templates

dmitryax:
- Remove all "heritage" labels
- Set all "app.kubernetes.io/managed-by" labels to "Helm"

This is needed to avoid making extra changes to run the chart as an EKS add-on which doesn't support Values.Release.Service


.chloggen/remove-usage-of-release-service.yaml
```diff
@@ -0,0 +1,14 @@
...
+component: all
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update labels in all resources
+# One or more tracking issues related to the change
+issues: [2027]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  - Remove all "heritage" labels
```
jinja2:
```suggestion
  - Remove "heritage" label from all objects managed by the chart.
```

PR #2049, opened 2025-09-08
Remove cert-manager subchart

dmitryax:
- cert-manager is no longer included as a subchart dependency
- Users who need cert-manager for webhook certificates must install it separately

### DIFF
--- a/.chloggen/remove-cert-manager.yaml
+++ b/.chloggen/remove-cert-manager.yaml
@@ -1,0 +1,14 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: operator
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove cert-manager subchart
+# One or more tracking issues related to the change
+issues: [2049]
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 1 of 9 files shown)


.chloggen/remove-cert-manager.yaml
```diff
@@ -0,0 +1,14 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove cert-manager subchart
+# One or more tracking issues related to the change
+issues: [2049]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  - cert-manager is no longer included as a subchart dependency
+  - Users who need cert-manager for webhook certificates must install it separately
```
crobert-1: Would it be helpful if we include more specific directions? Maybe a link to the [installation docs](https://cert-manager.io/docs/installation/) (as I see you've included in another doc.)

crobert-1 (commented):
I see a few remaining `cert-manager` references in `functional_tests/functional/functional_test.go`, I'm not sure if they're still needed or if we can delete them.

crobert-1 (approved):

.chloggen/remove-cert-manager.yaml
dmitryax: The instructions are included in docs/auto-instrumentation-install.md, added the link here

dmitryax:
> I see a few remaining cert-manager references in functional_tests/functional/functional_test.go, I'm not sure if they're still needed or if we can delete them.

Thanks for catching this. I removed them

PR #2065, opened 2025-09-16
[chore] Add script for publishing EKS Add-on artifacts

dmitryax:
The script to push artifacts for EKS Add-on to ECR. Has to be run manually since it requires AWS authentication with Okta

### DIFF
--- a/eks_addon/overrides/values.yaml
+++ b/eks_addon/overrides/values.yaml
@@ -1,0 +1,24 @@
+# EKS Add-on requires minimal configuration for the Helm chart to be rendered correctly.
+# splunkObservability.realm and splunkObservability.accessToken must be set by the user or
+# set to empty strings if splunkPlatform is used instead.
+splunkObservability:
+  realm: CHANGEME
+  accessToken: CHANGEME
+
+# Default values for running in AWS EKS
+cloudProvider: aws
+distribution: eks
+
+# Only ECR repositories can be used for EKS Add-on images.
+image:
+  otelcol:
+    repository: ${ECR_OTELCOL_REPO}
+  fluentd:
+    repository: ${ECR_FLUENTD_REPO}
+    tag: ${ECR_FLUENTD_REPO_TAG}
+  initPatchLogDirs:
+    repository: public.ecr.aws/docker/library/almalinux
+    tag: "9"
+  validateSecret:
+    repository: public.ecr.aws/docker/library/almalinux
+    tag: "9"
--- a/eks_addon/release.sh
+++ b/eks_addon/release.sh
@@ -1,0 +1,173 @@
+#!/bin/bash
+# Builds, packages, and pushes artifacts for the Splunk OpenTelemetry Collector EKS Add-on.
+#
+# Parameters:
+#   --dry-run : (Optional) Enables dry-run mode that skips pushing artifacts.
+#   --chart-version VERSION : (Optional) Overrides the chart version from Chart.yaml.
+#
+# Note: This script requires OKTA_AWS_ROLE_ARN to be set in the environment.
+
+# Enable bash strict mode to fail fast
+set -euo pipefail
+
+# Check required tools
+for tool in "okta-aws-login" "aws" "yq" "helm" "docker"; do
+  command -v "${tool}" &>/dev/null || { echo "❌ Required command '${tool}' is not installed or not in PATH"; exit 1; }
+done
+
+# Parse command line arguments
+DRY_RUN_PREFIX=""
+CHART_VERSION=""
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --dry-run)
+      DRY_RUN_PREFIX="echo 🚧 [DRY-RUN] "
+      shift
+      ;;
+    --chart-version)
+      CHART_VERSION="$2"
+      echo "ℹ️ Chart version override provided: $CHART_VERSION"
+      shift 2
+      ;;
+    *)
+      echo "❌ Unknown argument: $1"
+      exit 1
+      ;;
+  esac
+done
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+CHART_DIR="$SCRIPT_DIR/../helm-charts/splunk-otel-collector"
+if [[ -z "$CHART_VERSION" ]]; then
+  CHART_VERSION=$(yq e ".version" "${CHART_DIR}/Chart.yaml")
+fi
+CHART_APPVERSION=$(yq e ".appVersion" "${CHART_DIR}/Chart.yaml")
+ECR_REGION="us-east-1"
+ECR_REGISTRY="709825985650.dkr.ecr.${ECR_REGION}.amazonaws.com"
+ECR_OTELCOL_REPO="${ECR_REGISTRY}/splunk/images/splunk-otel-collector"
+ECR_FLUENTD_REPO="${ECR_REGISTRY}/splunk/docker.io/splunk/fluentd-hec"
+ECR_FLUENTD_REPO_TAG="1.3.3-linux"
+ECR_HELM_NAMESPACE="${ECR_REGISTRY}/splunk/charts"
+ECR_HELM_REPO="${ECR_HELM_NAMESPACE}/splunk-otel-collector"
+
+aws_okta_auth() {
+  echo "🌐 Authenticating with Okta (browser will open)..."
+  okta-aws-login > /tmp/okta_creds.txt 2>&1 || { echo "❌ AWS Authentication failed"; rm -f /tmp/okta_creds.txt; exit 1; }
+
+  # Extract and evaluate AWS credentials
+  eval "$(grep -E "^(AWS_|export)" /tmp/okta_creds.txt)"
+  rm -f /tmp/okta_creds.txt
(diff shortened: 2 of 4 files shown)


eks_addon/overrides/values.yaml
crobert-1, line 1: If there's any expectation of a user manually modifying this file it would probably be good to document required environment variables in this file's header comment, or somewhere that can be referenced.

eks_addon/overrides/values.yaml
```diff
@@ -0,0 +1,24 @@
+# EKS Add-on requires minimal configuration for the Helm chart to be rendered correctly.
+# splunkObservability.realm and splunkObservability.accessToken must be set by the user or
+# set to empty strings if splunkPlatform is used instead.
+splunkObservability:
+  realm: us0
```
crobert-1:
```suggestion
  realm: CHANGEME
```
Should we have a default here like `us0`, or should we require users to specify the value?

eks_addon/overrides/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+  accessToken: CHANGE_ME
```
crobert-1:
```suggestion
  accessToken: CHANGEME
```
Nit: Existing examples all use `CHANGEME`

eks_addon/overrides/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+    repository: ${ECR_OTELCOL_REPO}
+  fluentd:
+    repository: ${ECR_FLUENTD_REPO}
+    tag: ${ECR_FLUENTD_REPO_TAG}
+  initPatchLogDirs:
+    repository: public.ecr.aws/docker/library/almalinux
+    tag: "9"
+  validateSecret:
+    repository: public.ecr.aws/docker/library/almalinux
+    tag: "9"
```
crobert-1, line 24: Would it be helpful to use more specific tag values here for future reference and upgrade clarity?

eks_addon/release.sh
```diff
@@ -0,0 +1,174 @@
+#!/bin/bash
+# Builds, packages, and pushes artifacts for the Splunk OpenTelemetry Collector EKS Add-on.
```
crobert-1, line 2:
```suggestion
# Builds, packages, and pushes artifacts for the Splunk Distribution of the OpenTelemetry Collector EKS Add-on.
```
Nit: [Following docs](https://help.splunk.com/en/splunk-observability-cloud/manage-data/splunk-distribution-of-the-opentelemetry-collector/get-started-with-the-splunk-distribution-of-the-opentelemetry-collector/collector-for-kubernetes/kubernetes-eks-add-on) wording, if that's the right reference for this.

eks_addon/release.sh
```diff
@@ -0,0 +1,174 @@
...
+  eval "$(grep -E "^(AWS_|export)" /tmp/okta_creds.txt)"
+  rm -f /tmp/okta_creds.txt
+  AUTH_AWS_ECR_PASSWORD=$(aws ecr get-login-password --region "$ECR_REGION")
+  [[ -n "$AUTH_AWS_ECR_PASSWORD" ]] || { echo "❌ Failed to get ECR login password."; exit 1; }
+
+  # Docker login to default config - set -e will fail on error
+  echo "$AUTH_AWS_ECR_PASSWORD" | docker login --username AWS --password-stdin "$ECR_REGISTRY"
+
+  # Helm login - set -e will fail on error
+  echo "$AUTH_AWS_ECR_PASSWORD" | helm registry login --username AWS --password-stdin "$ECR_REGISTRY"
```
crobert-1: I don't know if these comments are super helpful, other than the reminder about `set -e` functionality. `docker login` and `helm registry login` are pretty straightforward.

eks_addon/release.sh
```diff
@@ -0,0 +1,174 @@
...
+  echo "✅ AWS Authentication successful"
+}
+
+copy_docker_image_to_ecr() {
+  # Get the otelcol repository from values.yaml
+  local src_repo=$(yq e ".image.otelcol.repository" "${CHART_DIR}/values.yaml")
+  [[ -n "$src_repo" && "$src_repo" != "null" ]] || { echo "❌ Error: Could not find otelcol repository in values.yaml"; exit 1; }
+
+  local src="${src_repo}:${CHART_APPVERSION}"
+  local dst="${ECR_OTELCOL_REPO}:${CHART_APPVERSION}"
```
crobert-1: Nit: `dest` might be slightly more clear than `dst`, not a big deal with this small of a method though.

eks_addon/release.sh
```diff
@@ -0,0 +1,174 @@
...
+ECR_REGION="us-east-1"
+ECR_OTELCOL_REPO="${ECR_REGISTRY}/splunk/images/splunk-otel-collector"
+ECR_FLUENTD_REPO="${ECR_REGISTRY}/splunk/docker.io/splunk/fluentd-hec"
+ECR_FLUENTD_REPO_TAG="1.3.3-linux"
+ECR_HELM_NAMESPACE="${ECR_REGISTRY}/splunk/charts"
+ECR_HELM_REPO="${ECR_HELM_NAMESPACE}/splunk-otel-collector"
+
+aws_okta_auth() {
+  echo "🌐 Authenticating with Okta (browser will open)..."
+  okta-aws-login > /tmp/okta_creds.txt 2>&1 || { echo "❌ AWS Authentication failed"; rm -f /tmp/okta_creds.txt; exit 1; }
```
crobert-1, line 55: This command is hanging for me, even after the browser says I've successfully authenticated:
```
$ ./eks_addon/release.sh 
🌐 Authenticating with Okta (browser will open)...
```
Any debugging information or comments we could add to clarify usage?

eks_addon/release.sh
```diff
@@ -0,0 +1,174 @@
...
+    return 1
+  fi
+
+  # Copy chart to a temporary build directory
+  local build_dir="${SCRIPT_DIR}/build"
+  rm -rf "${build_dir}"
+  mkdir -p "${build_dir}"
+  cp -R "${CHART_DIR}" "${build_dir}/"
+
+  # Function to modify the Helm chart to meet EKS Add-on requirements
```
crobert-1: It might be better to move this comment to the method itself, and make it a bit more descriptive of why the modifications are needed.

crobert-1 (commented):
I'm not very familiar with this area, mostly just nits or clarifications.

Unless there's some reason not to, a README under `eks_addon` would be helpful to communicate its purpose and how to use it.

eks_addon/release.sh
```diff
@@ -0,0 +1,174 @@
...
+done
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+CHART_DIR="$SCRIPT_DIR/../helm-charts/splunk-otel-collector"
+if [[ -z "$CHART_VERSION" ]]; then
+  CHART_VERSION=$(yq e ".version" "${CHART_DIR}/Chart.yaml")
+fi
+CHART_APPVERSION=$(yq e ".appVersion" "${CHART_DIR}/Chart.yaml")
+ECR_REGISTRY="709825985650.dkr.ecr.us-east-1.amazonaws.com"
+ECR_REGION="us-east-1"
```
jinja2: AWS_ACCOUNT_ID="${AWS_ACCOUNT_ID:-709825985650}"

```suggestion
ECR_REGION="${ECR_REGION:-us-east-1}"
ECR_REGISTRY="709825985650.dkr.ecr.${ECR_REGION}.amazonaws.com"
```
also, this account id is safe to have in the repo? is this the ecr url used by addon users?

General question - will aws make the image available in other regions?

eks_addon/overrides/values.yaml
dmitryax: I'm fine with any. Let's use CHANGEME

eks_addon/overrides/values.yaml
dmitryax: I'm not sure about it. The scripts are very basic. So chances that minor version upgrade can affect them is extremely low. So I would keep it to the major version. We use `latest` in values.yaml

eks_addon/release.sh
dmitryax: Yeah. It's expected. It should be run with `OKTA_AWS_ROLE_ARN` env var set. It'll be part of the release doc

eks_addon/release.sh
dmitryax: > General question - will aws make the image available in other regions?

I guess there is some sort of implicit CDN in place. But `709825985650.dkr.ecr.us-east-1.amazonaws.com` is all over the AWS docs. And we don't have write access to any other regions anyway

dmitryax: > also, this account id is safe to have in the repo? is this the ecr url used by addon users?

Yes, it's the common account for addon users

dmitryax: > ECR_REGION="${ECR_REGION:-us-east-1}"

I'm don't see any use case when the ECR region would need to be overridden. I made it `ECR_REGION` used in `ECR_REGISTRY`  as you suggested though

eks_addon/overrides/values.yaml
dmitryax: done

eks_addon/release.sh
jinja2: AWS making it available in different regions would be ideal. I was worried we'd have to run this per-region. Thanks

PR #2067, opened 2025-09-18
fix missing receiver_creator receiver

omrozowicz-splunk:
**Description:** In https://github.com/signalfx/splunk-otel-collector-chart/pull/1944 `useControlPlaneMetricsHistogramData` was enabled by default, what made `receiver_creator` receiver disappear from the default `metrics` pipeline. This is the code that caused this:

```
    metrics:
      receivers:
        - hostmetrics
        - kubeletstats
        - otlp
        {{- if not .Values.featureGates.useControlPlaneMetricsHistogramData }}
        - receiver_creator
        {{- end }}
```

Currently, when `autodetect.prometheus` is set to `true`, `receiver_creator` receiver is created in the configuration and not included in the pipeline.

**Testing:** Manual testing

**Documentation:** N/A

### DIFF
--- a/examples/disable-persistence-queue-traces/rendered_manifests/configmap-agent.yaml
+++ b/examples/disable-persistence-queue-traces/rendered_manifests/configmap-agent.yaml
@@ -454,6 +454,7 @@ data:
           - hostmetrics
           - kubeletstats
           - otlp
+          - receiver_creator
           - signalfx
         metrics/agent:
           exporters:
--- a/examples/disable-persistence-queue-traces/rendered_manifests/daemonset.yaml
+++ b/examples/disable-persistence-queue-traces/rendered_manifests/daemonset.yaml
@@ -31,7 +31,7 @@ spec:
         component: otel-collector-agent
         release: default
       annotations:
-        checksum/config: ec28b03a604898417e168b73e108ca0d2f0790694cfc683882d9294588725b61
+        checksum/config: de11788e52330431dc6c4f43192868d45a5df8dd2e89fecf43bd46a027efc679
         kubectl.kubernetes.io/default-container: otel-collector
     spec:
       hostNetwork: true
--- a/examples/enable-persistence-queue/rendered_manifests/configmap-agent.yaml
+++ b/examples/enable-persistence-queue/rendered_manifests/configmap-agent.yaml
@@ -454,6 +454,7 @@ data:
           - hostmetrics
           - kubeletstats
           - otlp
+          - receiver_creator
           - signalfx
         metrics/agent:
           exporters:
--- a/examples/enable-persistence-queue/rendered_manifests/daemonset.yaml
+++ b/examples/enable-persistence-queue/rendered_manifests/daemonset.yaml
@@ -31,7 +31,7 @@ spec:
         component: otel-collector-agent
         release: default
       annotations:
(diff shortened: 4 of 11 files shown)


jinja2:
Good catch! But i think we have more configs to check for in this. There is also the autodetect.istio. And then there is the complex situation of both the autodect + controlPlaneHistograms being enabled which right now will add same receivers to both pipelines.

omrozowicz-splunk:
> Good catch! But i think we have more configs to check for in this. There is also the autodetect.istio. And then there is the complex situation of both the autodect + controlPlaneHistograms being enabled which right now will add same receivers to both pipelines.

Hey @jinja2, you're right! I added `autodetect.istio` to the condition. 

> complex situation of both the autodect + controlPlaneHistograms being enabled

can you explain what part exactly you mean? because the condition currently is:

```
{{- if or (not .Values.featureGates.useControlPlaneMetricsHistogramData) .Values.autodetect.prometheus .Values.autodetect.istio }}
```

so this will be rendered when useControlPlaneMetricsHistogramData is **not** enabled.

jinja2:
> > Good catch! But i think we have more configs to check for in this. There is also the autodetect.istio. And then there is the complex situation of both the autodect + controlPlaneHistograms being enabled which right now will add same receivers to both pipelines.
> 
> Hey @jinja2, you're right! I added `autodetect.istio` to the condition.
> 
> > complex situation of both the autodect + controlPlaneHistograms being enabled
> 
> can you explain what part exactly you mean? because the condition currently is:
> 
> ```
> {{- if or (not .Values.featureGates.useControlPlaneMetricsHistogramData) .Values.autodetect.prometheus .Values.autodetect.istio }}
> ```
> 
> so this will be rendered when useControlPlaneMetricsHistogramData is **not** enabled.

@omrozowicz-splunk isn't the condition just 3 ORs? It looks like `receiver_creator` will be added to the metrics pipeline when any of these is true:
1. Histogram is disabled, OR
1. Prometheus autodetect is enabled, OR
1. Istio autodetect is enabled

And when both `useControlPlaneMetricsHistogramData` and `autodetect` features are enabled, we get the same `receiver_creator` configuration deployed to both the regular `metrics` pipeline and the `metrics/histogram` pipeline.

omrozowicz-splunk:
Ok, now I think also those conditions are wrong. For example:
```
{{- if and (or (eq .Values.distribution "openshift") (eq .Values.distribution "")) (not (.Values.featureGates.useControlPlaneMetricsHistogramData)) }}
```
Previously there was no `not (.Values.featureGates.useControlPlaneMetricsHistogramData)` and these was rendered by default for Splunk Platform. I think we should take `.Values.featureGates.useControlPlaneMetricsHistogramData` into consideration only when o11y is enabled and create something like this:
```
{{- define "splunk-otel-collector.useControlPlaneMetricsHistogramDataCondition" -}}
{{- or
      (and
        (not .Values.featureGates.useControlPlaneMetricsHistogramData)
        (eq (include "splunk-otel-collector.splunkO11yEnabled" .) "true")
      )
      (eq (include "splunk-otel-collector.splunkO11yEnabled" .) "false")
-}}
{{- end -}}
```

@jinja2, please let me know if this makes sense. I might be missing context on whether disabling smartagent/* receivers was intentional

PR #2072, opened 2025-09-22
[chore] fix flaky functional tests

omrozowicz-splunk:
**Description:** Functional tests (`functional`) are pretty unstable, empirically I noticed it passes every time when I introduce time or add a few retries.

You can see how it fails on a release PR: https://github.com/signalfx/splunk-otel-collector-chart/pull/2068

**Testing:** N/A

**Documentation:** N/A


omrozowicz-splunk:
it is indeed a change based on empiric testing: it was passing every time on a fork's release branch with this change. It would be better to fix whatever is really wrong with the test logic than provide such a workaround (without the certainty that It'll work every time). Closing this.

PR #2076, opened 2025-09-23
[OTL-3800] Adjust log batch settings

dloucasfx:

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_common.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_common.tpl
@@ -456,12 +456,12 @@ splunk_hec/platform_logs:
     storage: file_storage/persistent_queue
     {{- end }}
     num_consumers: {{ .Values.splunkPlatform.sendingQueue.numConsumers }}
-  {{- if .Values.featureGates.noDropLogsPipeline }}
-  batcher:
-    enabled: true
-    flush_timeout: 200ms
-    min_size: 2048
-  {{- end }}
+    {{- if .Values.featureGates.noDropLogsPipeline }}
+    batch:
+      flush_timeout: 200ms
+      min_size: 2048
+      sizer: items
+    {{- end }}
 {{- end }}
 
 {{/*
--- a/.chloggen/sendingQueueBatch.yaml
+++ b/.chloggen/sendingQueueBatch.yaml
@@ -1,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: log
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update `sending_queue` batch settings when `noDropLogsPipeline` is enabled
+# One or more tracking issues related to the change
+issues: [2076]
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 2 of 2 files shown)


helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -456,12 +456,12 @@ splunk_hec/platform_logs:
     storage: file_storage/persistent_queue
     {{- end }}
     num_consumers: {{ .Values.splunkPlatform.sendingQueue.numConsumers }}
-  {{- if .Values.featureGates.noDropLogsPipeline }}
-  batcher:
```
crobert-1, line 460: Upstream documentation doesn't make it clear that `batcher` is deprecated in favor of `batch`. Would you be able to open a PR [updating the README](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/256b4786593e97918a3140a4326041ab3768c9af/exporter/splunkhecexporter/README.md#L78) to make it more clear?

dloucasfx: updating the hec exporter readme is on my to do list, but yeah the deprecation is in the code
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/splunkhecexporter/config.go#L81-L82
and this is the PR https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/7e05ef93c7a921354417fccec5338ebd4d924811

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -456,12 +456,12 @@ splunk_hec/platform_logs:
...
-    enabled: true
-    flush_timeout: 200ms
-    min_size: 2048
-  {{- end }}
+    {{- if .Values.featureGates.noDropLogsPipeline }}
+    batch:
+      flush_timeout: 200ms
+      min_size: 500
```
dmitryax: Why dropping it to 500? that seem too low

dloucasfx: the queue size is 1000 , I figured half is good for min?

dloucasfx: also note that with sizer set in batch to items (which is needed) the queue size becomes items (no longer requests) , at least my understanding from reading the doc

dmitryax: > the queue size is 1000

those are requests not items

> also note that with sizer set in batch to items (which is needed) the queue size becomes items (no longer requests 

The old `batcher` was using items as well

dloucasfx: > those are requests not items

they will be set to items based on this
````
sizer: Overrides the sizer set at the sending_queue level for batching. Available options:
items: number of the smallest parts of each signal (spans, metric data points, log records);
bytes: the size of serialized data in bytes (the least performant option).
````

dmitryax: `sending_queue` has it's own `sizer` option which is `requests` by default. `batch.sizer` overrides sizer for the batches not for the queue

dloucasfx: Thanks for confirming, adjusted the value. I will, also update the doc upstream.

helm-charts/splunk-otel-collector/templates/config/_common.tpl
dloucasfx: @crobert-1 https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42912

PR #2078, opened 2025-09-24
[chore] Add tests for EKS Addon

dmitryax:


.github/workflows/eks-addon-tests.yaml
```diff
@@ -0,0 +1,34 @@
...
+  pull_request:
+  push:
+    branches: [ main ]
+
+jobs:
+  eks-addon-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # TODO: Use ci-matrix.json instead
```
crobert-1, line 13: It should be pretty simple to just copy+paste the existing [`get-test-matrix` job from other workflows](https://github.com/signalfx/splunk-otel-collector-chart/blob/4e864ab1ae7b1b9cdeaa7925a9c4603e0d9ef81c/.github/workflows/functional_test.yaml#L23), can we just do it now? The follow up could be to move to a reusable job so it's in a central place.

.github/workflows/eks-addon-tests.yaml
```diff
@@ -0,0 +1,34 @@
...
+        k8s_version: ["1.28", "1.29", "1.30", "1.31", "1.32", "1.33"]
+    
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
```
crobert-1: Might be good to add a comment after the hash specifying which version this is

eks_addon/common.sh
```diff
@@ -0,0 +1,73 @@
...
+$(cat "${override}")
+EOF" > "${tmp_file}"
+
+      # Merge override with the corresponding file in the chart
+      yq eval-all 'select(fileIndex==0) * select(fileIndex==1)' -i "${target_file}" "${tmp_file}"
+      rm -f "${tmp_file}"
+    fi
+  done
+
+  echo "⏳ Moving values.schema.json to aws_mp_configuration_schema.json and removing unsupported properties ..."
```
crobert-1:
```suggestion
  echo "⏳ Moving values.schema.json to aws_mp_configuration_schema.json and removing disabled properties ..."
```
Nit: For consistency with wording below, feel free to drop if this is incorrect.

eks_addon/common.sh
```diff
@@ -0,0 +1,73 @@
...
+    "instrumentation"
+    "certmanager"
+    "cert-manager"
+    "targetAllocator"
+  )
+  for prop in "${disabled_properties[@]}"; do
+      yq e "del(.properties.\"${prop}\")" -i "${EKS_CHART_DIR}/aws_mp_configuration_schema.json"
+  done
+
+  echo "✅ Successfully prepared chart for for EKS Add-on at ${EKS_CHART_DIR}"
```
crobert-1:
```suggestion
  echo "✅ Successfully prepared chart for EKS Add-on at ${EKS_CHART_DIR}"
```

eks_addon/common.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+# Common functions for EKS Add-on preparation and testing
+
+# Enable bash strict mode
+set -euo pipefail
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+CHART_DIR="$SCRIPT_DIR/../helm-charts/splunk-otel-collector"
+ORIG_CHART_DIR="$SCRIPT_DIR/../helm-charts/splunk-otel-collector"
```
crobert-1: Do we need both `CHART_DIR` and `ORIG_CHART_DIR`?

eks_addon/common.sh
```diff
@@ -0,0 +1,73 @@
...
+BUILD_DIR="${SCRIPT_DIR}/build"
```
crobert-1, line 9: Is there any potential to conflict in the future where `./build` would be used for something else? I'm wondering if it'd be worthwhile to name it more specifically since we're wiping it out every time the `prepare_chart` function is run.

eks_addon/run_tests.sh
```diff
@@ -0,0 +1,107 @@
...
+  echo "❌ Error: K8S_VERSION environment variable is required"
+  exit 1
+fi
+
+# EKS Add-on specific test functions
+test_helm_lint() {
+  echo "⏳ Running helm lint on ${EKS_CHART_DIR} ..."
+  
+  if ! helm lint "${EKS_CHART_DIR}"; then
+    echo "❌ Helm lint test failed"
```
crobert-1, line 19: Will the test run include the output of `helm lint` in logs? If not, is there some way to log it here? (Same question for `test_helm_template` and others.)

eks_addon/run_tests.sh
```diff
@@ -0,0 +1,107 @@
...
+    echo "❌ Found forbidden Release properties in templates:"
+    echo "${forbidden_release_refs}"
+    return 1
+  fi  
+  echo "✅ No forbidden Release properties found in templates"
+}
+
+prepare_chart
+
+test_helm_lint
```
crobert-1: All of these functions are returning 1 on failure, is that properly bubbled up to the workflow? (I would have expected some check here to confirm return values are 0, but maybe I'm forgetting how bash return codes work.)

eks_addon/run_tests.sh
```diff
@@ -0,0 +1,107 @@
...
+source "${SCRIPT_DIR}/common.sh"
+
+if [[ -z "${K8S_VERSION}" ]]; then
+  echo "❌ Error: K8S_VERSION environment variable is required"
+  exit 1
+fi
+
+# EKS Add-on specific test functions
+test_helm_lint() {
+  echo "⏳ Running helm lint on ${EKS_CHART_DIR} ..."
```
crobert-1: I wonder if it would be helpful at the beginning of each test to also log the function name for easier searching when they fail. I find it a lot easier to search function names in repos vs. plaintext.

.github/workflows/eks-addon-tests.yaml
dmitryax: It's not only that. We need to also update the script that keep the matrix up to date. I'd move that to follow up

eks_addon/common.sh
dmitryax: `$SCRIPT_DIR` is `eks_addon` so it's already specific enough I think

eks_addon/common.sh
dmitryax: Good catch. thanks!

.github/workflows/eks-addon-tests.yaml
dmitryax: Switched to v5. Not sure why we use the sha in other places

eks_addon/run_tests.sh
dmitryax: Yes, both stdout and stderr will be included. We don't pipe it anywhere else

eks_addon/run_tests.sh
dmitryax: Yes, with `set -euo pipefail`, any failed test throws the error

eks_addon/run_tests.sh
dmitryax: I simplified the logs to include the test name only

.github/workflows/eks-addon-tests.yaml
crobert-1: You're saying it would be a new key:value pair in `ci-matrix.json`? (And we can't use an existing one?)

dmitryax: Currently, we have particular group per specific target in `ci-matrix.json`, no generic "Supported k8s versions". Another problem is that not all of them are being watched and updated. I'd like clean it up separately, then adopt here.

PR #2079, opened 2025-09-25
[chore] update testAgentMetrics in functional_tests

jinja2:
Attempts to fix flaky agent metrics tests. I added hostmetrics to its own expected testdata file and re-using function used in istio test to select metric batch.

### DIFF
--- a/functional_tests/functional/functional_test.go
+++ b/functional_tests/functional/functional_test.go
@@ -1125,31 +1125,30 @@ func testAgentMetricsTemplate(t *testing.T, metricsSink *consumertest.MetricsSin
 	expectedMetricsFile := filepath.Join(testDir, expectedValuesDir, expectedFileName)
 	expectedMetrics, err := golden.ReadMetrics(expectedMetricsFile)
 	require.NoError(t, err, "Failed to read expected metrics from %s", expectedFileName)
-	
+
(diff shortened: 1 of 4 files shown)


functional_tests/functional/functional_test.go
```diff
@@ -1107,174 +1107,108 @@ func testAgentMetrics(t *testing.T) {
...
+		pmetrictest.ChangeResourceAttributeValue("container.image.name", containerImageShorten),
+		pmetrictest.ChangeResourceAttributeValue("host.name", replaceWithStar),
+		pmetrictest.ChangeResourceAttributeValue("service_instance_id", replaceWithStar),
+		pmetrictest.IgnoreScopeVersion(),
+		pmetrictest.IgnoreResourceMetricsOrder(),
+		pmetrictest.IgnoreMetricsOrder(),
+		pmetrictest.IgnoreScopeMetricsOrder(),
+		pmetrictest.IgnoreMetricDataPointsOrder(),
+		pmetrictest.IgnoreDatapointAttributesOrder(),
+		pmetrictest.IgnoreSubsequentDataPoints(metricNames...),
```
jinja2, line 1209: ended up adding this for all metric names due to the high instances of datapoint count mismatch. we can make the expected testdata more readable by now keeping samples of unique metric names only

PR #2083, opened 2025-09-25
[agent] Fix config when O11y metrics are disabled

crobert-1:

Repro:
- Update values.yaml to have:
```
splunkObservability:
  metricsEnabled: false
```
- Install/upgrade chart
Errors from agent pod logs:
```
Error: invalid configuration: service::pipelines::metrics/histograms: references receiver "receiver_creator" which is not configured
exporters::signalfx: invalid root_path: invalid root_path: stat /hostfs: no such file or directory
2025/09/24 23:17:27 main.go:94: application run finished with error: invalid configuration: service::pipelines::metrics/histograms: references receiver "receiver_creator" which is not configured
exporters::signalfx: invalid root_path: invalid root_path: stat /hostfs: no such file or directory
```

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -952,9 +952,11 @@ exporters:
     api_url: {{ include "splunk-otel-collector.o11yApiUrl" . }}
     {{- end }}
     access_token: ${SPLUNK_OBSERVABILITY_ACCESS_TOKEN}
+    {{- if eq (include "splunk-otel-collector.o11yMetricsEnabled" $) "true" }}
     sync_host_metadata: true
-    {{- if and (not .Values.isWindows) (eq (include "splunk-otel-collector.o11yMetricsEnabled" $) "true") }}
+    {{- if not .Values.isWindows }}
     root_path: /hostfs
+    {{- end }}
     {{- end }}
 
   # To send entities (applicable only if discovery mode is enabled)
--- a/examples/enable-trace-sampling/rendered_manifests/configmap-agent.yaml
+++ b/examples/enable-trace-sampling/rendered_manifests/configmap-agent.yaml
@@ -31,7 +31,6 @@ data:
         api_url: https://api.CHANGEME.signalfx.com
         correlation: null
         ingest_url: https://ingest.CHANGEME.signalfx.com
-        sync_host_metadata: true
       signalfx/histograms:
         access_token: ${SPLUNK_OBSERVABILITY_ACCESS_TOKEN}
         api_url: https://api.CHANGEME.signalfx.com
--- a/examples/enable-trace-sampling/rendered_manifests/daemonset.yaml
+++ b/examples/enable-trace-sampling/rendered_manifests/daemonset.yaml
@@ -31,7 +31,7 @@ spec:
         component: otel-collector-agent
         release: default
       annotations:
-        checksum/config: c6f121d82127473ebcde25781f37aca213dde57c2e449513387da2ca8f1f161c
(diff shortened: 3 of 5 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -951,7 +951,7 @@ exporters:
     {{- end }}
     access_token: ${SPLUNK_OBSERVABILITY_ACCESS_TOKEN}
     sync_host_metadata: true
-    {{- if not .Values.isWindows }}
+    {{- if and (not .Values.isWindows) (eq (include "splunk-otel-collector.o11yMetricsEnabled" $) "true") }}
```
crobert-1: `root_path` is only relevant when the `hostmetrics` receiver is configured and enabled ([reference](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/signalfxexporter/README.md)), which is not the case if metrics are disabled. There's no need to include `root_path` if metrics are disable.d

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -1216,7 +1216,7 @@ service:
         - resource
       exporters: [otlphttp/entities]
 
-    {{- if .Values.featureGates.useControlPlaneMetricsHistogramData }}
+    {{- if and .Values.featureGates.useControlPlaneMetricsHistogramData (eq (include "splunk-otel-collector.metricsEnabled" .) "true")}}
```
crobert-1: The `receiver_creator` receiver is only configured if metrics are enabled for platform or o11y cloud. If metrics are disabled this pipeline serves no purpose, regardless of the histogram feature flag.

.chloggen/fixo11ymetricsdisabled.yaml
```diff
@@ -0,0 +1,14 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix agent configuration when metrics are disabled for Splunk O11y Cloud
+# One or more tracking issues related to the change
+issues: []
```
crobert-1:
```suggestion
issues: [2083]
```

dmitryax (approved):

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
jinja2: Hmm, shouldn't we disable the sync_host_metadata config also then? We use the root_path to gather metadata for the host property updates. If we are not setting the root_path but sync is enabled, aren't we just using the wrong ( assuming the default is `/`) path then cause the sync is still enabled?

crobert-1: Thanks for pointing this out, fixed

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -952,10 +952,12 @@ exporters:
     api_url: {{ include "splunk-otel-collector.o11yApiUrl" . }}
     {{- end }}
     access_token: ${SPLUNK_OBSERVABILITY_ACCESS_TOKEN}
+    {{- if eq (include "splunk-otel-collector.o11yMetricsEnabled" $) "true" }}
```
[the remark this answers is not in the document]
crobert-1: The exporter is only defined if `splunk-otel-collector.splunkO11yEnabled` is true, so I believe we we should stick with `o11yMetricsEnabled` here over `metricsEnabled.

PR #2084, opened 2025-09-25
[chore] Add noDropLogs functional test

dloucasfx:
Added a bunch of k8s helper functions.
- `GetPodLogs`
- `GetPods`
- `CopyFileToPod`

And introduced a new functional test to validate the `noDropLogsPipeline` option works as expected.

### DIFF
--- a/functional_tests/internal/common.go
+++ b/functional_tests/internal/common.go
@@ -5,6 +5,7 @@ package internal
 
 import (
 	"context"
+	"io"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -27,6 +28,9 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
 )
 
 var DefaultNamespace = "default"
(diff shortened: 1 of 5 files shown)


functional_tests/internal/common.go
```diff
@@ -124,6 +128,78 @@ func MaybeUpdateExpectedLogsResults(t *testing.T, file string, logs *plog.Logs)
...
+			Stdin:     true,
+			Stdout:    true,
+			Stderr:    true,
+			TTY:       false,
+		}, scheme.ParameterCodec)
+
+	exec, err := remotecommand.NewSPDYExecutor(config, "POST", req.URL())
+	require.NoError(t, err, "failed to create SPDY executor")
+
+	err = exec.StreamWithContext(t.Context(), remotecommand.StreamOptions{
```
crobert-1, line 158: I disagree, we should use `t.Context` here. (Also, there are linters that complain if you use `context.Background`.)

dloucasfx: yeah, I disagree with a lot Copilot is saying :D human power :D

functional_tests/logs/no_drop_logs_test.go
```diff
@@ -0,0 +1,119 @@
...
+	// 2. Log sink is available, and all logs are exported successfully.
+	// This can be verified by matching the number of log records to the test log file line count
+	t.Run("NoDropLogs", func(t *testing.T) {
+		time.Sleep(10 * time.Second)
+		podLogs := internal.GetPodLogs(t, clientset, internal.DefaultNamespace, podName, containerName, 100)
+		require.Contains(t, podLogs, "Exporting failed. Rejecting data.", "expected log message not found in pod logs")
+
+		logsConsumer := internal.SetupHECLogsSink(t)
+
+		internal.WaitForLogs(t, 2, logsConsumer)
```
dmitryax, line 80: Why do we get 2 batches? How is this calculated?

dloucasfx: I am not 100% sure, but I got 2 batches when I run it locally and it matched when I ran it in CI.
I set the sending queue size to 3 , ie: 3 requests, so the filelog receiver must have been sending log in more than 3 requests, resulting of 2 batches.

dmitryax: It might be default batch size of the filelog receiver. We need to clearly document this number

dmitryax: Looks like the filelog receiver batch size is [100](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3651cff09653d78f72d31380afac73728712d26c/pkg/stanza/fileconsumer/internal/reader/factory.go#L31). It doesn't add up. We need to figure this out and document to avoid test flakiness

dloucasfx: > Looks like the filelog receiver batch size is [100](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3651cff09653d78f72d31380afac73728712d26c/pkg/stanza/fileconsumer/internal/reader/factory.go#L31). It doesn't add up.

Right, this actually makes sense; 
In the test I set the queue size to 3 ( 3 requests ) and as you have shared, the filelog receiver batch size is 100, in other words, the sender queue will be marked full after 3 requests of 100 logs each  ie: 300 logs are consumed with each exporter batch. So we need 2 exporter batch to go through the 600 log test file

jinja2: @dloucasfx do you mind including a comment with the reason for the 2 batches? It'll be helpful context for others who might have to interact with this test

PR #2102, opened 2025-10-03
Bind the OTLP HTTP port 4318 to the host network

dloucasfx:


.chloggen/OTLPPortWindows.yaml
```diff
@@ -0,0 +1,12 @@
...
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Bind the OTLP HTTP port 4318 to the host network
+# One or more tracking issues related to the change
+issues: [2102]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
jinja2:
```suggestion
subtext: |
  This change is effective only for Windows nodes, where the port binding was previously missing.
  For Linux nodes, the hostNetwork setting already ensures the port is bound, so no change in behavior should occur for default chart config.
```
I think making a clarification would be helpful

PR #2103, opened 2025-10-05
Enable additional attributes

tgtanyagoel:
change_type: enhancement
component: clusterReceiver
note: "Enable additional attributes (k8s.kubelet.version) for k8s cluster receiver"

### DIFF
--- a/.chloggen/attributes.yaml
+++ b/.chloggen/attributes.yaml
@@ -1,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: clusterReceiver
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Enable additional attributes (k8s.kubelet.version) for k8s cluster receiver"
+# One or more tracking issues related to the change
+issues: [2103]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/examples/add-filter-processor/rendered_manifests/configmap-cluster-receiver.yaml
+++ b/examples/add-filter-processor/rendered_manifests/configmap-cluster-receiver.yaml
@@ -133,6 +133,8 @@ data:
         resource_attributes:
           k8s.container.status.last_terminated_reason:
             enabled: true
+          k8s.kubelet.version:
+            enabled: true
       k8s_events:
         auth_type: serviceAccount
       prometheus/k8s_cluster_receiver:
--- a/examples/add-filter-processor/rendered_manifests/deployment-cluster-receiver.yaml
+++ b/examples/add-filter-processor/rendered_manifests/deployment-cluster-receiver.yaml
@@ -30,7 +30,7 @@ spec:
         component: otel-k8s-cluster-receiver
         release: default
       annotations:
-        checksum/config: 2e77337f2bf042ab3619a655c81fd96969615455aa65c8dc3b138cde8cb3d3e3
+        checksum/config: 6d5bb95325dca27ebcc02ca607dd7906e2bd47907b88fa64670641726a712231
     spec:
       serviceAccountName: default-splunk-otel-collector
       nodeSelector:
--- a/examples/add-kafkametrics-receiver/rendered_manifests/configmap-cluster-receiver.yaml
+++ b/examples/add-kafkametrics-receiver/rendered_manifests/configmap-cluster-receiver.yaml
@@ -133,6 +133,8 @@ data:
         resource_attributes:
           k8s.container.status.last_terminated_reason:
             enabled: true
+          k8s.kubelet.version:
+            enabled: true
       k8s_events:
         auth_type: serviceAccount
       prometheus/k8s_cluster_receiver:
--- a/examples/add-kafkametrics-receiver/rendered_manifests/deployment-cluster-receiver.yaml
+++ b/examples/add-kafkametrics-receiver/rendered_manifests/deployment-cluster-receiver.yaml
@@ -30,7 +30,7 @@ spec:
         component: otel-k8s-cluster-receiver
         release: default
       annotations:
-        checksum/config: 2e77337f2bf042ab3619a655c81fd96969615455aa65c8dc3b138cde8cb3d3e3
+        checksum/config: 6d5bb95325dca27ebcc02ca607dd7906e2bd47907b88fa64670641726a712231
     spec:
       serviceAccountName: default-splunk-otel-collector
       nodeSelector:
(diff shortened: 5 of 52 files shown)


.chloggen/functest.yaml
crobert-1, line 1: Test changes don't need a changelog entry. (Changelogs are meant to be things that impact users, or what users may care about.)

functional_tests/functional/functional_test.go
crobert-1, line 1: There are a couple sections of commented out code in this test, can we remove them? Is there any value in keeping them?

functional_tests/functional/functional_test.go
```diff
@@ -931,16 +931,55 @@ func testK8sClusterReceiverMetrics(t *testing.T) {
 	expectedMetrics, err := golden.ReadMetrics(expectedMetricsFile)
 	require.NoError(t, err)
 
-	metricNames := []string{"k8s.node.condition_ready", "k8s.namespace.phase", "k8s.pod.phase", "k8s.replicaset.desired", "k8s.replicaset.available", "k8s.daemonset.ready_nodes", "k8s.daemonset.misscheduled_nodes", "k8s.daemonset.desired_scheduled_nodes", "k8s.daemonset.current_scheduled_nodes", "k8s.container.ready", "k8s.container.memory_request", "k8s.container.memory_limit", "k8s.container.cpu_request", "k8s.container.cpu_limit", "k8s.deployment.desired", "k8s.deployment.available", "k8s.container.restarts", "k8s.container.cpu_request", "k8s.container.memory_request", "k8s.container.memory_limit"}
+	// Collect expected metric names
+	var metricNames []string
```
crobert-1: I like the idea of simply using the expected metrics here instead of a pre-generated list of metrics, but can you confirm that the generated list matches the original value of `metricNames`? Were any metrics left out of `metricNames` originally, and if so, why?

functional_tests/functional/functional_test.go
```diff
@@ -931,16 +931,55 @@ func testK8sClusterReceiverMetrics(t *testing.T) {
...
+
+	// metricNames := []string{"k8s.node.condition_ready", "k8s.namespace.phase", "k8s.pod.phase", "k8s.replicaset.desired", "k8s.replicaset.available", "k8s.daemonset.ready_nodes", "k8s.daemonset.misscheduled_nodes", "k8s.daemonset.desired_scheduled_nodes", "k8s.daemonset.current_scheduled_nodes", "k8s.container.ready", "k8s.container.memory_request", "k8s.container.memory_limit", "k8s.container.cpu_request", "k8s.container.cpu_limit", "k8s.deployment.desired", "k8s.deployment.available", "k8s.container.restarts", "k8s.container.cpu_request", "k8s.container.memory_request", "k8s.container.memory_limit"}
+
 	replaceWithStar := func(string) string { return "*" }
 
 	var selectedMetrics *pmetric.Metrics
 	for h := len(metricsConsumer.AllMetrics()) - 1; h >= 0; h-- {
 		m := metricsConsumer.AllMetrics()[h]
 
+		// Build a set for quick lookup
```
crobert-1: The goal here is to simply speed up the comparison process, is that correct? Did you notice a considerable decrease in how long the test takes with this change?

functional_tests/functional/testdata/expected_kind_values/expected_dotnet_traces.yaml
crobert-1, line 1: Do you happen to know why the testdata was changed for the dotnet/java/nodejs/python here? Since none of the test logic or functionality for these tests is changed I'm surprised the test data needs updated.

functional_tests/functional/testdata/expected_kind_values/expected_cluster_receiver.yaml
```diff
@@ -3897,6 +3897,9 @@ resourceMetrics:
                     - key: receiver
                       value:
                         stringValue: k8scluster
+                    - key: k8s.kubelet.version
```
crobert-1, line 3900: Do you happen to know why the `k8s.kubelet.version` attribute is only added to one data point in the test data? I would have expected it to be added to all of them.

.chloggen/functest.yaml
tgtanyagoel: Noted, done

functional_tests/functional/functional_test.go
tgtanyagoel: We haven't left out any in the original hardcoded list- but it seems to have a few metrics mentioned twice [k8s.container.cpu_request,k8s.container.memory_request,k8s.container.memory_limit] causing a count mismatch

The dynamic generated list has the same metrics as printed here: https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/18470034353/job/52621361916#step:7:1065

functional_tests/functional/testdata/expected_kind_values/expected_dotnet_traces.yaml
tgtanyagoel: My assumption is there were changes made but these files were not updated over time, so when I pulled the generated artifact and updated it, it brought all the changed data to sync. This can be reversed from this PR if that helps. The only testdata file relevant here is [this](https://github.com/signalfx/splunk-otel-collector-chart/pull/2103/files/513ebc2021e163b895a0b0abe11944962284d9bd#diff-c89952ffd44d3129ad8e4428b6814bbd641ab68424d75e995b1be0fd2127bd0d)

functional_tests/functional/testdata/expected_kind_values/expected_cluster_receiver.yaml
tgtanyagoel: I noticed this as well, but realized that attributes are inconsistent across metrics in the current testdata as well. `k8s.node.uid` only exists on one metric as well
Is there a better way to validate this?

functional_tests/functional/functional_test.go
tgtanyagoel: Yes, and to declutter the debug logs. The delta in time ended up being just a few (10-15) seconds

functional_tests/functional/functional_test.go
tgtanyagoel: cleaned them.

functional_tests/functional/functional_test.go
dmitryax: This is not related to the PR, right? This optimization saves time, but also complicates the test which is already pretty hard to maintain. I'd say let's avoid adding it. 

In general, any PR should have changes only related to its goal without unrelated patches. Even if they are net positive, they should be made separately.

functional_tests/functional/functional_test.go
dmitryax: Same comment here. Simply removing the duplicates, can be a nice separate PR, but I don't think this is required for this PR's tests to pass.

I believe this test was failing on main, and it was fixed recently, so we should be good now

functional_tests/functional/functional_test.go
tgtanyagoel: Removed this optimization

functional_tests/functional/functional_test.go
tgtanyagoel: reverted the changes

dmitryax (approved):

PR #2106, opened 2025-10-07
Changes in collector configuration in agent mode to support Secure Application features in Splunk Observability Cloud

tarora-splunk:
**Description:**
This PR includes changes to the collector chart in agent mode to generate configuration that is needed to receive Secure Application data and send to the Splunk Observability Cloud.

**Testing:**
Verified the rendered configurations by doing "make render" locally


tarora-splunk:

tarora-splunk:
@dmitryax please review this PR.

.chloggen/featsecureappagentmode.yaml
```diff
@@ -0,0 +1,8 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Configuration changes to support Secure Application features in Splunk Observability Cloud
+# One or more tracking issues related to the change
+issues: []
```
dmitryax: PR number can be used as well

```suggestion
issues: [2106]
```

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -109,6 +109,13 @@ Whether logs enabled for any destination.
 {{- include "splunk-otel-collector.platformLogsEnabled" . }}
 {{- end -}}
 
+{{/*
+Whether SecureApp is enabled (boolean flag).
+*/}}
+{{- define "splunk-otel-collector.secureAppEnabled" -}}
+{{- .Values.splunkObservability.secureAppEnabled -}}
```
dmitryax: The helper is redundant, let's just use `.Values.splunkObservability.secureAppEnabled`

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -173,6 +173,9 @@ splunkObservability:
   # If you don't use AlwaysOn Profiling for Splunk APM, you can disable it.
   profilingEnabled: false
 
+  # Enables Secure Application features
+  secureAppEnabled: true
```
dmitryax: Given that it's just a pass-through and doesn't introduce any significant overhead, do we actually need this to be configurable?

helm-charts/splunk-otel-collector/templates/_helpers.tpl
tarora-splunk: done

helm-charts/splunk-otel-collector/values.yaml
tarora-splunk: Given that we are still in alpha, we would like to have a knob to turn it off. Shall we set the value to false by default?

dmitryax: If it's something that we eventually want to have enabled by default regardless, this should go to the `featureGates` section. If we want to keep this knob forever, this one is good.

Maybe as the initial release this one can be disabled by default to avoid accidental issues. Once we have the integration tests, we can enable it by default

tarora-splunk: Makes sense, changed to false. Will enable it when adding integration tests.

PR #2092, opened 2025-10-01
[chore][istio] Add integration test for OTLP traces

crobert-1:
**Description:** <Describe what has changed.>

This PR introduces an integration test for OTLP traces coming from Istio. It uses an existing deployment of Istio on a Kind cluster, with some extra resource configs to ensure tracing is enabled and being sent to expected destinations.

PR #2108, opened 2025-10-07
[chart] Move istio sidecar injection annotation to label

crobert-1:

The [annotation](https://istio.io/latest/docs/reference/config/annotations/#SidecarInject) `sidecar.istio.io/inject` has been deprecated in favor of the [label](https://istio.io/latest/docs/reference/config/labels/#SidecarInject) with the same name.

_Also fixes some typos in comments._

### DIFF
--- a/examples/autodetect-istio/rendered_manifests/daemonset.yaml
+++ b/examples/autodetect-istio/rendered_manifests/daemonset.yaml
@@ -30,10 +30,10 @@ spec:
         app: splunk-otel-collector
         component: otel-collector-agent
         release: default
+        sidecar.istio.io/inject: "false"
       annotations:
         checksum/config: d2d16cc55113029807342c5b68b2d04773bafd92ce2f3c0c0df99eb6638c3dea
         kubectl.kubernetes.io/default-container: otel-collector
-        sidecar.istio.io/inject: "false"
     spec:
       hostNetwork: true
       dnsPolicy: ClusterFirstWithHostNet
--- a/.chloggen/movedeprecatedistioannotations.yaml
+++ b/.chloggen/movedeprecatedistioannotations.yaml
@@ -1,0 +1,13 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: chart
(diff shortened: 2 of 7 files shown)


examples/autodetect-istio/rendered_manifests/daemonset.yaml
```diff
@@ -30,10 +30,10 @@ spec:
         app: splunk-otel-collector
         component: otel-collector-agent
         release: default
+        sidecar.istio.io/inject: "false"
```
dmitryax, line 33: Which istio version introduced this support? If it's recent, we might want to keep both for now. Otherwise, we need to update the docs/comments for supported istio versions, if we have any.

functional_tests/istio/testdata/expected_istiod.yaml
```diff
@@ -10975,50 +10975,6 @@ resourceMetrics:
                         stringValue: http
                   timeUnixNano: "1000000"
               isMonotonic: true
-          - name: sidecar_injection_skip_total
```
crobert-1, line 10978: This [metric was added in a recent PR](https://github.com/signalfx/splunk-otel-collector-chart/pull/2092#discussion_r2400232285), but it looks like since the functionality is now a label instead of annotation this metric is no longer scraped/sent.

examples/autodetect-istio/rendered_manifests/daemonset.yaml
crobert-1: From documentation, the [annotation has been deprecated](https://istio.io/v1.15/docs/reference/config/annotations) since [Istio v1.15](https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases), which has been EOL since April 2023.

The label with the same name has been in place since [Istio v1.16](https://istio.io/v1.16/docs/reference/config/labels/), and labeled `beta` stability since that release.

dmitryax: Sounds good then 👍

crobert-1:
> Curios why do they migrate it and if this is a general trend in k8s 🤔

I found a [stackoverflow comment](https://stackoverflow.com/a/78862392) explaining a bit of the difference in functionality.

At a higher level, I had to look up the difference between annotations and labels when I saw it was deprecated, [this section](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#attaching-metadata-to-objects) seemed relevant. My guess is that since Istio is looking for a specific value and will act/not act depending on it, a label more closely represents the intended functionality than an annotation. 🤷

PR #2288, opened 2026-03-12
[agent] Drop server.address and server.port dimensions from Prometheus metrics

crobert-1:

To avoid hitting dimension limit. Add Istio autodetect example with histogram feature gate disabled to show config works properly for both feature gate enabled and disabled.

### DIFF
--- a/.chloggen/fixistioattributes.yaml
+++ b/.chloggen/fixistioattributes.yaml
@@ -3,13 +3,14 @@ change_type: breaking
 # The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
 component: agent
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Drop `server.address` and `server.port` dimensions from Istio metrics
+note: Drop `server.address` and `server.port` dimensions from Prometheus metrics
 # One or more tracking issues related to the change
 issues: [2288]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
-  Istio metrics can run into dimension limits due to how many Istio-specific dimensions there
-  are. To avoid data points being rejected due to exceeding the dimension limit,
-  these dimensions will now be dropped from Istio metrics.
+  Prometheus metrics (especially coming from Istio) can run into dimension limits.
+  To avoid data points being rejected due to exceeding the dimension limit, these
+  dimensions will now be dropped when either `autodetect.prometheus` or `autodetect.istio`
+  is enabled.
--- a/.chloggen/fixistioattributes.yaml
+++ b/.chloggen/fixistioattributes.yaml
@@ -1,0 +1,15 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Drop `server.address` and `server.port` dimensions from Istio metrics
+# One or more tracking issues related to the change
+issues: []
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Istio metrics can run into dimension limits due to how many Istio-specific dimensions there
+  are. To avoid data points being rejected due to exceeding the dimension limit,
+  these dimensions will now be dropped from Istio metrics.
--- a/.chloggen/fixistioattributes.yaml
+++ b/.chloggen/fixistioattributes.yaml
@@ -5,7 +5,7 @@ component: agent
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: Drop `server.address` and `server.port` dimensions from Istio metrics
 # One or more tracking issues related to the change
-issues: []
+issues: [2288]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
--- a/.chloggen/fixistioattributes.yaml
+++ b/.chloggen/fixistioattributes.yaml
@@ -13,4 +13,5 @@ subtext: |
   Prometheus metrics (especially coming from Istio) can run into dimension limits.
   To avoid data points being rejected due to exceeding the dimension limit, these
   dimensions will now be dropped when either `autodetect.prometheus` or `autodetect.istio`
-  is enabled.
+  is enabled. Users may reference the data point attribute `service.instance.id` as an
+  alternate source of the server address and port.
--- a/functional_tests/istio/testdata/istio_values.yaml.tmpl
+++ b/functional_tests/istio/testdata/istio_values.yaml.tmpl
@@ -8,6 +8,7 @@ autodetect:
   istio: true
 
 agent:
+
   controlPlaneMetrics:
     apiserver:
       enabled: false
@@ -28,3 +29,6 @@ clusterReceiver:
 
 environment: dev
 clusterName: dev-operator
+
+featureGate:
+  useControlPlaneMetricsHistogramData: false
--- a/functional_tests/istio/testdata/istio_values.yaml.tmpl
+++ b/functional_tests/istio/testdata/istio_values.yaml.tmpl
@@ -8,13 +8,6 @@ autodetect:
   istio: true
 
 agent:
-  config:
-    service:
-      pipelines:
-        metrics:
-          receivers:
-            - receiver_creator
-
   controlPlaneMetrics:
     apiserver:
       enabled: false
--- a/functional_tests/istio/testdata/istio_values.yaml.tmpl
+++ b/functional_tests/istio/testdata/istio_values.yaml.tmpl
@@ -8,6 +8,13 @@ autodetect:
   istio: true
 
 agent:
+  config:
+    service:
+      pipelines:
+        metrics:
+          receivers:
+            - receiver_creator
+
   controlPlaneMetrics:
     apiserver:
       enabled: false
@@ -28,6 +35,3 @@ clusterReceiver:
 
 environment: dev
 clusterName: dev-operator
-
-featureGates:
-  useControlPlaneMetricsHistogramData: false
--- a/functional_tests/istio/testdata/istio_values.yaml.tmpl
+++ b/functional_tests/istio/testdata/istio_values.yaml.tmpl
@@ -8,7 +8,6 @@ autodetect:
   istio: true
 
 agent:
-
   controlPlaneMetrics:
     apiserver:
       enabled: false
@@ -30,5 +29,5 @@ clusterReceiver:
 environment: dev
 clusterName: dev-operator
 
-featureGate:
+featureGates:
   useControlPlaneMetricsHistogramData: false
--- a/helm-charts/splunk-otel-collector/templates/_helpers.tpl
+++ b/helm-charts/splunk-otel-collector/templates/_helpers.tpl
@@ -509,17 +509,3 @@ Usage:
 {{- end -}}
 {{- end -}}
 {{- end -}}
-
-{{/*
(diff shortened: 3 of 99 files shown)


.chloggen/fixistioattributes.yaml
```diff
@@ -0,0 +1,15 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Drop `server.address` and `server.port` dimensions from Istio metrics
```
dmitryax: It will be applied to all prometheus metrics, which I think is ok, but we need to be explicit about it

dmitryax: Or if we can apply metrics filters in the transform processor to isolate it to istio only, that would be better. Or make a dedicated pipeline...

crobert-1: Unfortunately the other resource attributes (eg `service.name`) don't always provide a definitive way to determine when metrics are coming from Istio vs. other sources. I think the only other option is a dedicated pipeline.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -1372,6 +1401,10 @@ service:
         - resource/add_agent_k8s
         - resourcedetection
         - resource
+        {{- if or .Values.autodetect.prometheus .Values.autodetect.istio }}
```
crobert-1, line 1404: This is technically a bug_fix/enhancement as well, since `attributes/istio` was not being applied in this case before. Check diff in the `examples/autodetect-istio` agent configmap to see the addition of this processor to the histogram pipeline when it wasn't there previously.

functional_tests/istio/testdata/istio_values.yaml.tmpl
crobert-1, line 1: The existing config adding `receiver_creator` to the metrics pipeline was a workaround to avoid a bug in the chart. `useControlPlaneMetricsHistogramData` is being disabled here instead to allow a closer to valid test than the workaround. Once the underlying bug is fixed the feature gate will be able to be enabled again (as it is by default).

.chloggen/fixistioattributes.yaml
crobert-1: Moved to using a dedicated pipeline with a `forward` connector to the `metrics` pipeline.

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -509,3 +509,17 @@ Usage:
...
 {{- end -}}
+
+{{/*
+Returns true if a separate pipeline for autodetected Prometheus or Istio metrics is required
+TODO: Feature gate reference needs updated after bug with histograms pipeline is resolved.
+The reference here is to keep existing behavior consistent.
+*/}}
+{{- define "splunk-otel-collector.useSeparatePrometheusMetricsPipeline" -}}
+{{- and
+  (not .Values.featureGates.useControlPlaneMetricsHistogramData)
```
dmitryax, line 520: Why do we need to differentiate the behavior because and don't drop `server.address` and `server.port` for native histograms?

crobert-1: The inclusion of the feature gate `useControlPlaneMetricsHistogramData` variable is very misleading. The goal of this separate pipeline being introduced is to only drop `server.address` and `server.port` when they're coming from the `receiver_creator`. The `receiver_creator` receiver will be in the `metrics` pipeline when `useControlPlaneMetricsHistogramData` is disabled, and it will be in the `metrics/histograms` pipeline when the feature gate is enabled. (This functionality of switching between pipelines has resulted in a few bugs that will be resolved in https://github.com/signalfx/splunk-otel-collector-chart/pull/2308, happy to include more information here if it'd be helpful).

When `receiver_creator` is in the `metrics/histogram` pipeline we don't need the separate pipeline. This is because `receiver_creator` is the only receiver in the `metrics/histogram` pipeline, so we can just enable the processors dropping attributes when autodetect is enabled in the pipeline itself.

dmitryax: I see. This seems very complicated and error-prone. What if we just move the `receiver_creator` to a new pipeline that drops `server.address` and `server.port` regardless and uses `signalfx` or `signalfx/histograms` as an exporter conditionally?
```
      receivers:
       - receiver_creator
      processors:
        - memory_limiter
        - batch
        - resource/add_agent_k8s
        - resourcedetection
        - resource
        - attributes/istio
        - transform/istio
      exporters:
        - signalfx{{ if .Values.featureGates.useControlPlaneMetricsHistogramData}}/histograms{{ end}}
```

@crobert-1 @jinja2 WDYT?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -976,6 +976,15 @@ processors:
         key: source_canonical_revision
       - action: delete
         key: destination_canonical_revision
+
+  # This processor is used to remove excessive attributes from Istio metrics to avoid running into the dimensions limit.
+  # These attributes are resource attributes coming from Prometheus scraping, which are eventually converted into
+  # data point attributes in the SignalFx exporter, counting against the dimension limit.
+  transform/istio:
```
dmitryax: Should this be?
```suggestion
  transform/drop_server_attrs:
```

helm-charts/splunk-otel-collector/templates/_helpers.tpl
crobert-1: It would end up being more complicated than that. The example you've given here would need to account for Splunk Platform metrics as well gateway enabled. (These are both bugs that are hit by the existing `metrics/histograms` pipeline that would be resolved by https://github.com/signalfx/splunk-otel-collector-chart/pull/2308).

Theoretically this could work, but it still seems pretty complicated and error-prone, especially since it'll practically just be duplicating the `metrics` pipeline until #2308 is ready, at which point we'd just be deleting this new pipeline.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
crobert-1: Good call, updated

helm-charts/splunk-otel-collector/templates/_helpers.tpl
dmitryax: > it'll practically just be duplicating the metrics pipeline until https://github.com/signalfx/splunk-otel-collector-chart/pull/2308 is ready, at which point we'd just be deleting this new pipeline.

Right, I suggest we move the `receiver_creator` into a new pipeline, but why would we need to delete later? Don't we still want `attributes/istio` and `transform/drop_server_attrs` processors exclusively applied to `receiver_creator`? A separate pipeline is probably cleaner than using `forward` connector

crobert-1: I guess if we did a duplicate pipeline we could pull out the processor and exporter logic into its own template helper that's just reused across pipelines. One declaration would minimize potential to drift.

That still leaves us with the histogram bug that skips the gateway, but since that's an existing bug it's not made worse by this change. I'd probably make the config consistent with existing behavior and then fix the actual bugs.

A couple other alternatives:
1. Modify the transform processor to only drop these attributes from metrics that match the istio naming scheme, metric names that start with `envoy_.*`, `istio_.*`, and `pilot_.*`.
2. Modify the Istio prometheus scraper to drop the server attributes at the time of scraping.

These alternatives would only really work if the Istio prometheus scraper is the only one that we're trying to filter these attributes out of. If we want to drop them from all `receiver_creator` Prometheus endpoints we'd be better off sticking with the transform processor.

dmitryax: > Modify the Istio prometheus scraper to drop the server attributes at the time of scraping.

This doesn't work because the server attributes are added by the receiver, they aren't scraped

> If we want to drop them from all receiver_creator Prometheus endpoints we'd be better off sticking with the transform processor.

I think this is the cleanest solution. I don't see any value in those server attributes. We should be getting enough of identifying attributes already

@jinja2 curious to hear your thought on this

crobert-1: > This doesn't work because the server attributes are added by the receiver, they aren't scraped

I wasn't aware, thanks for clarifying! Yeah, I guess it's a question of forward connector vs. "duplicate" pipelines then. I don't have too strong of an opinion here. I'll defer to you and @jinja2 👍

jinja2: Could we use an OTTL scope-based guard in transform?

Tmk, our main issue is that `server.address` and `server.port` are general OTel semconv attributes, not Prometheus-specific. Other receiver could legitimately have them and we don't want to drop in that case. Assuming the receiver always sets its [instrumentation scope name](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/metadata/generated_status.go#L11), we could scope the deletion (did not test the ottl):


```
transform/drop_server_attrs:
  error_mode: ignore
  metric_statements:
    - context: scope
      statements:
        - delete_key(resource.attributes, "server.address") where instrumentation_scope.name == "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver"
        - delete_key(resource.attributes, "server.port") where instrumentation_scope.name == "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver"
```        

This should be safe to add directly to metrics and metrics/histograms pipelines, especially given the info isn't actually lost since `service.instance.id` already has `host:port` ([instance is split to get the address/port](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/prom_to_otlp.go#L46-L61)).


Additional questions which might make this not workable -
1) Does this work as expected with receiver_creator-wrapped sub-receivers, 2) any performance implications with transform iterating over  `context: scope` (resources x scopes vs resources). I think prom receiver emits 1 scope per resource, so it might be equivalent to `context: resource` in practice

crobert-1: I was able to test and it does work as expected, `server.port` and `server.address` are only being dropped from the prometheus receiver metrics.

Regarding concerns:
1. This works properly for the receiver_creator-wrapped receivers. I confirmed on Istio related metrics that `server.address` and `server.port` are being dropped.
2. I ran pprof against the agent pod for CPU and memory profiles and neither gave any indication of a performance impact due to adding this processor. Also from my testing, only one scope was used for the resource. Since context is automatically inferred and [scope](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/contexts/ottlscope/README.md) doesn't have many more variables than [resource](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/contexts/ottlresource/README.md), I don't believe this will have a noteworthy performance impact.

This approach works for me, do you have any specific concerns @dmitryax?

dmitryax: I like this approach. It should make the PR much simpler 👍

functional_tests/istio/testdata/istio_values.yaml.tmpl
crobert-1: I've decided to leave as-is since the rest of this PR has been simplified.

helm-charts/splunk-otel-collector/templates/_helpers.tpl
crobert-1: _I've implemented the approach suggested by @jinja2_

.chloggen/fixistioattributes.yaml
```diff
@@ -0,0 +1,16 @@
...
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Drop `server.address` and `server.port` dimensions from Prometheus metrics
+# One or more tracking issues related to the change
+issues: [2288]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Prometheus metrics (especially coming from Istio) can run into dimension limits.
```
jinja2, line 13: Can we mention users can get this info from the service.instance.id attr instead?

jinja2 (approved):

dmitryax (approved):

PR #2369, opened 2026-04-11
Remove resourcedetector from cluster receiver event/object

jinja2:
**Description:**

The logs (k8s_events), logs/objects, and logs/events pipelines in the cluster receiver were using the full resourcedetection processor, which adds host-level attributes (host.id, host.name, cloud.provider, etc.). These pipelines carry cluster-scoped data where host identity is the cluster receiver pod, not the source of the event.

Align with the metrics pipeline which correctly uses only `resourcedetection/k8s_cluster_name` for auto-detect distributions.

### DIFF
--- a/.chloggen/fix-cluster-receiver-event-resourcedetection.yaml
+++ b/.chloggen/fix-cluster-receiver-event-resourcedetection.yaml
@@ -5,7 +5,7 @@ component: clusterReceiver
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: Remove resourcedetection processor from cluster-scoped event and object pipelines
 # One or more tracking issues related to the change
-issues: []
+issues: [2369]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
--- a/.chloggen/fix-cluster-receiver-event-resourcedetection.yaml
+++ b/.chloggen/fix-cluster-receiver-event-resourcedetection.yaml
@@ -1,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: clusterReceiver
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove resourcedetection processor from cluster-scoped event and object pipelines
+# One or more tracking issues related to the change
+issues: []
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The cluster receiver's `logs` (k8s_events), `logs/objects`, and `logs/events` pipelines
+  incorrectly used the full `resourcedetection` processor, which added host and
+  cloud-provider attributes (host.id, host.name, cloud.provider, cloud.platform,
+  cloud.account.id, cloud.region, etc.) to cluster-scoped data. These attributes reflected
+  the cluster receiver pod rather than the source of the event or object, making them
+  misleading.
--- a/.chloggen/fix-cluster-receiver-event-resourcedetection.yaml
+++ b/.chloggen/fix-cluster-receiver-event-resourcedetection.yaml
@@ -11,8 +11,8 @@ issues: [2369]
 # Use pipe (|) for multiline entries.
 subtext: |
   The cluster receiver's `logs` (k8s_events), `logs/objects`, and `logs/events` pipelines
-  incorrectly used the full `resourcedetection` processor, which added host and
+  previously used the `resourcedetection` processor, which added host and
   cloud-provider attributes (host.id, host.name, cloud.provider, cloud.platform,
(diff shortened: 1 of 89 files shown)


helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -367,7 +367,7 @@ Whether the clusterName configuration option is optional
 Whether the helm chart should detect the cluster name automatically
 */}}
 {{- define "splunk-otel-collector.autoDetectClusterName" -}}
-{{- and (include "splunk-otel-collector.clusterNameOptional" .) (not .Values.clusterName) }}
+{{- and (eq (include "splunk-otel-collector.clusterNameOptional" .) "true") (not .Values.clusterName) }}
```
jinja2, line 370: Not directly related to the change but this is a no-op fix. The problem here was how Go templates evaluate truthiness. `include` returns a string, and in Go templates any non-empty string is truthy. So for distributions like AKS where clusterNameOptional returns "false", the old check would treat it as truthy because "false" is a non-empty string.

In practice, i think we were probably not hitting the issue since we have JSON schema enforcing clusterName for distributions where clusterNameOptional is false.

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -418,7 +418,9 @@ service:
         - memory_limiter
         - batch
         - attributes/drop_event_attrs
-        - resourcedetection
+        {{- if eq (include "splunk-otel-collector.autoDetectClusterName" .) "true" }}
+        - resourcedetection/k8s_cluster_name
+        {{- end }}
```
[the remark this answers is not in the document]
jinja2: Fixed here https://github.com/signalfx/splunk-otel-collector-chart/pull/2369/changes#r3067565873

jinja2:
We need to decide on the appropriate `host` field for logs in Splunk platform

jinja2:
### Splunk HEC metadata for affected pipelines

| Pipeline | `source` | `sourcetype` | `index` | `host` (before) | `host` (after) |
|---|---|---|---|---|---|
| `logs` (k8s_events) | `kubernetes` | `kube:events` | `main` (default) | cluster receiver pod name or cloud-provider-specific hostname | node name (`K8S_NODE_NAME`) |
| `logs/objects` | `kubernetes` | `kube:object:<resource>` (e.g. `kube:object:services`) | `main` (default) | cluster receiver pod name or cloud-provider-specific hostname | node name (`K8S_NODE_NAME`) |
| `logs/events` | n/a (signalfx only) | n/a | n/a | n/a | n/a |

The Splunk HEC `host` field maps to `host.name` on the resource ([docs](https://docs.splunk.com/Documentation/Splunk/latest/Data/Aboutdefaultfields)). Previously it was set by the full `resourcedetection` processor - the pod name on vanilla Kubernetes, or a cloud-provider-specific hostname (e.g. EC2 private DNS) on managed clusters. After this change, a dedicated `resource/add_cluster_host` processor sets it to the node name via the downward API. Cluster-scoped data has no true originating host; the host running the cluster receiver is the closest physical device.

**Note:** The `logs` and `logs/objects` pipelines can also export to `otlp_http/o11y_events` (the v3/event endpoint) when enabled. The Splunk-specific metadata (`host.name`, `com.splunk.source`, `com.splunk.sourcetype`, `com.splunk.index`) will also be present on data sent to the v3/event endpoint.

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -247,6 +247,19 @@ processors:
...
+  # defines as "the network host from which the event originated"
+  # (https://docs.splunk.com/Documentation/Splunk/latest/Data/Aboutdefaultfields).
+  # Cluster-scoped data has no originating host; use the node running
+  # the cluster receiver as the closest physical device.
+  resource/add_cluster_host:
+    attributes:
+      - action: insert
+        key: host.name
+        value: "${K8S_NODE_NAME}"
+  {{- end }}
```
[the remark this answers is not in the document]
jinja2: it is unconditionally defined now

.chloggen/fix-cluster-receiver-event-resourcedetection.yaml
```diff
@@ -0,0 +1,18 @@
...
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: clusterReceiver
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove resourcedetection processor from cluster-scoped event and object pipelines
+# One or more tracking issues related to the change
+issues: [2369]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
```
crobert-1, line 12: Nit: I don't know if it's really necessary, maybe everyone reading this will understand, but it might be helpful to add one sentence explaining that these host-level attributes should be set to the values of the source of the logs rather than the single cluster receiver pod that's doing the scraping.

The subtext explains what was happening and what's changed, but it doesn't explain why it was incorrect.

crobert-1 (approved):

dmitryax (approved):

.chloggen/fix-cluster-receiver-event-resourcedetection.yaml
jinja2: updated

PR #2370, opened 2026-04-13
simplify filelog operators with container operator

omrozowicz-splunk:
**Description:** Replaces the manual `router` + `regex_parser/json_parser` + recombine chain for Docker/CRI-O/containerd log parsing with the single container operator from opentelemetry-collector-contrib. This operator handles format auto-detection, timestamp parsing, multiline P/F recombination, and k8s metadata extraction from file path (including Windows) natively, significantly reducing config complexity.

`container` operator doesn't combine docker formatted events, this is why recombine operator was added:
```
      - type: recombine
        id: docker-recombine
        if: attributes.logtag == nil
        combine_field: body
        source_identifier: attributes["log.file.path"]
        is_last_entry: body endsWith "\n"
        combine_with: ""
        max_log_size: {{ $.Values.logsCollection.containers.maxRecombineLogSize }}
```

**Link to Splunk idea:** N/A

**Testing:** manual tests/integration tests

**Documentation:** N/A


omrozowicz-splunk:
Tested on `containerd` microk8s, resource attributes look the same (top current version, bottom version with this PR'a build):
image
image

omrozowicz-splunk:
Also tested `container` operator on Windows, but I tested it directly with 3 files formatted like cri-o, containerd and docker. We don't need different parsing for Windows anymore:
```
Resource attributes:
     -> k8s.pod.uid: Str(49cc7c1fd3702c40b2686ea7486091d6)
     -> k8s.container.name: Str(nginx)
     -> com.splunk.sourcetype: Str(kube:container:nginx)
     -> com.splunk.source: Str(C:\var\log\pods\default_my-nginx-pod_49cc7c1fd3702c40b2686ea7486091d6\nginx\1.log)
     -> k8s.namespace.name: Str(default)
     -> k8s.pod.name: Str(my-nginx-pod)
     -> k8s.container.restart_count: Str(1)
ScopeLogs #0
ScopeLogs SchemaURL:
InstrumentationScope
LogRecord #0
ObservedTimestamp: 2026-04-13 13:27:58.3651045 +0000 UTC
Timestamp: 2024-04-13 12:59:37.505201169 +0000 UTC
SeverityText:
SeverityNumber: Unspecified(0)
Body: Str(ERROR: something went wrong)
Attributes:
     -> log.iostream: Str(stderr)
     -> logtag: Str(F)
```

config:
```
receivers:
  filelog:
    include:
      - C:\var\log\pods\*\*\*.log
    exclude:
      - C:\var\log\pods\*\*\*.log*.gz
      - C:\var\log\pods\*\*\*.log*.tmp
    include_file_path: true
    include_file_name: false
    start_at: beginning
    encoding: utf-8
    fingerprint_size: 1kb
    max_log_size: 1MiB
    poll_interval: 200ms
    operators:
      - type: container
        id: container-parser
        add_metadata_from_filepath: true
        max_log_size: 1048576
      - type: add
        id: handle_empty_log
        if: body == nil
        field: body
        value: ""
      - type: add
        field: resource["com.splunk.sourcetype"]
        value: EXPR("kube:container:"+resource["k8s.container.name"])
      - type: move
        from: attributes["log.file.path"]
        to: resource["com.splunk.source"]
```

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -753,14 +671,12 @@ receivers:
...
       {{- range $.Values.logsCollection.containers.multilineConfigs }}
       - type: recombine
         id: {{ include "splunk-otel-collector.newlineKey" . | quote}}
-        output: clean-up-log-record
         source_identifier: resource["com.splunk.source"]
-        combine_field: attributes.log
-        is_first_entry: '(attributes.log) matches {{ .firstEntryRegex | quote }}'
+        combine_field: body
+        is_first_entry: 'body matches {{ .firstEntryRegex | quote }}'
         max_log_size: {{ $.Values.logsCollection.containers.maxRecombineLogSize }}
```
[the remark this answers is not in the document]
omrozowicz-splunk: maybe let's introduce `noop` operator instead?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -753,14 +671,12 @@ receivers:
           - output: {{ include "splunk-otel-collector.newlineKey" . | quote }}
             expr: {{ include "splunk-otel-collector.newlineExpr" . | quote }}
         {{- end }}
```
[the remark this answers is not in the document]
jinja2: I think this is a legit bug we need to fix. We can use the `noop` operator as the convergence target for this too.

jinja2:
Please update [this comment in values.yaml](https://github.com/signalfx/splunk-otel-collector-chart/blob/splunk-otel-collector-0.148.0/helm-charts/splunk-otel-collector/values.yaml#L592) which states `cri-o` is a valid containerRuntime values. Looking at the [stanza code](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/operator/parser/container/parser.go#L28), it should be `crio`. We could also add an explicit schema check for this config (not sure whats enforced in schema at present).

jinja2:
There's no functional test exercising multiline today. Since we already have patterns for this kind of test (e.g., no_drop_logs, log_attributes_test, etc.), could you add a multiline test following the same approach? I would suggest adding the test in a separate PR which is merged before this PR, so we can verify that switching from the previous method to new does not introduce any breaking change.

We can set up the test as follows:

- A static log file checked into testdata/ containing a mix of single-line logs and realistic multiline entries. For e.g, `Java stack traces (at com.example...)`, `Python tracebacks (Traceback (most recent call last):)`, `Go panics (goroutine 1 [running]:)`. Sprinkle in some single-line entries so we can verify that we are not dropping these.
- A busybox deployment that mounts the file via a ConfigMap and does cat /testdata/input.log && sleep 30 in a loop.
- A test values template that enables multilineConfigs targeting that namespace/pod/container with timestamp-based (?) firstEntryRegex.
- Assertions on the HEC log sink that:
    Each multiline stack trace arrives as a single log record with the full trace in body
    Single-line entries pass through as individual records (this also acts as a regression test for the missing default route)

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -812,7 +812,8 @@
               "type": "boolean"
             },
             "containerRuntime": {
-              "type": "string"
+              "type": "string",
+              "enum": ["", "docker", "containerd", "crio"]
             },
```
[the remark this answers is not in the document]
omrozowicz-splunk: we could use something like `format: {{ .Values.logsCollection.containers.containerRuntime | replace "cri-o" "crio" }} ` but I think it is better to change changelog entry to `breaking` and restrict values.json.schema so helm upgrade fails when we set `containerRuntime` to be `crio-o`. If we use this normalisation the user never knows the value changed and might have stale config sitting in their values file indefinitely.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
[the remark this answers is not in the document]
omrozowicz-splunk: Updated for both

omrozowicz-splunk:
> There's no functional test exercising multiline today. Since we already have patterns for this kind of test (e.g., no_drop_logs, log_attributes_test, etc.), could you add a multiline test following the same approach? I would suggest adding the test in a separate PR which is merged before this PR, so we can verify that switching from the previous method to new does not introduce any breaking change.
> 
> We can set up the test as follows:
> 
> * A static log file checked into testdata/ containing a mix of single-line logs and realistic multiline entries. For e.g, `Java stack traces (at com.example...)`, `Python tracebacks (Traceback (most recent call last):)`, `Go panics (goroutine 1 [running]:)`. Sprinkle in some single-line entries so we can verify that we are not dropping these.
> * A busybox deployment that mounts the file via a ConfigMap and does cat /testdata/input.log && sleep 30 in a loop.
> * A test values template that enables multilineConfigs targeting that namespace/pod/container with timestamp-based (?) firstEntryRegex.
> * Assertions on the HEC log sink that:
>   Each multiline stack trace arrives as a single log record with the full trace in body
>   Single-line entries pass through as individual records (this also acts as a regression test for the missing default route)

Opened https://github.com/signalfx/splunk-otel-collector-chart/pull/2376

omrozowicz-splunk:
> Please update [this comment in values.yaml](https://github.com/signalfx/splunk-otel-collector-chart/blob/splunk-otel-collector-0.148.0/helm-charts/splunk-otel-collector/values.yaml#L592) which states `cri-o` is a valid containerRuntime values. Looking at the [stanza code](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/operator/parser/container/parser.go#L28), it should be `crio`. We could also add an explicit schema check for this config (not sure whats enforced in schema at present).

updated the comment, before we had no validation in `values.schema.json`. I added the validation for `containerRuntime`

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -650,7 +650,7 @@ logsCollection:
     # Set useSplunkIncludeAnnotation flag to `true` to collect logs from pods with `splunk.com/include: true` annotation and ignore others.
     # All other logs will be ignored.
     useSplunkIncludeAnnotation: false
-    # maxRecombineLogsSize sets the maximum size in bytes of a message recombined from cri-o, containerd and docker log entries.
+    # maxRecombineLogsSize sets the maximum size in bytes of a message recombined from crio, containerd and docker log entries.
```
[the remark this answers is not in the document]
omrozowicz-splunk: fixed

test/README.md
```diff
@@ -10,7 +10,7 @@
 ## Setup local environment
 
 #### Start Minikube
-    # Set environment variable for container runtime. Options are: docker, cri-o and containerd
+    # Set environment variable for container runtime. Options are: docker, crio and containerd
```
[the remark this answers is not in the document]
omrozowicz-splunk: fixed

omrozowicz-splunk:
> > Please update [this comment in values.yaml](https://github.com/signalfx/splunk-otel-collector-chart/blob/splunk-otel-collector-0.148.0/helm-charts/splunk-otel-collector/values.yaml#L592) which states `cri-o` is a valid containerRuntime values. Looking at the [stanza code](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/operator/parser/container/parser.go#L28), it should be `crio`. We could also add an explicit schema check for this config (not sure whats enforced in schema at present).
> 
> updated the comment, before we had no validation in `values.schema.json`. I added the validation for `containerRuntime`

Also I updated changelog entry to `breaking` as `containerRuntime: cri-o` will stop working from now on

helm-charts/splunk-otel-collector/values.schema.json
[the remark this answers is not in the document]
omrozowicz-splunk: upgrade fails with:
```
splunk-otel-collector:
- logsCollection.containers.containerRuntime: logsCollection.containers.containerRuntime must be one of the following: "", "docker", "containerd", "crio"
```

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -753,14 +671,14 @@ receivers:
...
+        default: post-multiline
       {{- range $.Values.logsCollection.containers.multilineConfigs }}
       - type: recombine
         id: {{ include "splunk-otel-collector.newlineKey" . | quote}}
-        output: clean-up-log-record
+        output: post-multiline
         source_identifier: resource["com.splunk.source"]
-        combine_field: attributes.log
-        is_first_entry: '(attributes.log) matches {{ .firstEntryRegex | quote }}'
+        combine_field: body
```
jinja2, line 706: Would this be a breaking change for users with extraOperators which might refer to `attributes.log` (from this change, it looks like with container parser this is now in `body`)? There might be other impacted fields too and I am worried that this might silently break extraOperators config. Do we have a full accounting of how the data available to `extraOperators` change with this PR?

Could we add a migration guide that maps old fields to new ones with a before/after example? A simple e.g.:

```
# Before
extraOperators:
  - type: filter
    expr: 'attributes.log matches "health_check"'

# After
extraOperators:
  - type: filter
    expr: 'body matches "health_check"'
```

omrozowicz-splunk: Good point! I forgot about extraOperators. But I see only `attributes.log` -> `body` is the case here. These are the mappings we're removing now:
```
  # Move out attributes to Attributes
  - type: move
    from: attributes.uid
    to: resource["k8s.pod.uid"]
  - type: move
    from: attributes.restart_count
    to: resource["k8s.container.restart_count"]
  - type: move
    from: attributes.container_name
    to: resource["k8s.container.name"]
  - type: move
    from: attributes.namespace
    to: resource["k8s.namespace.name"]
  - type: move
    from: attributes.pod_name
    to: resource["k8s.pod.name"]
  - type: move
    from: attributes.stream
    to: attributes["log.iostream"]
```

this is what `container` operator is doing by default according to the documentation. Also it extracts `attribute.time` that can be used in `extraOperators`. This is a snippet from `container` [docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/operators/container.md#parse-multiline-cri-container-logs-and-recombine-into-a-single-one):

```
{
  "timestamp": "2023-06-22 10:27:25.813799277 +0000 UTC",
  "body": "multiline containerd line that is super awesome",
  "attributes": {
    "time": "2023-06-22T10:27:25.813799277Z",
    "logtag": "F",
    "log.iostream":                "stdout",
    "log.file.path": "/var/log/pods/some_kube-controller-kind-control-plane_49cc7c1fd3702c40b2686ea7486091d6/kube-controller/1.log"
  },
  "resource": {
    "k8s.pod.name":                "kube-controller-kind-control-plane",
    "k8s.pod.uid":                 "49cc7c1fd3702c40b2686ea7486091d6",
    "k8s.container.name":          "kube-controller",
    "k8s.container.restart_count": "1",
    "k8s.namespace.name":          "some"
  }
}
```

omrozowicz-splunk: added note in changelog nad migration guide

jinja2: Hmm, I see time being deleted in the parser code though - https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/operator/parser/container/parser.go#L455. That doc snippet might not be accurate. @omrozowicz-splunk let's verify again the impact on extraOperator with actual container parser output, I am concerned about this change silently breaking user customizations

omrozowicz-splunk: right... so looks like there's a bug in the documentation. I ran it once again with the debug mode and this is what I see from the bare container:

```
2026-04-30T12:53:52.788+0200    info    ResourceLog #0
Resource SchemaURL: 
Resource attributes:
     -> k8s.pod.uid: Str(49cc7c1fd3702c40b2686ea7486091d6)
     -> k8s.container.name: Str(test-container)
     -> k8s.namespace.name: Str(default)
     -> k8s.pod.name: Str(test-pod)
     -> k8s.container.restart_count: Str(0)
ScopeLogs #0
ScopeLogs SchemaURL: 
InstrumentationScope  
LogRecord #0
ObservedTimestamp: 2026-04-30 10:53:47.588033 +0000 UTC
Timestamp: 2026-01-01 00:00:01 +0000 UTC
SeverityText: 
SeverityNumber: Unspecified(0)
Body: Str(CRIO_PARTIAL_STARTmiddle_chunkCRIO_PARTIAL_END)
Attributes:
     -> log.iostream: Str(stdout)
     -> logtag: Str(P)
     -> log.file.name: Str(0.log)
     -> log.file.path: Str(/Users/omrozowicz/otel-collector/container/default_test-pod_49cc7c1fd3702c40b2686ea7486091d6/test-container/0.log)
Trace ID: 
Span ID: 
Flags: 0
        {"resource": {"service.instance.id": "4fafd739-ecc4-4395-bfd0-7ef2215781d0", "service.name": "otelcol", "service.version": "v0.150.0"}, "otelcol.component.id": "debug", "otelcol.component.kind": "exporter", "otelcol.signal": "logs"}
```
indeed there's no `attributes.time` but the rest seems fine. Let me add this information to the PR.

omrozowicz-splunk: I'm testing it more with helm chart and indeed the behaviour is that no error is present (when we use `attributes.log` or `attributes.time`) and those values are visible as `null` in Splunk

UPGRADING.md
```diff
@@ -1,5 +1,60 @@
...
+#### `attributes.time` is no longer available in `extraOperators`
+
+Previously `attributes.time` contained the raw timestamp string and was available to `extraOperators`.
+The `container` operator sets the timestamp directly on the log record's `Timestamp` field and does not
+set `attributes.time`. Remove any references to `attributes.time` from your `extraOperators`.
+
+#### `extraOperators` using old operator IDs as `output` targets
+
+The following operator IDs no longer exist: `get-format`, `parser-docker`, `parser-crio`, `parser-containerd`,
+`crio-recombine`, `containerd-recombine`, `clean-up-log-record`.
```
jinja2, line 70: Q will the collector error on startup if any of these are still being referenced in extraOps? Or will it be a silent failure?

I am wondering if it is worth adding some extra check in the chart which fails install/upgrade for mentions of these in extraOperators

omrozowicz-splunk: it fails clearly with:
```
Error: failed to build pipelines: failed to create "file_log" receiver for data type "logs": operator 'clean-up-log-record' does not exist: {"operator_id":"add2"}
2026/05/06 08:49:46 main.go:139: application run finished with error: failed to build pipelines: failed to create "file_log" receiver for data type "logs": operator 'clean-up-log-record' does not exist: {"operator_id":"add2"}
```

UPGRADING.md
```diff
@@ -1,5 +1,60 @@
 # Upgrade guidelines
 
+## 0.150.0 to 0.151.0
```
jinja2:
```suggestion
## 0.151.0 to 0.152.0
```

UPGRADING.md
```diff
@@ -1,5 +1,60 @@
...
+    extraOperators:
+      - type: filter
+        expr: 'body matches "health_check"'
+```
+
+#### `attributes.time` is no longer available in `extraOperators`
+
+Previously `attributes.time` contained the raw timestamp string and was available to `extraOperators`.
+The `container` operator sets the timestamp directly on the log record's `Timestamp` field and does not
+set `attributes.time`. Remove any references to `attributes.time` from your `extraOperators`.
```
jinja2: Can't a user just update their operator logic to use `timestamp` field instead?
 
Could a user make below substitution instead of dropping the operator which uses attr.time?
 
```
# Before
- type: filter
  expr: 'attributes.time < "2026-01-01"'
# After
- type: filter
  expr: 'timestamp.Before(date("2026-01-01T00:00:00Z"))'
````

File: helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl (hunk 6 above)
Comment (jinja2) [reply helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl:706]: @omrozowicz-splunk can we make a note in the upgrade guide that these failures are silent (i assume even enabling debug logging won't show use of invalid/not present attrs)

Comment (omrozowicz-splunk) [reply helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl:706]: done

File: UPGRADING.md (hunk 8 above)
Comment (omrozowicz-splunk) [reply UPGRADING.md]: updated

File: UPGRADING.md (hunk 9 above)
Comment (omrozowicz-splunk) [reply UPGRADING.md]: Yes, it works this way, I just tested that. Updated the doc.

File: UPGRADING.md [hunk 10]
```diff
@@ -1,5 +1,75 @@
...
+logsCollection:
+  containers:
+    extraOperators:
+      - type: filter
+        expr: 'body matches "health_check"'
+```
+
+If `attributes.log` is still referenced, no error is raised — the value silently evaluates to `null`.
+
+#### `attributes.time` is no longer available in `extraOperators`
```
jinja2:
```suggestion
#### `attributes.time` is no longer available in `extraOperators`, use field `timestamp` instead
```

jinja2 (approved):
Left a small suggestion for one of the section headers, otherwise LGTM

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
jinja2: ok, i think we have covered the breaking changes from parser's output well enough in upgrading guide. Thanks

UPGRADING.md
omrozowicz-splunk: Added, also I updated changelog

PR #2381, opened 2026-04-15
[chart] Replace Target Allocator with upstream

crobert-1:

Previously, this chart had templates and logic implementing the target allocator functionality locally. This PR moves this functionality to simply directly rely on the [upstream target allocator helm chart](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-target-allocator) as a subchart.


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1325,42 +1325,24 @@ obi:
...
+# The Target Allocator is run as a deployment with a replica count of 1.
 # It discovers scraping configurations from ServiceMonitor and PodMonitor CRDs and
 # assigns them to collectors.
-# Related documentation: https://github.com/open-telemetry/opentelemetry-operator/tree/main/cmd/otel-allocator
+# Related documentation: https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-target-allocator
 # Not available in EKS Add-on.
 ################################################################################
 
-targetAllocator:
+targetallocator:
```
crobert-1, line 1338: This target allocator has been changed to be all lower case to accommodate the naming scheme of services in Kubernetes.

```
Error: UPGRADE FAILED: failed to create resource: ServiceAccount "soc-chart-targetAllocator-ta" is invalid: metadata.name: Invalid value: "soc-chart-targetAllocator-ta": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

The [service name](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/c738eda8d0bfa36513acdc9601165a308d9f506b/charts/opentelemetry-target-allocator/templates/service.yaml#L5) is being set by the [chart's name](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/c738eda8d0bfa36513acdc9601165a308d9f506b/charts/opentelemetry-target-allocator/templates/_helpers.tpl#L10), aka the alias set in `Chart.yaml`. The alias must match the values.yaml declaration to plumb the values down to the subchart, overriding its values.

The solution here is to either modify upstream's logic to override the passed in value and convert the service name to lower case, or rename this option. I'm open to suggestions either way.

examples/with-target-allocator/rendered_manifests/targetallocator/clusterRoleBinding.yaml
crobert-1, line 1: [Previous manifest](https://github.com/signalfx/splunk-otel-collector-chart/blob/9d1f0aae3243edcfbbfabcb8f6aca9e0d1f53ad5/examples/with-target-allocator/rendered_manifests/targetAllocator-clusterRoleBinding.yaml)

examples/with-target-allocator/rendered_manifests/targetallocator/deployment.yaml
crobert-1, line 1: [Previous manifest](https://github.com/signalfx/splunk-otel-collector-chart/blob/9d1f0aae3243edcfbbfabcb8f6aca9e0d1f53ad5/examples/with-target-allocator/rendered_manifests/targetAllocator-deployment.yaml)

examples/with-target-allocator/rendered_manifests/targetallocator/service.yaml
crobert-1, line 1: [Previous manifest](https://github.com/signalfx/splunk-otel-collector-chart/blob/9d1f0aae3243edcfbbfabcb8f6aca9e0d1f53ad5/examples/with-target-allocator/rendered_manifests/targetAllocator-service.yaml)

examples/with-target-allocator/rendered_manifests/targetallocator/serviceAccount.yaml
crobert-1, line 1: [Previous manifest](https://github.com/signalfx/splunk-otel-collector-chart/blob/9d1f0aae3243edcfbbfabcb8f6aca9e0d1f53ad5/examples/with-target-allocator/rendered_manifests/targetAllocator-serviceAccount.yaml)

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -567,13 +567,13 @@ receivers:
...
 
-  {{- if .Values.targetAllocator.enabled  }}
+  {{- if .Values.targetallocator.enabled  }}
   prometheus/ta:
     config:
       global:
         scrape_interval: 30s
     target_allocator:
-      endpoint: http://{{ template "splunk-otel-collector.fullname" . }}-ta.{{ template "splunk-otel-collector.namespace" . }}.svc.cluster.local:80
+      endpoint: http://{{ .Release.Name }}-targetallocator-ta.{{ template "splunk-otel-collector.namespace" . }}.svc.cluster.local:80
```
crobert-1: This change is because now the TA's service name is being generated in the subchart's service logic, it's no longer directly templated by the `splunk-otel-collector-chart`.

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -1446,71 +1446,13 @@
       "type": "object",
       "additionalProperties": true
     },
-    "targetAllocator": {
+    "targetallocator": {
       "description": "Target Allocator configuration.",
       "type": "object",
-      "additionalProperties": false,
+      "additionalProperties": true,
```
crobert-1, line 1452: Requiring strict checking on the sub-chart would end up limiting which versions of the upstream target allocator chart our SOCC would be compatible with. I think this choice makes sense for now to reduce maintenance burden, but there might be some kind of happy medium or alternative, feedback is welcome.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1325,42 +1325,24 @@ obi:
...
-    prometheus_cr:
-      enabled: true
-      scrapeInterval: 30s
-      # An empty value means any service monitor will be accepted.
-      service_monitor_selector: {}
-      # An empty value means any pod monitor will be accepted.
-      pod_monitor_selector: {}
-
-    filter_strategy: relabel-config
+  targetAllocator:
```
crobert-1, line 1340: I decided to only declare options that are not the subchart's defaults (other than the `image.repository`, as it might be odd to just include the `image.tag`). The most important part here is to include the `config.collector_selector.matchlabels` value to ensure the target allocator properly selects the collector agents for job assignments.

UPGRADING.md
```diff
@@ -1,5 +1,56 @@
 # Upgrade guidelines
 
+## 0.150.0 to 0.151.0
```
crobert-1, line 3: _I'll update version depending on when this gets merged_

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -567,13 +567,13 @@ receivers:
...
 
-  {{- if .Values.targetAllocator.enabled  }}
+  {{- if .Values.targetallocator.enabled  }}
   prometheus/ta:
     config:
       global:
         scrape_interval: 30s
     target_allocator:
-      endpoint: http://{{ template "splunk-otel-collector.fullname" . }}-ta.{{ template "splunk-otel-collector.namespace" . }}.svc.cluster.local:80
+      endpoint: http://{{ include "splunk-otel-collector.targetAllocatorFullname" . }}-ta.{{ template "splunk-otel-collector.namespace" . }}.svc.cluster.local:80
```
crobert-1, line 576: The collector's namespace should be safe here as the default helm install functionality is to install subcharts in the same namespace as the parent, unless overridden. (Our chart does not override the target allocator's namespace)

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -509,3 +481,19 @@ Usage:
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Create the target allocator endpoint to match sub-chart's naming logic
```
crobert-1:
```suggestion
Create the target allocator endpoint to match subchart's naming logic
```

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
crobert-1, line 576: The target allocator name here needs to be the fullname with `-ta` appended to match the [service name](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/c738eda8d0bfa36513acdc9601165a308d9f506b/charts/opentelemetry-target-allocator/templates/service.yaml#L5).

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -509,3 +481,19 @@ Usage:
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Create the target allocator endpoint to match subchart's naming logic
```
crobert-1, line 486: [reference](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/c738eda8d0bfa36513acdc9601165a308d9f506b/charts/opentelemetry-target-allocator/templates/_helpers.tpl#L6)

examples/with-target-allocator/rendered_manifests/targetallocator/configmap.yaml
```diff
@@ -1,26 +1,26 @@
...
+    app.kubernetes.io/name: targetallocator
     app.kubernetes.io/instance: default
-    app.kubernetes.io/version: "0.150.0"
-    app: splunk-otel-collector
-    chart: splunk-otel-collector-0.150.0
-    release: default
 data:
   targetallocator.yaml: |
-    allocation_strategy: per-node
+    allocation_strategy: consistent-hashing
```
jinja2: Can we keep per-node with a values.yaml override instead? I assume this applies to the daemonset

examples/with-target-allocator/rendered_manifests/targetallocator/service.yaml
```diff
@@ -0,0 +1,16 @@
+---
+# Source: splunk-otel-collector/charts/targetallocator/templates/service.yaml
+apiVersion: v1
+kind: Service
+metadata:
```
jinja2, line 5: Why are the normal labels for helm release missing from this object? I think we should fix them in upstream if its not our doing

unittests/image_pull_secrets_test.yaml
```diff
@@ -36,27 +36,6 @@ tests:
       - isNull:
           path: imagePullSecrets
 
-  # ---------------------------------------------------------------------------
-  # Target Allocator ServiceAccount (separate SA from the shared one)
-  # ---------------------------------------------------------------------------
-  - it: should set imagePullSecrets on the Target Allocator ServiceAccount
```
jinja2, line 42: Does upstream chart not have an option to set imagePullSecrets? Could be on the service account or directly on deployment. We'll have to update upstream if this is not configurable since we have customers pulling from private image repos

UPGRADING.md
```diff
@@ -1,5 +1,62 @@
...
+
+### Target allocator functionality now uses upstream chart as subchart
+
+Instead of relying on local versioning and implementation of target allocator functionality, the
```
jinja2, line 7: Can you try an helm upgrade scenario for this? I am wondering if the old resources are fully removed be helm with upgrade and new come up correctly.

helm-charts/splunk-otel-collector/values.yaml
jinja2: `targetallocator.targetAllocator` nesting can be bit confusing, but i can't think of a better alias name

unittests/image_pull_secrets_test.yaml
crobert-1: Yes, it does have the [`imagePullSecrets` option](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/ca193856683276606e609d563ef87dfd1b2c3b8f/charts/opentelemetry-target-allocator/values.yaml#L38). I removed validation from here since it was validating a local template, when now it's an upstream functionality. Let me know if we should still try to find some way to validate here.

examples/with-target-allocator/rendered_manifests/targetallocator/configmap.yaml
crobert-1: Right, `per-node` only works with the collector deployed as a daemonset. I'll update.

examples/with-target-allocator/rendered_manifests/targetallocator/service.yaml
crobert-1: Upstream [doesn't add any labels](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/ca193856683276606e609d563ef87dfd1b2c3b8f/charts/opentelemetry-target-allocator/templates/service.yaml#L3). Can you help me understand what those labels are used for, and why they need to be included here?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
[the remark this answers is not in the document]
crobert-1: Fixed, namespace override doesn't apply to subchart

UPGRADING.md
crobert-1: I've added upgrade tests. They're essentially just checking to make sure the old resources no longer exist and the new ones do exist. Ideally the tests would also be filtering resource types by labels, but until upstream has the labels for component names on each resource type it's not very effective to filter on the labels.

Let me know if you had something else in mind, or any other requests around upgrade testing.

examples/with-target-allocator/rendered_manifests/targetallocator/service.yaml
crobert-1: Coming back to this, the original target allocator service also [didn't have labels defined](https://github.com/signalfx/splunk-otel-collector-chart/blob/a7271858153e208acf008f14068429c96bc00573/examples/with-target-allocator/rendered_manifests/targetAllocator-service.yaml#L5), so this isn't a change in behavior from what I can tell. Is there a reason labels are now required when they weren't before?

unittests/image_pull_secrets_test.yaml
jinja2, line 42: Can we add new TA equivalent test here for completeness? Unless the subchart's template is not available in this framework.

examples/with-target-allocator/rendered_manifests/targetallocator/service.yaml
jinja2: > Is there a reason labels are now required when they weren't before?

They are not required, just standard practice to always label objects maintained by helm release with a [common set of labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/). With labels set, a user can easily look up objects in a release with `kubectl get -lrelease=my-release` etc. Helm itself tracks stuff with metadata in annotations but those are not selectable with kubectl like labels are.

This won't break our chart and we can still go ahead with this as-is. But I would prefer to fix upstream to have the labels added (I can make an update on upstream)

examples/with-target-allocator/rendered_manifests/targetallocator/serviceAccount.yaml
```diff
@@ -0,0 +1,15 @@
+---
+# Source: splunk-otel-collector/charts/targetallocator/templates/serviceAccount.yaml
+apiVersion: v1
+kind: ServiceAccount
+automountServiceAccountToken: false
```
jinja2, line 7: Ah, this is disabled by default in TA. Can you make sure that serviceaccount token is mounted in pod spec?

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1345,42 +1345,21 @@ obi:
...
+    config:
+      allocation_strategy: per-node
+      collector_selector:
+        matchlabels:
+          component: otel-collector-agent
```
jinja2, line 1404: Oh, should we have the release name here too? what was the old TAs matchlabels?

examples/with-target-allocator/rendered_manifests/targetallocator/deployment.yaml
```diff
@@ -0,0 +1,92 @@
...
+        app.kubernetes.io/part-of: opentelemetry-target-allocator
+        app.kubernetes.io/version: "0.126.0"
+        app.kubernetes.io/name: targetallocator
+        app.kubernetes.io/instance: default
+    spec:
+      serviceAccountName: default-targetallocator-ta
+      automountServiceAccountToken: false
+      containers:
+        - name: targetallocator
+          image: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator:0.126.0
```
jinja2: I think we are downgrading the image version here. Is this the latest subchart? In which case, maybe the subchart is not being release at the same cadence as the image for TA

unittests/image_pull_secrets_test.yaml
crobert-1: I've added an equivalent 👍

helm-charts/splunk-otel-collector/values.yaml
crobert-1: View the changes to `examples/with-target-allocator/rendered_manifests/targetallocator/configmap.yaml` (moved from the original target allocator's configmap). The `matchlabels` configured is identical to its value before.

Note also this is just getting moved in `values.yaml` to its new configurable location ([old config](https://github.com/signalfx/splunk-otel-collector-chart/blob/f4c820a67c3f5db95100b5a20425f98403660da1/helm-charts/splunk-otel-collector/values.yaml#L1416)).

examples/with-target-allocator/rendered_manifests/targetallocator/deployment.yaml
crobert-1: I've upgraded the subchart, bumping the version here.

examples/with-target-allocator/rendered_manifests/targetallocator/serviceAccount.yaml
crobert-1: Yes, the `serviceaccount-token` is mounted in the [pod spec of the deployment](https://github.com/signalfx/splunk-otel-collector-chart/pull/2381/files#diff-56710bc494fb25c984ae658b78f5ea03d60d421548e5e8eb594baee496894b0eR44).

jinja2 (approved):

PR #2384, opened 2026-04-16
[chore] precommit detect untracked rendered files

jinja2:
Catch errors in render-examples.sh

render-examples.sh runs render tasks as background subshells (&), but bash's wait with no arguments always returns 0 - even when children fail. This silently swallows render failures: `rm -rf` deletes the rendered manifests before helm template runs, and if helm template fails the deleted files are never recreated.

Additionally, in CI the pre-commit render hook relies on git diff to detect changes, which only sees modifications to tracked files. If a PR deletes rendered manifests, make render recreates them as untracked files that git diff cannot detect so the pre-commit check passes incorrectly.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
     hooks:
     - id: render
       name: Create the rendered Kubernetes manifest resources for the project examples
-      entry: make render
+      entry: bash ci_scripts/check-render.sh
       language: system
       pass_filenames: false
     - id: go-fmt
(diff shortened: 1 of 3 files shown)


jinja2:
The deleted rendered_manifests is intentional to verify that pre-commit check fails - https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/24533426471/job/71722309761?pr=2384. I'll add them back after verification

jinja2:
new output format for render - 
```

[enabled-pprof-extension] SUCCESS
[only-metrics-platform] SUCCESS
[splunk-platform-with-sourcetypes] SUCCESS
[enable-operator-and-auto-instrumentation] SUCCESS

========================================
ERROR: 1 example(s) failed to render:
  - /Users/jjain/work/splunk-otel-collector-chart/ci_scripts/../examples/default/
========================================
make: *** [render] Error 1
```

crobert-1 (approved):

PR #2388, opened 2026-04-20
[chore] throw proper error on index-out-of-range panic

jinja2:

Display proper error instead of [panic](https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/24587549741/job/72175386990?pr=2386#step:9:200) when file has no resources.


jinja2:
We probably need to add retry logic around the CopyFileFromPod + readAndNormalizeMetrics calls if we keep seeing this error

crobert-1 (approved):

PR #2390, opened 2026-04-22
Add workflow to certify collector dev image against functional tests

akhileshsingh85:
**Description:**
- Updated workflow_dispatch for functional_test_v2 workflow. 
- The workflow_dispatch will run all the test using collector dev image.
- Using the UI, user can also pass the desired docker build version.

Minor: 
Code refactor to move common code into action
Moved the collector images parsing logic in to common code.

Testing:
- Regular execution of the functional_test_v2 runs with latest GA version. 
- Manually triggered the functional_test workflow and verified that dev images is being used for testing.


.github/workflows/dev-image-test.yaml
```diff
@@ -0,0 +1,126 @@
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      - name: Get test matrix
+        id: test_matrix
+        run: |
+          matrix=$(jq -r '.functional_test_v2 | tostring' ci-matrix.json | tr -d '\n')
+          echo "matrix=${matrix}" >> "$GITHUB_OUTPUT"
+    outputs:
+      matrix: ${{ steps.test_matrix.outputs.matrix }}
+
+  functional-test:
```
jinja2: Can we refactor the functional-test actions to be re-usable actions instead?

.github/workflows/dev-image-test.yaml
```diff
@@ -0,0 +1,126 @@
+name: Dev Image Tests
+
+on:
+  schedule:
+    - cron: "0 6 * * *"
```
jinja2: Let's start with a manual dispatch first so it can be tested

.github/workflows/dev-image-test.yaml
```diff
@@ -0,0 +1,126 @@
...
+  set-image:
+    runs-on: ubuntu-latest
+    outputs:
+      image_repository: ${{ steps.parse.outputs.image_repository }}
+      image_tag: ${{ steps.parse.outputs.image_tag }}
+    steps:
+      - name: Parse image into repository and tag
+        id: parse
+        run: |
+          IMAGE="${COLLECTOR_IMAGE}"
```
jinja2: I think this would be better as a change in the test framework itself. All tests use the same helper to install the chart. We can accept image and value overrides via environment variables in there and merge into the test's values. Something like:

OTELCOL_IMAGE_REPO / OTELCOL_IMAGE_TAG - override the collector image
HELM_EXTRA_SET - optionally, set extra key=value overrides

.github/workflows/functional_test_v2.yaml
```diff
@@ -47,76 +52,95 @@ jobs:
     outputs:
       matrix: ${{ steps.test_matrix.outputs.matrix }}
 
+  collector-image-config:
```
akhileshsingh85, line 55: Common code to parse the collector name. Also evaluate that cloud test need to be executed.

.github/actions/functional-test/action.yaml
```diff
@@ -0,0 +1,117 @@
...
+      shell: bash
+      env:
+        KUBE_TEST_ENV: kind
+        K8S_VERSION: ${{ inputs.k8s-version }}
+        OTELCOL_IMAGE_REPO: ${{ inputs.otelcol-image-repo }}
+        OTELCOL_IMAGE_TAG: ${{ inputs.otelcol-image-tag }}
+      run: |
+        TEARDOWN_BEFORE_SETUP=true \
+          UPDATE_EXPECTED_RESULTS=${{ inputs.update-expected-results }} \
+          SUITE=${{ inputs.suite }} \
```
jinja2: can you move these to env too? and reference them with `SUITE="$SUITE" UPDATE_EXPECTED_RESULTS="$UPDATE_EXPECTED_RESULTS"` in run. Shell interpolation is supposed to be an anti-pattern, might as well fix it

functional_tests/internal/chart.go
```diff
@@ -128,6 +131,34 @@ func ChartInstallOrUpgrade(t *testing.T, testKubeConfig string, valuesFile strin
...
+func applyEnvOverrides(t *testing.T, values map[string]any) {
+	if repo := os.Getenv("OTELCOL_IMAGE_REPO"); repo != "" {
+		t.Logf("OTELCOL_IMAGE_REPO override: image.otelcol.repository=%s", repo)
+		require.NoError(t, strvals.ParseInto("image.otelcol.repository="+repo, values))
+	}
+	if tag := os.Getenv("OTELCOL_IMAGE_TAG"); tag != "" {
+		t.Logf("OTELCOL_IMAGE_TAG override: image.otelcol.tag=%s", tag)
+		require.NoError(t, strvals.ParseInto("image.otelcol.tag="+tag, values))
+	}
+	if extra := os.Getenv("HELM_EXTRA_SET"); extra != "" {
```
jinja2: We are not taking HELM_EXTRA_SET as input on workflow though

.github/workflows/functional_test_v2.yaml
```diff
@@ -47,76 +52,95 @@ jobs:
...
+
+          case "$EVENT_NAME" in
+            workflow_dispatch)
+              if [[ -n "$COLLECTOR_IMAGE" ]]; then
+                otelcol_image_repo="${COLLECTOR_IMAGE%:*}"
+                otelcol_image_tag="${COLLECTOR_IMAGE##*:}"
+                if [[ "$otelcol_image_repo" == "$otelcol_image_tag" ]]; then
+                  otelcol_image_tag="latest"
+                fi
+              fi
```
jinja2: Can we move this to go code instead? Then `OTELCOL_IMAGE=foo:bar make functionaltest` works locally with the same contract as CI

.github/workflows/functional_test_v2.yaml
```diff
@@ -47,76 +52,95 @@ jobs:
...
+
+              if [[ "$COLLECTOR_IMAGE" == *-dev* ]]; then
+                cloud_tests_enabled=true
+              fi
```
jinja2, line 78: Is the intent here to make sure the image is safe? I think a check for the image being from our docker repo like below might be better - 
```
workflow_dispatch)
  if [[ -n "$COLLECTOR_IMAGE" && "$COLLECTOR_IMAGE" != quay.io/signalfx/* ]]; then
    echo "::error::COLLECTOR_IMAGE must be a quay.io/signalfx/* image"
    exit 1
  fi
  cloud_tests_enabled=true
  ;;
```

functional_tests/internal/chart.go
akhileshsingh85: Removed it.

.github/workflows/functional_test_v2.yaml
akhileshsingh85: Made change to move the logic in the go code.

jinja2 (approved):

PR #2392, opened 2026-04-23
OTL-3816 Add an OTLP log ingest option in helm chart

akhileshsingh85:
**Description:** Add OTLP log ingest option to send logs to Splunk Connect for OTLP instead of HEC

A new `splunkPlatform.otlpIngest` configuration block allows routing logs via OTLP
  (gRPC or HTTP) to Splunk Connect for OTLP instead of the default HEC exporter.
  
  Enable it by setting:
  ```yaml
  splunkPlatform:
    logsEnabled: true
    otlpIngest:
      enabled: true
      endpoint: "<splunk-connect-for-otlp-endpoint>"
      protocol: grpc   # or "http"
      insecure: false
  ```

  When `otlpIngest.enabled` is true and `logsEnabled` is true, the agent, gateway,
  and cluster receiver log pipelines will use `otlp/platform_logs` (gRPC) or
  `otlp_http/platform_logs` (HTTP) exporters instead of `splunk_hec/platform_logs`.
  Use a base endpoint for HTTP, for example `http://splunk-host:4318`.
  The `splunkPlatform.endpoint` field is not required for logs when using OTLP
  ingest, but it is still required when `splunkPlatform.metricsEnabled` or
  `splunkPlatform.tracesEnabled` sends metrics or traces to Splunk Platform via HEC.

**Testing:** Added unit test


jinja2:
Can you add an example (I'd recommend this with some info on how to use the splunk connect TA as the destination) or add a diff of how the config is changing with this PR? It is very difficult to review the changes due to all the template functions.

akhileshsingh85:
@jinja2 Added example and also included the information into the advanced configuration README.md file.

.chloggen/otl-3816-otlp-log-ingest.yaml
```diff
@@ -0,0 +1,37 @@
...
+  A new `splunkPlatform.otlpIngest` configuration block allows routing logs via OTLP
+  (gRPC or HTTP) to Splunk Connect for OTLP instead of the default HEC exporter.
+
+  Enable it by setting:
+  ```yaml
+  splunkPlatform:
+    logsEnabled: true
+    otlpIngest:
+      enabled: true
+      endpoint: "<splunk-connect-for-otlp-endpoint>"
```
jinja2: Will this need a port to be set? can you update it with the proper form "host:port" or only "host" and port is derived from protocol?

docs/advanced-configuration.md
```diff
@@ -26,6 +26,109 @@ splunkObservability:
...
+The rendered Collector log pipeline changes from `splunk_hec/platform_logs` to the
+OTLP exporter selected by `splunkPlatform.otlpIngest.protocol`:
+
+```diff
+ processors:
++  resource/otlp_platform_logs:
++    attributes:
++      - key: com.splunk.index
++        value: main
++        action: insert
```
jinja2: Why do we need a new index processor? Shouldn't the same processor as the current hec pipeline has to be used?

docs/advanced-configuration.md
```diff
@@ -26,6 +26,109 @@ splunkObservability:
 clusterName: my-k8s-cluster
 ```
 
+## Send logs to Splunk Platform with Splunk Connect for OTLP
+
+By default, this chart sends logs to Splunk Enterprise or Splunk Cloud Platform with
+the HEC exporter configured by `splunkPlatform.endpoint` and `splunkPlatform.token`.
+To send logs to the Splunk Connect for OTLP Technical Add-on instead, enable
+`splunkPlatform.otlpIngest` and point `splunkPlatform.otlpIngest.endpoint` at the
+OTLP receiver exposed by the add-on. See the
```
jinja2, line 35: It is not clear to me if the token is needed or not when sending to the TA. If I set the token, will it be passed to TA or ignored?

test/otlp_schema_validation_test.go
```diff
@@ -0,0 +1,173 @@
...
+	args := append([]string{
+		"template", "test-release", "helm-charts/splunk-otel-collector",
+	}, extraArgs...)
+	cmd := exec.Command("helm", args...)
+	cmd.Dir = repoRoot
+	out, err := cmd.CombinedOutput()
+	return string(out), err
+}
+
+func TestOtlpLogIngestSchemaValidation(t *testing.T) {
```
jinja2: I don't think we need to introduce these types of tests. If you want to verify schema, can these be moved to the unittests instead?

.chloggen/otl-3816-otlp-log-ingest.yaml
```diff
@@ -0,0 +1,37 @@
...
+      protocol: grpc   # or "http"
+      insecure: false
+  ```
+
+  When `otlpIngest.enabled` is true and `logsEnabled` is true, the agent, gateway,
+  and cluster receiver log pipelines will use `otlp/platform_logs` (gRPC) or
+  `otlp_http/platform_logs` (HTTP) exporters instead of `splunk_hec/platform_logs`.
+  Use a base endpoint for HTTP, for example `http://splunk-host:4318`.
```
jinja2: I am not sure I understand what this means. Are you trying to say that for `protocol: http`, the user should add the scheme - `http` or `https` in the `endpoint` value? Can you rephrase this to be more clear?

.chloggen/otl-3816-otlp-log-ingest.yaml
```diff
@@ -0,0 +1,37 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add OTLP log ingest option to send logs to Splunk Connect for OTLP instead of HEC
+
+# One or more tracking issues related to the change
+issues: [2392]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
```
jinja2, line 16: I would suggest adding the link to advanced config section for further details

docs/advanced-configuration.md
```diff
@@ -26,6 +26,109 @@ splunkObservability:
...
++  metricsEnabled: true
+   otlpIngest:
+     enabled: true
+     endpoint: splunk-connect-for-otlp.example.com:4317
+     protocol: grpc
+```
+
+For review, the log destination changes from HEC to OTLP like this:
+
+```diff
```
jinja2: A few related thoughts on the docs changes -

- Looking at how our doc is structured today, we don't show rendered collector config in advanced-configuration.md even when a feature meaningfully changes pipelines - we show input values.yaml + a short prose/bullet explanation. Rendered output lives in `examples/<feature>/rendered_manifests/`, which this PR already adds. Inline rendered/diff snippets here will drift from the implementation and I recommend dropping the YAML diffs.

- Document which `splunkPlatform.*` settings apply to the OTLP exporter and which don't. In the PR, the OTLP exporter template only honors `splunkPlatform.otlpIngest.{endpoint,protocol,insecure}`, `splunkPlatform.index (as a resource attribute)`, and the shared `retryOnFailure / sendingQueue` blocks.
A short table here listing each `splunkPlatform.*` field and whether it applies to HEC, OTLP, or both would make this much easier for users to reason about.

- Quick question on TLS: since we're exposing `otlpIngest.insecure`, presumably Splunk Connect for OTLP's listener supports TLS. If so, it'd be good to also wire ca_file / cert_file / key_file / insecure_skip_verify (and timeout) through to the OTLP exporter . If there's no near-term use case for it, that's fine, just call it out in the docs as "not supported yet" and we can do it as a follow-up.

docs/advanced-configuration.md
jinja2: Ah, index is currently set on hec exporter, so we need this new processor for otlp

jinja2:
As a follow up, we can exercise the pipeline changes with our functional tests by enabling the otlp option with relevant tests like this [k8sevents](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/functional_tests/k8sevents/testdata/k8sevents_values.yaml.tmpl) one (we add new processors for otlp)

examples/splunk-connect-for-otlp/rendered_manifests/configmap-agent.yaml
```diff
@@ -0,0 +1,311 @@
...
+          - otlp/platform_logs
+          processors:
+          - memory_limiter
+          - k8s_attributes
+          - filter/logs
+          - batch
+          - resourcedetection
+          - resource
+          - resource/logs
+          - resource/otlp_platform_logs
```
jinja2, line 230: Can you make sure that the processors are maintaining the correct precedence? I believe the index/sourcetype set as annotations should always take priority. There is a functional test for this too, you could possibly add a test case with this otlp pipeline enabled to make sure it is working as intended.

jinja2:
Can you update the [NOTES here](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/NOTES.txt#L2) to prompt the with the otlp endpoint when the relevant settings are configured?

examples/splunk-connect-for-otlp/rendered_manifests/secret-splunk.yaml
```diff
@@ -0,0 +1,18 @@
...
+    app.kubernetes.io/name: splunk-otel-collector
+    helm.sh/chart: splunk-otel-collector-0.150.0
+    app.kubernetes.io/managed-by: Helm
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/version: "0.150.0"
+    app: splunk-otel-collector
+    chart: splunk-otel-collector-0.150.0
+    release: default
+type: Opaque
+data:
```
jinja2: Could we guard the secret creation on whether any field will actually be written, so OTLP-only users don't have a dead secret object?

akhileshsingh85: Resolved it.

docs/advanced-configuration.md
akhileshsingh85: Thanks for the TLS comment. Updated the cert properties.

.chloggen/otl-3816-otlp-log-ingest.yaml
akhileshsingh85: Updated the doc with endpoint example for protocol: http

test/otlp_schema_validation_test.go
akhileshsingh85: Moved to unittests.

helm-charts/splunk-otel-collector/templates/NOTES.txt
```diff
@@ -1,3 +1,6 @@
+{{- if eq (include "splunk-otel-collector.platformLogsViaOtlpEnabled" .) "true" }}
+Splunk OpenTelemetry Collector is installed and configured to send Splunk Platform logs to OTLP endpoint "{{ .Values.splunkPlatform.otlpIngest.endpoint }}".
+{{ end }}
 {{- if eq (include "splunk-otel-collector.splunkPlatformEnabled" .) "true" }}
 Splunk OpenTelemetry Collector is installed and configured to send data to Splunk Platform endpoint "{{ .Values.splunkPlatform.endpoint }}".
 {{ end }}
```
jinja2: Hmm, is the check `include "splunk-otel-collector.splunkPlatformEnabled" .` now sufficient for the case where splunkPlatform is enabled but only with logs + otlp?

jinja2:
@akhileshsingh85 the pre-commit check is failing. the updates lgtm otherwise

helm-charts/splunk-otel-collector/templates/NOTES.txt
akhileshsingh85: Updated the check for HEC endpoint.

unittests/notes_test.yaml
akhileshsingh85, line 1: Added test for the note.txt
It may not be required, but better to have some check. Let me know if this need to be removed.

akhileshsingh85:
> @akhileshsingh85 the pre-commit check is failing. the updates lgtm otherwise

Please trigger the copilot review. I also added few more checks for the otlp grpc host:port validation.

jinja2 (approved):

PR #2395, opened 2026-04-23
Use local action for creating PRs

breedx-splk:
In our continued effort to reduce the number of 3rd party potentially risky github actions, this PR adds a local github action that simply uses the `gh` commandline tool to create PRs. There were at least 4 usages of the external action, and those sites have all been changed to use this new local action.

Spice level: 🌶️🌶️ 

If we are relying on some unknown side effects or artifacts or weird behavior, it might not be captured, so proper review is appreciated. Thanks!

_(Disclaimer: AI assisted, human scrutinized.)_

### DIFF
--- a/.github/actions/create-pr/action.yaml
+++ b/.github/actions/create-pr/action.yaml
@@ -1,0 +1,50 @@
+name: Create PR
+description: Commit local changes, push a branch, and create or update a pull request with gh
+
+inputs:
+  branch:
+    description: Branch to push
+    required: true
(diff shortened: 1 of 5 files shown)


jinja2 (approved):
Merging this. Once its on main we can test run and make fixes to the workflows if needed

jinja2:
workflow run with these changes - https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/24853613330/job/72760485317

PR #2396, opened 2026-04-23
[chart] Replace deprecated filelog references with file_log

crobert-1:

Also, enforce no configurations reference the deprecated `filelog` alias. This is enforced to make sure no users have silent failures where they're attempting to customize chart configuration, but the receiver name they're attempting to reference no longer exists. Here's a [more thorough description](https://github.com/signalfx/splunk-otel-collector-chart/pull/2326#discussion_r3002347480) of the case we're concerned about.


examples/histogram-example/histogram-example-values.yaml
dmitryax, line 1: Why another example is needed here?

crobert-1: Added by accident, thanks for catching this! _I've removed._

dmitryax (approved):

PR #2398, opened 2026-04-24
OTL-3707 Update default toleration for aws batch node support

akhileshsingh85:
**Description:** Update Default toleration for aws batch node support

**Testing:** Existing Unit test


jinja2:
I think we can just add the tolerations in values.yaml. We probably add some for control plane already. On non-batch node clusters, the tolerations will be harmless and users can always override with their own set with custom values.

akhileshsingh85:
@jinja2 Updated the PR with just the value.yaml file update.

.chloggen/eks-batch-nodes.yaml
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Document how to run the Collector agent daemonset on AWS Batch-managed EKS nodes
```
jinja2: This isn't just a doc change anymore, please update to mention that the default list of toleration now includes AWS batch nodes

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -809,6 +809,13 @@ tolerations:
   - key: node-role.kubernetes.io/infra
     effect: NoSchedule
     operator: Exists
+  # AWS Batch node taint - allows scheduling on Batch-managed nodes.
+  - key: batch.amazonaws.com/batch-node
+    operator: Exists
+    effect: NoSchedule
```
jinja2: With `operator: Exists` and no effect specified, the toleration will match both the effects. Unless the AWS docs mention not to do this for some reason, I think we can keep just one of these without the effect set.

.chloggen/eks-batch-nodes.yaml
akhileshsingh85: Update the note and subtext.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -809,6 +810,9 @@ tolerations:
   - key: node-role.kubernetes.io/infra
     effect: NoSchedule
     operator: Exists
+  # AWS Batch node taint - allows scheduling on Batch-managed nodes.
```
jinja2:
```suggestion
```
We already document it at tolerations

akhileshsingh85: Thanks. Removed the online comment from here.

jinja2 (approved):

PR #2404, opened 2026-04-29
agent: add component label to ds

jinja2:

- Adds `component: otel-collector-agent` to the agent DaemonSet's `spec.selector.matchLabels`, gated behind `featureGates.daemonSetComponentSelector`. The agent, gw and clusterreceiver all share `app` and `release` labels. The old ds selector only matched on those two labels, some tools that use selectors to group pods would see gw and cluster-receiver pods as belonging to the agent ds. Adding `component: otel-collector-agent` to the selector disambiguates the three workloads.

- Existing installations are not affected - the feature gate defaults to false so helm upgrade works without any manual steps. Users who need the fix can opt in and follow the documented steps.


.chloggen/daemonset-component-selector.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
jinja2, line 2: technically a bug_fix, but it is not default enabled so enhancement seems appropriate

jinja2:
The error from eks-addon is legit. I think the Release.Install check will be an issue for it

.chloggen/daemonset-component-selector.yaml
```diff
@@ -0,0 +1,17 @@
...
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add `component: otel-collector-agent` to the agent DaemonSet's selector.matchLabels"
```
crobert-1, line 6: It might make more sense to have the `note` be the actual purpose/reasoning behind this change, and then move the label and technical notes to the `subtext`?

UPGRADING.md
```diff
@@ -1,5 +1,71 @@
 # Upgrade guidelines
 
+## 0.150.0 to 0.151.0
+
+### New feature gate: `component` label in agent DaemonSet selector
+
+A new feature gate `featureGates.daemonSetComponentSelector` (disabled by
+default) adds `component: otel-collector-agent` to the agent DaemonSet's
+`spec.selector.matchLabels`. This prevents workload-management tools from incorrectly grouping gateway and cluster-receiver pods under the
```
crobert-1:
```suggestion
`spec.selector.matchLabels`. This prevents workload management tools from incorrectly grouping gateway and cluster receiver pods under the
```
Nit: In searching the internet it looks like both of these are commonly just written as two words.

UPGRADING.md
```diff
@@ -1,5 +1,71 @@
 # Upgrade guidelines
 
+## 0.150.0 to 0.151.0
```
crobert-1:
```suggestion
## 0.151.0 to 0.152.0
```

crobert-1 (approved):
_Just some nits_

PR #2416, opened 2026-05-09
[draft] Multi-tenant logs and customizable log sourcetype

jinja2:
Draft chart implementation for supporting different log routing config. Not for merge - shared as a small draft so the proposed values surface, helpers, and rendered pipeline shape are easy to visualize to start work.


jinja2:
This has gone stale, I'll open a new one.

PR #2425, opened 2026-05-20
feat: mounting secret as a file instead of using env variables

dhruvanshus-crest:
**Description:** mounting secret as a file instead of using env variables
 Mounting secret as a file instead of env variables for both Splunk HEC token and Splunk observability access token to fix the security vulnerability - [SITG](https://www.stigviewer.com/stigs/kubernetes/2025-05-16/finding/V-242415)

**Link to ADDON Jira:** [ADDON-87370](https://splunk.atlassian.net/browse/ADDON-87370)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -828,6 +830,12 @@ rbac:
 secret:
   create: true
   name: ""
+  # File mode for files mounted from the Splunk secret. The default is group
+  # read-only so token files are not readable by other users in the container.
+  defaultMode: "0440"
+  # fsGroup applied to pods that mount the Splunk secret, unless the pod already
+  # defines fsGroup. This lets non-root containers read the group-readable token files.
+  fsGroup: 999
```
jinja2: Can we move this to workload's podSecurityContext setting instead?  `gateway.podSecurityContext`/`clusterReceiver.podSecurityContext` already exist and are the conventional home for `fsGroup`. `secret.fsGroup` reads as scoped to just the secret but it is pod-wide and every fsGroup-eligible volume in the pod is affected. The agent only has a container-level `securityContext`, which can't carry `fsGroup`, so it needs a new `agent.podSecurityContext` (this also makes the agent consistent with the other workloads).

jinja2:
We have a set of tests that run the chart in hosted clusters, but a fork PR won't trigger them. I'd prefer these changes run against those tests; we should look into what's needed internally for proper access.

A few things these tests may answer (we have an AKS cluster with Windows nodes and an OpenShift cluster wired into CI) - or, if you've already looked into them, please comment:

1. Windows - I understand fsGroup isn't supported on Windows. Have we verified that `${file:/otel/etc/...}` resolves on Windows and is readable by the collector process?
2. OpenShift - OpenShift admits every pod through a [SecurityContextConstraints](https://docs.redhat.com/en/documentation/openshift_container_platform/4.11/html/authentication_and_authorization/managing-pod-security-policies) (SCC). The default SCC uses `fsGroup: MustRunAs` with a per-namespace allocated GID range, and OpenShift normally auto-assigns one from that range. The [SCC our chart creates allows any fsGroup](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/securityContextConstraints.yaml#L28-L29), but that's configurable. I want to confirm whether GID 999 falls within the default OpenShift SCC's allowed range. At a minimum we should document this for OpenShift users.

helm-charts/splunk-otel-collector/values.yaml
dhruvanshus-crest: yes, sure. I've moved this to `podSecurityContext`, removed `secret.fsGroup` and added new `agent.podSecurityContext`.

dhruvanshus-crest:
Closing this pull request as the fork PR doesn't trigger the tests that run the chart in hosted cluster. Raised the PR - https://github.com/signalfx/splunk-otel-collector-chart/pull/2441

PR #2429, opened 2026-05-27
Add entity events mode for k8s property updates

dmitryax:
Add a new chart-level feature gate splunkObservability.entityEventsEnabled (disabled by default) that switches the k8s_cluster receiver from PATCH to PUT (entity events) mode for property updates.

- When enabled, sets metadata_exporters to empty on k8s_cluster receiver
- Adds logs/entity_events pipeline routing k8s_cluster data through signalfx exporter with memory_limiter and resource processors
- Enables exporter.signalfx.consumeEntityEvents feature gate on the collector

This allows users to opt into entity events semantics while maintaining backward compatibility. When the gate is disabled (default), the chart produces identical output to today.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
@@ -3,7 +3,7 @@ Config for the otel-collector k8s cluster receiver deployment.
 The values can be overridden in .Values.clusterReceiver.config
 */}}
 {{- define "splunk-otel-collector.clusterReceiverConfig" -}}
-{{- $useK8sEntityEvents := and (eq (include "splunk-otel-collector.o11yMetricsEnabled" .) "true") .Values.featureGates.useK8sEntityEvents -}}
+{{- $useEntityEventsForK8sProperties := and (eq (include "splunk-otel-collector.o11yMetricsEnabled" .) "true") .Values.featureGates.useEntityEventsForK8sProperties -}}
 extensions:
   {{- include "splunk-otel-collector.opampExtension" . | nindent 2 }}
   health_check:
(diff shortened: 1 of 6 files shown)


helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1468,6 +1468,11 @@ featureGates:
   # Requires Splunk Observability to be configured (splunkObservability.realm or ingestUrl must be set).
   # This feature is experimental and may change or be removed in future releases.
   enableK8sEntities: false
+  # Enable entity events mode for Kubernetes property updates. When enabled, the k8s_cluster receiver
+  # skips metadata_exporters (PATCH mode) and routes entity events to the signalfx exporter using PUT semantics.
+  # Also enables the exporter.signalfx.consumeEntityEvents collector feature gate on the cluster receiver.
+  # Requires splunkObservability.metricsEnabled.
+  useK8sEntityEvents: false
```
jinja2: Can we rename this? With the existing featuregate `enableK8sEntities` similarly named, I find this new config name confusing. What about `useEntityEventsForK8sProperties`?

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -307,7 +310,7 @@ exporters:
       "X-Splunk-Instrumentation-Library": o11yevents
   {{- end }}
 
-  {{- if eq (include "splunk-otel-collector.k8sEntitiesEnabled" .) "true" }}
+  {{- if and (eq (include "splunk-otel-collector.splunkO11yEnabled" .) "true") .Values.featureGates.enableK8sEntities }}
```
jinja2: Any specific reason for this change? The helper `splunk-otel-collector.k8sEntitiesEnabled` evaluates the same condition https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/_helpers.tpl#L388

jinja2 (approved):

PR #2432, opened 2026-05-28
[agent, gateway] Remove SignalFx exporter from trace pipelines

crobert-1:

The `signalfx` exporter is no longer required to be used in trace pipelines to enable trace correlation.

### DIFF
--- a/.chloggen/removesignalfxtraces.yaml
+++ b/.chloggen/removesignalfxtraces.yaml
@@ -5,7 +5,7 @@ component: agent, gateway
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: Remove `signalfx` exporter from trace pipelines
 # One or more tracking issues related to the change
-issues: []
+issues: [2432]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
--- a/.chloggen/removesignalfxtraces.yaml
+++ b/.chloggen/removesignalfxtraces.yaml
@@ -1,0 +1,14 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: agent, gateway
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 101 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -1261,11 +1257,7 @@ service:
         {{- end }}
         - resourcedetection
         - resource
-        {{/*
-        The attribute `deployment.environment` is not being set on metrics sent to Splunk Observability because it's already synced as the `sf_environment` property.
-        More details: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/signalfxexporter#traces-configuration-correlation-only
-        */}}
-        {{- if (and .Values.splunkPlatform.metricsEnabled .Values.environment) }}
+        {{- if .Values.environment }}
```
dmitryax: This should not be changed. We still don't need to send `deployment.environment` with metrics

crobert-1: Reverted

.chloggen/removesignalfxtraces.yaml
```diff
@@ -0,0 +1,14 @@
...
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove `signalfx` exporter from trace pipelines
+# One or more tracking issues related to the change
+issues: [2432]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The `signalfx` exporter is no longer required for trace correlation. Sending traces via OTLP will
+  enable all trace correlation functionality.
```
jinja2, line 14: `Sending traces via OTLP will enable all trace correlation functionality.` Should we maybe drop this? It reads like otlp has some correlation functionality, which is not the case. It is just that backend now handles this. Well, we do send the right metadata in trace with otlp 🤔 . Feel free to ignore this.

jinja2 (approved):

.chloggen/removesignalfxtraces.yaml
crobert-1: I agree this may be misleading, I'm not sure what's best. My goal was to try to make it clear that trace correlation is still supported in our backend, but I'm not sure how best to convey that.

crobert-1: I think I'll leave it as-is for now, but happy to reword in any way that's more clear.

PR #2435, opened 2026-06-01
[chore] fix create-pr composite action

jinja2:

Fixes the `create-pr` composite action, which broke the scheduled dependency-update workflows (e.g. "Check for new chart dependency updates") with `! [rejected] ... (stale info)` on push.

`actions/checkout` only fetches the base branch, so there was no remote-tracking ref for the target branch (`update-operator`, etc.). When the branch already existed on the remote from a prior run, `git push --force-with-lease` couldn't establish a valid lease and was rejected.

The action now fetches the target branch first so `--force-with-lease` has an up-to-date base to compare against (keeping its safety), and skips committing/pushing when there's nothing staged instead of failing the step on an empty diff.

### DIFF
--- a/.github/actions/create-pr/action.yaml
+++ b/.github/actions/create-pr/action.yaml
@@ -36,8 +36,15 @@ runs:
       run: |
         git config user.name "${{ inputs.author-name }}"
         git config user.email "${{ inputs.author-email }}"
+        # Fetch the remote branch (if it already exists from a previous run) so that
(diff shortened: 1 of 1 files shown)


jinja2:
Workflow run with PR branch - https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/26776048278/job/78927819259

Workflow run with error - https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/26775246545

crobert-1 (approved):

PR #2440, opened 2026-06-02
add SecureApp auto-instrumentation support via OTel Operator

rahulsplunk:
## Summary

- **D1 (`operator/_helpers.tpl`)**: When `secureAppEnabled=true`, swap the Java instrumentation image to the CSA variant (`secureAppCsaImage`), and inject `SPLUNK_SECUREAPP_AGENT_ENABLED=true` for Node.js (no image swap needed — `@splunk/secureapp-agent` is bundled in `@splunk/otel` v4.7.0+).
- **D2 (`config/_otel-agent.tpl`)**: Add `routing/logs` connector, `logs/split` pipeline (OTLP entrypoint), `logs/secureapp` pipeline, and `otlp_http/secureapp` exporter with `X-Splunk-Instrumentation-Library: secureapp` header. Routes secureapp-scoped OTel logs to `/v3/event` on the correct internal ingest domain. The `default_pipelines` on the routing connector is conditional — `[logs]` when `logsEnabled=true`, else `[]` — to avoid referencing a non-existent pipeline in pure-SecureApp mode.
- **`values.yaml` / `values.schema.json`**: Add `secureAppEnabled` (bool) and `secureAppCsaImage` (string) fields under `splunkObservability`.
- **Unit tests**: New `unittests/secureapp_test.yaml` covering all 8 S/L/P combinations; updated `operator_customized_test.yaml` for Node.js injection.

## Validation

Validated end-to-end on Docker Desktop Kubernetes → lab0:
- **Java**: Attack events (SSRF, SQL, RCE, Log4Shell, Deserialization, Command) appear in SecureApp UI under `lab0-experiment` / `java-secureapp-test-app` within ~45s of triggering.
- **Node.js**: 72 runtime vulnerabilities detected and visible in SecureApp UI under `lab0-experiment` / `node-secureapp-test-app` (packages: `handlebars`, `axios`, `tar`, `lodash`, etc.).

## Test plan

- [ ] `helm template` with `secureAppEnabled=true` and each combination of `logsEnabled`/`profilingEnabled` — verify no pipeline reference errors
- [ ] `helm unittest` passes (`helm plugin install https://github.com/helm-unittest/helm-unittest` then `helm unittest helm-charts/splunk-otel-collector`)
- [ ] Deploy to a test cluster with `secureAppEnabled=true` — verify Java attack events appear in SecureApp UI
- [ ] Deploy with `secureAppEnabled=true` and a Node.js instrumented app — verify vulnerability findings appear in SecureApp UI
- [ ] Deploy with `secureAppEnabled=false` (default) — verify no change to existing pipelines

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.chloggen/secureapp-agent-resource-processor.yaml
+++ b/.chloggen/secureapp-agent-resource-processor.yaml
@@ -1,0 +1,4 @@
+change_type: enhancement
+component: agent
+note: Add `resource` processor to `logs/secureapp` pipeline to inject `k8s.cluster.name` and `k8s.node.name` for vulnerability host mapping
+issues: [2440]
--- a/.chloggen/secureapp-secapp-1072.yaml
+++ b/.chloggen/secureapp-secapp-1072.yaml
@@ -1,9 +1,0 @@
-change_type: enhancement
-component: agent
-note: Add `resource` processor to `logs/secureapp` pipeline for k8s host attribute enrichment (SECAPP-1072)
-issues: [2440]
-subtext: |
-  The `resource` processor is now included in the `logs/secureapp` pipeline to inject
-  `k8s.cluster.name` and `k8s.node.name` attributes required for vulnerability host mapping.
-  Also fixes a pre-existing issue where `resource/add_environment` was unconditionally
-  referenced but is only defined when `.Values.environment` is set.
--- a/.chloggen/secureapp-secapp-1072.yaml
+++ b/.chloggen/secureapp-secapp-1072.yaml
@@ -1,0 +1,9 @@
+change_type: enhancement
+component: agent
+note: Add `resource` processor to `logs/secureapp` pipeline for k8s host attribute enrichment (SECAPP-1072)
+issues: [2440]
+subtext: |
+  The `resource` processor is now included in the `logs/secureapp` pipeline to inject
+  `k8s.cluster.name` and `k8s.node.name` attributes required for vulnerability host mapping.
+  Also fixes a pre-existing issue where `resource/add_environment` was unconditionally
+  referenced but is only defined when `.Values.environment` is set.
--- a/examples/enable-operator-and-auto-instrumentation/rendered_manifests/operator/admission-webhooks/operator-webhook.yaml
+++ b/examples/enable-operator-and-auto-instrumentation/rendered_manifests/operator/admission-webhooks/operator-webhook.yaml
@@ -1,5 +1,11 @@
 ---
 # Source: splunk-otel-collector/charts/operator/templates/admission-webhooks/operator-webhook.yaml
+---
+---
+# Source: splunk-otel-collector/charts/operator/templates/admission-webhooks/operator-webhook.yaml
+---
+---
+# Source: splunk-otel-collector/charts/operator/templates/admission-webhooks/operator-webhook.yaml
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
@@ -193,12 +199,6 @@ webhooks:
         scope: Namespaced
     sideEffects: None
     timeoutSeconds: 10
----
-# Source: splunk-otel-collector/charts/operator/templates/admission-webhooks/operator-webhook.yaml
----
----
-# Source: splunk-otel-collector/charts/operator/templates/admission-webhooks/operator-webhook.yaml
----
 ---
 # Source: splunk-otel-collector/charts/operator/templates/admission-webhooks/operator-webhook.yaml
 apiVersion: v1
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -1095,7 +1095,15 @@ service:
     # secure application events — always routed via routing/logs connector
     logs/secureapp:
       receivers: [routing/logs]
-      processors: [k8s_attributes, resourcedetection, resource/add_environment, memory_limiter, batch]
+      processors:
+        - k8s_attributes
+        - resourcedetection
+        - resource
+        {{- if .Values.environment }}
+        - resource/add_environment
+        {{- end }}
+        - memory_limiter
+        - batch
       exporters:
         {{- if .Values.gateway.enabled }}
         - otlp_grpc
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -1040,13 +1040,14 @@ exporters:
   {{- end }}
   {{- end }}
 
-{{- if and
-  (or (eq (include "splunk-otel-collector.logsEnabled" .) "true") (eq (include "splunk-otel-collector.profilingEnabled" .) "true"))
-  (.Values.splunkObservability.secureAppEnabled)
-}}
+{{- if .Values.splunkObservability.secureAppEnabled }}
 connectors:
   routing/logs:
+    {{- if or (eq (include "splunk-otel-collector.logsEnabled" .) "true") (eq (include "splunk-otel-collector.profilingEnabled" .) "true") }}
     default_pipelines: [logs]
+    {{- else }}
+    default_pipelines: []
+    {{- end }}
     table:
       - context: log
         condition: instrumentation_scope.name == "secureapp"
@@ -1091,28 +1092,23 @@ service:
   # In order to disable a default pipeline just set it to `null` in agent.config overrides.
   pipelines:
     {{- if .Values.splunkObservability.secureAppEnabled }}
-    # secure application events
+    # secure application events — always routed via routing/logs connector
     logs/secureapp:
-      receivers:
-        {{- if or (eq (include "splunk-otel-collector.logsEnabled" .) "true") (eq (include "splunk-otel-collector.profilingEnabled" .) "true") }}
-        - routing/logs
-        {{- else }}
-        - otlp
-        {{- end }}
-      processors: [memory_limiter, batch]
+      receivers: [routing/logs]
+      processors: [k8s_attributes, resourcedetection, resource/add_environment, memory_limiter, batch]
       exporters:
         {{- if .Values.gateway.enabled }}
         - otlp_grpc
         {{- else }}
         - otlp_http/secureapp
         {{- end }}
-    {{- end }}
-    {{- if or (eq (include "splunk-otel-collector.logsEnabled" .) "true") (eq (include "splunk-otel-collector.profilingEnabled" .) "true") }}
-    {{- if .Values.splunkObservability.secureAppEnabled }}
+    # receives all OTLP logs; routes secureapp scope to logs/secureapp, rest to default_pipelines
     logs/split:
       receivers: [otlp]
+      processors: [memory_limiter, batch]
       exporters: [routing/logs]
     {{- end }}
+    {{- if or (eq (include "splunk-otel-collector.logsEnabled" .) "true") (eq (include "splunk-otel-collector.profilingEnabled" .) "true") }}
     # default logs + profiling data pipeline
     logs:
       receivers:
--- a/helm-charts/splunk-otel-collector/values.schema.json
+++ b/helm-charts/splunk-otel-collector/values.schema.json
@@ -1554,6 +1554,11 @@
       "description": "Helm Chart Feature Gates.",
       "type": "object",
       "additionalProperties": true
+    },
+    "opentelemetry-ebpf-instrumentation": {
+      "description": "OpenTelemetry eBPF Instrumentation subchart configuration.",
+      "type": "object",
+      "additionalProperties": true
     }
   },
   "allOf": [
--- a/helm-charts/splunk-otel-collector/values.schema.json
+++ b/helm-charts/splunk-otel-collector/values.schema.json
@@ -280,6 +280,13 @@
               "const": true
             }
           }
+        },
+        {
+          "properties": {
+            "secureAppEnabled": {
+              "const": true
+            }
+          }
         }
       ]
     },
--- a/helm-charts/splunk-otel-collector/values.schema.json
+++ b/helm-charts/splunk-otel-collector/values.schema.json
@@ -250,10 +250,6 @@
           "description": "Whether Secure Application features are enabled for Splunk Observability",
           "type": "boolean"
         },
-        "secureAppCsaImage": {
-          "description": "CSA init-container image used when secureAppEnabled=true",
-          "type": "string"
-        },
         "infrastructureMonitoringEventsEnabled": {
           "description": "Send Kubernetes events to Splunk Observability Infrastructure Monitoring",
           "type": "boolean"
--- a/helm-charts/splunk-otel-collector/values.schema.json
+++ b/helm-charts/splunk-otel-collector/values.schema.json
@@ -250,6 +250,10 @@
           "description": "Whether Secure Application features are enabled for Splunk Observability",
           "type": "boolean"
         },
+        "secureAppCsaImage": {
+          "description": "CSA init-container image used when secureAppEnabled=true",
+          "type": "string"
+        },
         "infrastructureMonitoringEventsEnabled": {
           "description": "Send Kubernetes events to Splunk Observability Infrastructure Monitoring",
           "type": "boolean"
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -221,12 +221,10 @@ splunkObservability:
 
   # CSA init-container image used when secureAppEnabled=true.
   # Overrides instrumentation.spec.java.image for SecureApp instrumentation.
-  # PREREQUISITE (D3): Update this to the real published tag before merging — the
-  # splunk-otel-java-csa image has not yet been released. See https://github.com/signalfx/splunk-otel-java/releases
   # TODO(naming): Consider whether to auto-derive this from the APM image by appending '-csa'
   # to the repo name (e.g. splunk-otel-java:v2.27.0 -> splunk-otel-java-csa:v2.27.0).
   # Requires agent team to confirm lockstep release contract before implementing.
-  secureAppCsaImage: ghcr.io/signalfx/splunk-otel-java/splunk-otel-java-csa:PENDING_RELEASE
+  secureAppCsaImage: ghcr.io/signalfx/splunk-otel-java/splunk-otel-java-csa:v2.28.0
 
 ################################################################################
 # Logs collection engine:
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -219,12 +219,6 @@ splunkObservability:
   # Enables Secure Application features
   secureAppEnabled: false
 
-  # CSA init-container image used when secureAppEnabled=true.
-  # Overrides instrumentation.spec.java.image for SecureApp instrumentation.
-  # TODO(naming): Consider whether to auto-derive this from the APM image by appending '-csa'
-  # to the repo name (e.g. splunk-otel-java:v2.27.0 -> splunk-otel-java-csa:v2.27.0).
-  # Requires agent team to confirm lockstep release contract before implementing.
-  secureAppCsaImage: ghcr.io/signalfx/splunk-otel-java/splunk-otel-java-csa:v2.28.0
 
 ################################################################################
 # Logs collection engine:
--- a/helm-charts/splunk-otel-collector/values.yaml
+++ b/helm-charts/splunk-otel-collector/values.yaml
@@ -219,6 +219,15 @@ splunkObservability:
   # Enables Secure Application features
   secureAppEnabled: false
 
+  # CSA init-container image used when secureAppEnabled=true.
+  # Overrides instrumentation.spec.java.image for SecureApp instrumentation.
+  # PREREQUISITE (D3): Update this to the real published tag before merging — the
+  # splunk-otel-java-csa image has not yet been released. See https://github.com/signalfx/splunk-otel-java/releases
+  # TODO(naming): Consider whether to auto-derive this from the APM image by appending '-csa'
+  # to the repo name (e.g. splunk-otel-java:v2.27.0 -> splunk-otel-java-csa:v2.27.0).
+  # Requires agent team to confirm lockstep release contract before implementing.
+  secureAppCsaImage: ghcr.io/signalfx/splunk-otel-java/splunk-otel-java-csa:PENDING_RELEASE
+
 ################################################################################
 # Logs collection engine:
 # Only `otel` (native OpenTelemetry log collection) is supported.
--- a/unittests/resource_detection_test.yaml
+++ b/unittests/resource_detection_test.yaml
@@ -568,7 +568,7 @@ tests:
           path: data["relay"]
           pattern: "logs/k8s_entities:\\n\\s+exporters:\\n(?:\\s+- .+\\n)+\\s+processors:\\n(?:\\s+- .+\\n)*\\s+- resourcedetection/k8s_cluster_name\\n(?:\\s+- .+\\n)*\\s+receivers:"
 
-  - it: "agent — logs/secureapp pipeline has no resourcedetection"
+  - it: "agent — logs/secureapp pipeline has resourcedetection processor"
     set:
       distribution: openshift
       cloudProvider: aws
@@ -576,7 +576,7 @@ tests:
         secureAppEnabled: true
     template: configmap-agent.yaml
     asserts:
-      - notMatchRegex:
+      - matchRegex:
           path: data["relay"]
           pattern: "logs/secureapp:\\n\\s+exporters:\\n(?:\\s+- .+\\n)+\\s+processors:\\n(?:\\s+- .+\\n)*\\s+- resourcedetection\\n(?:\\s+- .+\\n)*\\s+receivers:"
 
--- a/unittests/secureapp_test.yaml
+++ b/unittests/secureapp_test.yaml
@@ -1,0 +1,256 @@
+suite: splunk-otel-collector.secureapp
+values:
+  - ./values/basic.yaml
+release:
+  name: test-release
+  namespace: test-ns
+templates:
+  - configmap-agent.yaml
+  - configmap-gateway.yaml
+
+# ============================================================================
+# SecureApp collector config tests
+#
+# Covers all 8 S/L/P combinations (S=secureAppEnabled, L=logsEnabled,
+# P=profilingEnabled) for the agent configmap. Key invariants:
+#   - S=0: routing connector, logs/split, logs/secureapp are all absent
+#   - S=1: routing connector, logs/split, logs/secureapp are always present
+#   - default_pipelines: [] when S=1,L=0,P=0
+#   - default_pipelines: [logs] when S=1 and (L=1 OR P=1)
+#   - logs/secureapp receivers always [routing/logs], never [otlp]
+#   - logs/secureapp processors: [k8s_attributes, resourcedetection,
+#       resource/add_environment, memory_limiter, batch]
+#   - logs/split processors: [memory_limiter, batch] only (no enrichment)
+#   - otlp_http/secureapp.logs_endpoint uses o11yIngestUrl helper (realm-aware)
+# ============================================================================
+
+tests:
+
+  # --------------------------------------------------------------------------
+  # S=0: secureApp disabled — no routing connector, no split, no secureapp
+  # --------------------------------------------------------------------------
+
+  - it: "S=0 — agent: routing/logs connector is absent"
+    set:
+      splunkObservability:
+        secureAppEnabled: false
+    template: configmap-agent.yaml
+    asserts:
+      - notMatchRegex:
+          path: data["relay"]
+          pattern: "routing/logs:"
+
+  - it: "S=0 — agent: logs/split pipeline is absent"
+    set:
+      splunkObservability:
+        secureAppEnabled: false
+    template: configmap-agent.yaml
+    asserts:
+      - notMatchRegex:
+          path: data["relay"]
+          pattern: "logs/split:"
+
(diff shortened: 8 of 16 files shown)


rahulsplunk:
@claude recheck

.chloggen/secureapp-secapp-1072.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: enhancement
+component: agent
+note: Add `resource` processor to `logs/secureapp` pipeline for k8s host attribute enrichment (SECAPP-1072)
+issues: [2440]
+subtext: |
+  The `resource` processor is now included in the `logs/secureapp` pipeline to inject
+  `k8s.cluster.name` and `k8s.node.name` attributes required for vulnerability host mapping.
+  Also fixes a pre-existing issue where `resource/add_environment` was unconditionally
+  referenced but is only defined when `.Values.environment` is set.
```
jinja2: Can you split this into a different note with type - bug_fix?

.chloggen/secureapp-secapp-1072.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: enhancement
+component: agent
+note: Add `resource` processor to `logs/secureapp` pipeline for k8s host attribute enrichment (SECAPP-1072)
```
jinja2: A few things:
- Drop the JIRA ref from the note - we don't reference internal tickets in our repo.
- This only describes the `resource` processor, but the PR's image changes (Java image swap, Node.js `SPLUNK_SECUREAPP_AGENT_ENABLED` injection, new image version for nodejs) aren't logged. Please add changelog entries for those and state they're gated behind the non-default SecureApp toggle (except for nodejs img bump).

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -219,6 +219,13 @@ splunkObservability:
   # Enables Secure Application features
   secureAppEnabled: false
 
+  # CSA init-container image used when secureAppEnabled=true.
+  # Overrides instrumentation.spec.java.image for SecureApp instrumentation.
+  # TODO(naming): Consider whether to auto-derive this from the APM image by appending '-csa'
+  # to the repo name (e.g. splunk-otel-java:v2.27.0 -> splunk-otel-java-csa:v2.27.0).
+  # Requires agent team to confirm lockstep release contract before implementing.
+  secureAppCsaImage: ghcr.io/signalfx/splunk-otel-java/splunk-otel-java-csa:v2.28.0
```
jinja2: I'd move the image override out of `splunkObservability` and into `instrumentation`. This image config is an instrumentation concern, whereas we try to keep `splunkObservability` about export config for O11y Cloud - so it's better placed next to the other `instrumentation.spec.<lang>.image` values. It also avoids a future of `secureApp<Lang>Image` scalars if/when we need image overrides for more languages, which isn't UX-friendly.

I suggest a sibling map (helper that renders the Instrumentation CR updated accordingly):
```yaml
instrumentation:
  secureApp:
    java:
      image: ghcr.io/signalfx/splunk-otel-java/splunk-otel-java-csa:v2.28.0
  spec:
    java:
      image: ghcr.io/signalfx/splunk-otel-java/splunk-otel-java:v2.28.0
```

unittests/secureapp_test.yaml
```diff
@@ -0,0 +1,279 @@
...
+        secureAppEnabled: true
+        profilingEnabled: false
+    template: configmap-agent.yaml
+    asserts:
+      - matchRegex:
+          path: data["relay"]
+          pattern: "logs/secureapp:\\n\\s+exporters:\\n(?:\\s+- .+\\n)+\\s+processors:\\n(?:\\s+- .+\\n)+\\s+receivers:\\n\\s+- routing/logs"
+      - notMatchRegex:
+          path: data["relay"]
+          pattern: "logs/secureapp:\\n\\s+exporters:\\n(?:\\s+- .+\\n)+\\s+processors:\\n(?:\\s+- .+\\n)+\\s+receivers:\\n\\s+- otlp"
```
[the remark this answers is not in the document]
rahulsplunk: ▎ Thank you for the review! We verified this claim by rendering the actual configmap output using helm template with secureAppEnabled=true. The OTel Collector Helm chart renders all pipeline      
  ▎ blocks in block-style YAML (multi-line with dashes), not flow-style bracket lists. For example, logs/secureapp renders as:                                                                        
  ▎                                                                                                                                                                                                   
  ▎ logs/secureapp:                                                                                                                                                                                   
  ▎   exporters:                               
  ▎   - otlp_http/secureapp                                                                                                                                                                           
  ▎   processors:                                                                                                                                                                                     
  ▎   - k8s_attributes                                                                                                                                                                                
  ▎   - resourcedetection                                                                                                                                                                             
  ▎   - resource                                                                                                                                                                                      
  ▎   - memory_limiter                                                                                                                                                                                
  ▎   - batch                                                                                                                                                                                         
  ▎   receivers:                                                                                                                                                                                      
  ▎   - routing/logs                                                                                                                                                                                  
  ▎                                                                                                                                                                                                   
  ▎ The existing regex patterns match this block-style output correctly. We ran helm unittest and all 17 tests pass. No changes needed here.

unittests/secureapp_test.yaml
```diff
@@ -0,0 +1,279 @@
...
+      - notMatchRegex:
+          path: data["relay"]
+          pattern: "logs/split:\\n\\s+exporters:\\n(?:\\s+- .+\\n)+\\s+processors:\\n(?:\\s+- .+\\n)*\\s+- k8s_attributes"
+      - notMatchRegex:
+          path: data["relay"]
+          pattern: "logs/split:\\n\\s+exporters:\\n(?:\\s+- .+\\n)+\\s+processors:\\n(?:\\s+- .+\\n)*\\s+- resourcedetection"
+      - notMatchRegex:
+          path: data["relay"]
+          pattern: "logs/split:\\n\\s+exporters:\\n(?:\\s+- .+\\n)+\\s+processors:\\n(?:\\s+- .+\\n)*\\s+- resource/add_environment"
+
```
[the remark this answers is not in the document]
rahulsplunk: ▎ Thank you for the review! We verified this claim by rendering the actual configmap output using helm template with secureAppEnabled=true. The OTel Collector Helm chart renders all pipeline      
  ▎ blocks in block-style YAML (multi-line with dashes), not flow-style bracket lists. For example, logs/secureapp renders as:                                                                        
  ▎                                                                                                                                                                                                   
  ▎ logs/secureapp:                                                                                                                                                                                   
  ▎   exporters:                               
  ▎   - otlp_http/secureapp                                                                                                                                                                           
  ▎   processors:                                                                                                                                                                                     
  ▎   - k8s_attributes                                                                                                                                                                                
  ▎   - resourcedetection                                                                                                                                                                             
  ▎   - resource                                                                                                                                                                                      
  ▎   - memory_limiter                                                                                                                                                                                
  ▎   - batch                                                                                                                                                                                         
  ▎   receivers:                                                                                                                                                                                      
  ▎   - routing/logs                                                                                                                                                                                  
  ▎                                                                                                                                                                                                   
  ▎ The existing regex patterns match this block-style output correctly. We ran helm unittest and all 17 tests pass. No changes needed here.

unittests/secureapp_test.yaml
```diff
@@ -0,0 +1,279 @@
...
+        realm: us1
+    template: configmap-agent.yaml
+    asserts:
+      - matchRegex:
+          path: data["relay"]
+          pattern: "otlp_http/secureapp:\\n\\s+headers:\\n(?:\\s+.+\\n)+\\s+logs_endpoint: https://ingest\\.us1\\."
+      - matchRegex:
+          path: data["relay"]
+          pattern: "X-Splunk-Instrumentation-Library: secureapp"
+
```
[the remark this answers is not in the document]
rahulsplunk: ▎ Thank you for the review! We verified this claim by rendering the actual configmap output using helm template with secureAppEnabled=true. The OTel Collector Helm chart renders all pipeline      
  ▎ blocks in block-style YAML (multi-line with dashes), not flow-style bracket lists. For example, logs/secureapp renders as:                                                                        
  ▎                                                                                                                                                                                                   
  ▎ logs/secureapp:                                                                                                                                                                                   
  ▎   exporters:                               
  ▎   - otlp_http/secureapp                                                                                                                                                                           
  ▎   processors:                                                                                                                                                                                     
  ▎   - k8s_attributes                                                                                                                                                                                
  ▎   - resourcedetection                                                                                                                                                                             
  ▎   - resource                                                                                                                                                                                      
  ▎   - memory_limiter                                                                                                                                                                                
  ▎   - batch                                                                                                                                                                                         
  ▎   receivers:                                                                                                                                                                                      
  ▎   - routing/logs                                                                                                                                                                                  
  ▎                                                                                                                                                                                                   
  ▎ The existing regex patterns match this block-style output correctly. We ran helm unittest and all 17 tests pass. No changes needed here.

unittests/secureapp_test.yaml
```diff
@@ -0,0 +1,279 @@
...
+      splunkPlatform:
+        endpoint: https://hec.example.com
+        token: test-hec-token
+        logsEnabled: true
+    template: configmap-agent.yaml
+    asserts:
+      - matchRegex:
+          path: data["relay"]
+          pattern: "routing/logs:\\n\\s+default_pipelines:\\n\\s+- logs\\n"
+
```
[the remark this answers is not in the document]
rahulsplunk: ▎ Thank you for the review! We verified this claim by rendering the actual configmap output using helm template with secureAppEnabled=true. The OTel Collector Helm chart renders all pipeline      
  ▎ blocks in block-style YAML (multi-line with dashes), not flow-style bracket lists. For example, logs/secureapp renders as:                                                                        
  ▎                                                                                                                                                                                                   
  ▎ logs/secureapp:                                                                                                                                                                                   
  ▎   exporters:                               
  ▎   - otlp_http/secureapp                                                                                                                                                                           
  ▎   processors:                                                                                                                                                                                     
  ▎   - k8s_attributes                                                                                                                                                                                
  ▎   - resourcedetection                                                                                                                                                                             
  ▎   - resource                                                                                                                                                                                      
  ▎   - memory_limiter                                                                                                                                                                                
  ▎   - batch                                                                                                                                                                                         
  ▎   receivers:                                                                                                                                                                                      
  ▎   - routing/logs                                                                                                                                                                                  
  ▎                                                                                                                                                                                                   
  ▎ The existing regex patterns match this block-style output correctly. We ran helm unittest and all 17 tests pass. No changes needed here.

unittests/secureapp_test.yaml
```diff
@@ -0,0 +1,279 @@
...
+    set:
+      splunkObservability:
+        secureAppEnabled: true
+        profilingEnabled: true
+    template: configmap-agent.yaml
+    asserts:
+      - matchRegex:
+          path: data["relay"]
+          pattern: "routing/logs:\\n\\s+default_pipelines:\\n\\s+- logs\\n"
+
```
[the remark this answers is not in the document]
rahulsplunk: ▎ Thank you for the review! We verified this claim by rendering the actual configmap output using helm template with secureAppEnabled=true. The OTel Collector Helm chart renders all pipeline      
  ▎ blocks in block-style YAML (multi-line with dashes), not flow-style bracket lists. For example, logs/secureapp renders as:                                                                        
  ▎                                                                                                                                                                                                   
  ▎ logs/secureapp:                                                                                                                                                                                   
  ▎   exporters:                               
  ▎   - otlp_http/secureapp                                                                                                                                                                           
  ▎   processors:                                                                                                                                                                                     
  ▎   - k8s_attributes                                                                                                                                                                                
  ▎   - resourcedetection                                                                                                                                                                             
  ▎   - resource                                                                                                                                                                                      
  ▎   - memory_limiter                                                                                                                                                                                
  ▎   - batch                                                                                                                                                                                         
  ▎   receivers:                                                                                                                                                                                      
  ▎   - routing/logs                                                                                                                                                                                  
  ▎                                                                                                                                                                                                   
  ▎ The existing regex patterns match this block-style output correctly. We ran helm unittest and all 17 tests pass. No changes needed here.

unittests/secureapp_test.yaml
```diff
@@ -0,0 +1,279 @@
...
+      splunkPlatform:
+        endpoint: https://hec.example.com
+        token: test-hec-token
+        logsEnabled: true
+    template: configmap-agent.yaml
+    asserts:
+      - matchRegex:
+          path: data["relay"]
+          pattern: "routing/logs:\\n\\s+default_pipelines:\\n\\s+- logs\\n"
+      - matchRegex:
```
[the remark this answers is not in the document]
rahulsplunk: ▎ Thank you for the review! We verified this claim by rendering the actual configmap output using helm template with secureAppEnabled=true. The OTel Collector Helm chart renders all pipeline      
  ▎ blocks in block-style YAML (multi-line with dashes), not flow-style bracket lists. For example, logs/secureapp renders as:                                                                        
  ▎                                                                                                                                                                                                   
  ▎ logs/secureapp:                                                                                                                                                                                   
  ▎   exporters:                               
  ▎   - otlp_http/secureapp                                                                                                                                                                           
  ▎   processors:                                                                                                                                                                                     
  ▎   - k8s_attributes                                                                                                                                                                                
  ▎   - resourcedetection                                                                                                                                                                             
  ▎   - resource                                                                                                                                                                                      
  ▎   - memory_limiter                                                                                                                                                                                
  ▎   - batch                                                                                                                                                                                         
  ▎   receivers:                                                                                                                                                                                      
  ▎   - routing/logs                                                                                                                                                                                  
  ▎                                                                                                                                                                                                   
  ▎ The existing regex patterns match this block-style output correctly. We ran helm unittest and all 17 tests pass. No changes needed here.

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -250,6 +250,10 @@
           "description": "Whether Secure Application features are enabled for Splunk Observability",
           "type": "boolean"
         },
+        "secureAppCsaImage": {
```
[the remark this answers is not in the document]
rahulsplunk: ▎ Great catch, thank you! You're absolutely right — the existing anyOf only allows metricsEnabled, tracesEnabled, or profilingEnabled to satisfy the constraint, which would cause helm schema 
  ▎ validation to reject a pure SecureApp deployment (secureAppEnabled=true with all other signals disabled). We'll add secureAppEnabled as a fourth option in the anyOf to fix this.

.chloggen/secureapp-secapp-1072.yaml
rahulsplunk: ▎ Good points, thank you. We'll:
  ▎ 1. Remove the internal JIRA reference from the note                                                                                                                                               
  ▎ 2. Split the entry into separate enhancement (resource processor) and bug_fix (resource/add_environment guard) entries                                                                            
  ▎ 3. Add separate changelog entries for the operator image changes (Java CSA image swap and Node.js SPLUNK_SECUREAPP_AGENT_ENABLED injection), noting they are gated behind secureAppEnabled        
  ▎ 4. Add a separate entry for the Node.js instrumentation image version bump (which is unconditional)

.chloggen/secureapp-secapp-1072.yaml
rahulsplunk: ▎ Good points, thank you. We'll:
  ▎ 1. Remove the internal JIRA reference from the note                                                                                                                                               
  ▎ 2. Split the entry into separate enhancement (resource processor) and bug_fix (resource/add_environment guard) entries                                                                            
  ▎ 3. Add separate changelog entries for the operator image changes (Java CSA image swap and Node.js SPLUNK_SECUREAPP_AGENT_ENABLED injection), noting they are gated behind secureAppEnabled        
  ▎ 4. Add a separate entry for the Node.js instrumentation image version bump (which is unconditional)

helm-charts/splunk-otel-collector/values.yaml
rahulsplunk: ▎ Thank you for the thoughtful suggestion — the separation of concerns makes sense and we're happy to move it. Since we're relatively new to this repo, could you help guide us on the preferred 
  ▎ location? Specifically:                                                                                                                                                                           
  ▎                                                                               
  ▎ - Should it live at instrumentation.spec.java.secureAppImage alongside the standard instrumentation.spec.java.image?                                                                              
  ▎ - Or would you prefer a separate sub-map, e.g. instrumentation.secureApp.java.image?
  ▎                                                                                                                                                                                                   
  ▎ We want to follow the existing conventions here rather than introduce something inconsistent. Once you point us to the right path we'll update values.yaml, values.schema.json, and the template  
  ▎ reference in _helpers.tpl. Happy to iterate quickly on this.

jinja2: I have suggested my preferred config shape in my original comment. Lmk if you have any other concerns

dmitryax:
@rahmathu please stop posting agent responses here. It takes more time from the reviewers

helm-charts/splunk-otel-collector/values.yaml
dmitryax: I agree, I don't think we need to expose any new configuration interface. The secureapp auto instrumentation should be triggered by existing `splunkObservability.secureAppEnabled`+`instrumentation.enabled`. If the image needs to be changed additionally, it can be done under `instrumentation` as Jina proposed

rahulsplunk: thank you guys for the details on this , while experimenting i saw these things . I wanted to check with you guys if i take your suggestion what can i do here to ensure the approach is solid . Sorry if this is getting a bit repetitive, i am bit new to this repo and working here 

The problem i see is the default value should move it 
  The template at _helpers.tpl:134 does:                                                                                                                                                                                              
`  {{- $libSpec = merge (dict "image" $.Values.instrumentation.secureApp.java.image) $libSpec -}}`

If `instrumentation.secureApp` is not set in a user's values, Helm returns nil for the nested key. merge (dict "image" nil) produces image: null in the rendered Instrumentation CR, which Kubernetes rejects — the instrumented pod fails to start.

This means the default value in `values.yaml` must also move from `splunkObservability.secureAppCsaImage` to `instrumentation.secureApp.java.image`. As long as the default is set there, users who don't override anything are safe. Users who were overriding `splunkObservability.secureAppCsaImage` would need to update their values — but since this feature hasn't shipped in any release yet, there are no existing users to break.

small table to capture my cases 
```bash
  ┌────────────────────────────────┬───────────────────────────────────────────────┐
  │              Mode              │                  What breaks                  │                                                                                                          
  ├────────────────────────────────┼───────────────────────────────────────────────┤                                                                                                          
  │ SecureApp enabled, any signals │ image: null in Instrumentation CR → pod fails │                                                                                                          
  ├────────────────────────────────┼───────────────────────────────────────────────┤                                                                                                          
  │ Pure SecureApp                 │ Same                                          │                                                                                                          
  ├────────────────────────────────┼───────────────────────────────────────────────┤                                                                                                          
  │ Gateway mode + SecureApp       │ Same                                          │                                                                                                          
  ├────────────────────────────────┼───────────────────────────────────────────────┤                                                                                                          
  │ SecureApp disabled             │ Safe (code path never executes)               │
  ├────────────────────────────────┼───────────────────────────────────────────────┤                                                                                                          
  │ instrumentation.enabled: false │ Safe (template skipped entirely)              │
  └────────────────────────────────┴───────────────────────────────────────────────┘
```
  Proposed implementation:                                      
                                                                                                                                                                                              
  values.yaml:
```yaml
  instrumentation:                                                                                                                                                                            
    secureApp:                                                                                                                                                                                
      java:                                                                                                                                                                                   
        image: ghcr.io/signalfx/splunk-otel-java/splunk-otel-java-csa:v2.28.0                                                                                                                 
    spec:                                                                    
      java:                                                                                                                                                                                   
        image: ...  # existing                                                                                                                                                                
 ```                                                                                                                                                                                             
  _helpers.tpl:134 — update reference:                                                                                                                                                        
  `{{- $libSpec = merge (dict "image" $.Values.instrumentation.secureApp.java.image) $libSpec -}}`                                                                                       
                                                                                                                                                                                              
values.schema.json — add secureApp to the instrumentation schema properties.                                                                                                                                                                                                                                                                                                           splunkObservability.secureAppCsaImage — remove entirely from values.yaml and schema. 

I wanted to make sure that i am not missing anything about how `instrumentation.secureApp` should behave when instrumentation.enabled=false — should accessing `$.Values.instrumentation.secureApp.java.image` still be valid in that case (it's gated by secureAppEnabled in the template, so it should be fine), or should we add an additional nil-guard?

Comment (rahulsplunk) [reply helm-charts/splunk-otel-collector/values.yaml]: @dmitryax  or @jinja2 can you please help close on these questions ?

Comment (dmitryax) [reply helm-charts/splunk-otel-collector/values.yaml]: We don't need `instrumentation.secureApp`

Comment (dmitryax) [reply helm-charts/splunk-otel-collector/values.yaml]: `instrumentation.secureApp.java.image` should have different default values based on `splunkObservability.secureAppEnabled`

Comment (rahulsplunk) [reply helm-charts/splunk-otel-collector/values.yaml]: @dmitryax  : 
1. to confirm: should the CSA image be hardcoded in the template (no user override), 
or 
2. should i expose a default at instrumentation.secureApp.java.image that users can override? Your 
last two comments confused me a bit from what was mentioned above from @jinja2 and you 

Before finalizing the implementation can you pl clarify exactly what seems right to you

Comment (dmitryax) [reply helm-charts/splunk-otel-collector/values.yaml]: I think we can avoid introducing `instrumentation.secureApp` completely and have make the `instrumentation.secureApp.java.image`'s default dynamic based on splunkObservability.secureAppEnabled
```
  instrumentation:                                                                                                                                                                            
    spec:                                                                                                                                                                                
      java:
        # The commented value is the default
        # However, it's changed to ghcr.io/signalfx/splunk-otel-java/splunk-otel-java-csa:v2.28.0 if splunkObservability.secureAppEnabled=true
        # image: ghcr.io/signalfx/splunk-otel-java/splunk-otel-java:v2.28.0
```


Another option is to define secureapp image separately but still under `instrumentation.spec.java`. Maybe `instrumentation.spec.java.saImage`

File: helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
Comment (dmitryax) [inline helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl:1]: Why this file needs to be changed at all? Seems unrelated

File: helm-charts/splunk-otel-collector/values.yaml
Comment (dmitryax) [inline helm-charts/splunk-otel-collector/values.yaml:1159]: Please don't make any unrelated changes

File: unittests/resource_detection_test.yaml [hunk 11]
```diff
@@ -568,15 +568,15 @@ tests:
           path: data["relay"]
           pattern: "logs/k8s_entities:\\n\\s+exporters:\\n(?:\\s+- .+\\n)+\\s+processors:\\n(?:\\s+- .+\\n)*\\s+- resourcedetection/k8s_cluster_name\\n(?:\\s+- .+\\n)*\\s+receivers:"
 
-  - it: "agent — logs/secureapp pipeline has no resourcedetection"
+  - it: "agent — logs/secureapp pipeline has resourcedetection processor"
```
dmitryax, line 571: Why?

helm-charts/splunk-otel-collector/values.yaml
rahulsplunk: thanks i will take option that you suggested first , please note the following:  

A user who wants to pin a specific CSA image version (e.g. for rollback during an incident) has no override path — they'd have to set instrumentation.spec.java.image directly, but that would be overridden by the template when secureAppEnabled=true. This is the tradeoff here

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
rahulsplunk: @dmitryax : I understand your concerns here, my objective here is 
1. to get the changes in for Java, NodeJS and Python SecureApp agents 
2. The pipeline changes are for router and exports the vulnerability events in the instrumented pods emit, the secure app requirement add the resource processor the host attribute enrichment 
3. without the collector pipeline changes , the events from the instrumented pods have no where to go . The 2 halves are deployed together and tightly coupled 

I clubbed the changes into one so we are just waiting on 1 PR . If you think this is not correct then i can raise another PR. Please point out the exact changes you would like separated apart from these mentioned

helm-charts/splunk-otel-collector/values.yaml
rahulsplunk: i would say the same reason as above

helm-charts/splunk-otel-collector/values.yaml
jinja2: @rahmathu Will this image be updated regularly, similar to splunk-otel-java image's update cadence? We have [workflows](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/.github/workflows/update_docker_images.yaml) in our repo which opens PRs to update image tags for auto-instrumentation and we should look into onboarding this image to the workflow too as a follow-up so it doesn't diverge from the base splunk-otel-java image.

@dmitryax I don't think we should hardcode the CSA image as a literal in tpl. Our [update_docker_images.yaml](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/.github/workflows/update_docker_images.yaml#L29) workflow auto-bumps instrumentation images by reading the tag from a declared YAML path and opening a PR. If we go the literal way, not only will a secureApp user be unable to update the tag through values, we also won't be able to onboard this image to our existing auto-upgrade workflow.

`instrumentation.spec.java.secureAppImage` sgtm if we want to avoid the `secureApp` sub-tree under `instrumentation.spec`, but the helper would have to unset it before rendering or it leaks as an invalid CR field - I'd rather keep the clean separation with the map under instrumentation, especially if we expect these for other languages. Wdyt?

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -1554,6 +1554,11 @@
       "description": "Helm Chart Feature Gates.",
       "type": "object",
       "additionalProperties": true
+    },
```
rahulsplunk, line 1557: @dmitryax  and @jinja2  : PLease note while updating the latest master a schema gap was seen which was causing  test failures . This is the reason for this change , i was trying to make sure this is getting fixed

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: The logs pipeline for secureapp events already exists. 

> add the resource processor the host attribute enrichment

This line seems like the only actual change. It should be a separate PR and it would require only 2 extra processors added (2 lines). It doesn't need whole template refactor.

examples/enable-operator-and-auto-instrumentation/rendered_manifests/operator/admission-webhooks/operator-webhook.yaml
```diff
@@ -1,5 +1,11 @@
 ---
 # Source: splunk-otel-collector/charts/operator/templates/admission-webhooks/operator-webhook.yaml
+---
+---
+# Source: splunk-otel-collector/charts/operator/templates/admission-webhooks/operator-webhook.yaml
+---
+---
+# Source: splunk-otel-collector/charts/operator/templates/admission-webhooks/operator-webhook.yaml
```
jinja2, line 8: These changes should not be happening and might be the cause of pre-commit failures. Are you using helm 4 when running `make render`?

.chloggen/secureapp-agent-resource-processor.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: agent
+note: Add `resource` processor to `logs/secureapp` pipeline to inject `k8s.cluster.name` and `k8s.node.name` for vulnerability host mapping
```
jinja2, line 3: From the changes, this does not seem to be the only the processor introduced in secureApp. There is also k8s_attr and resourcedetection. Can you clarify what attributes you are hoping to actually add with processors?

helm-charts/splunk-otel-collector/values.yaml
dmitryax: Sounds good. Let's keep it simpler with `instrumentation.spec.java.secureAppImage`

rahulsplunk: as per discussion 
1. we want to remove secureaApp from under the instrumentation and it should be placed under the spec under  java . So it will be like AI @rahulsplunk 

```
  instrumentation:                                                                                                                                                                            
    spec:                                                                    
      java:                                                                                                                                                                                   
        image: ...  # existing                                                                                                                                                                
        secureAppImage: ghcr.io/signalfx/splunk-otel-java/splunk-otel-java-csa:v2.28.0                                                                                                                                                                                                                                                                                                
```

update the image 
1. create a separate PR for the workflow to updated the release image, AI : @rahulsplunk 
3. add test for CSA specific auto instrumentation , separate PR for that AI : @rahulsplunk

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
rahulsplunk: @rahulsplunk  : create a separate PR for the logs where secureApp is not getting regular logs and only the ones it needs .

examples/enable-operator-and-auto-instrumentation/rendered_manifests/operator/admission-webhooks/operator-webhook.yaml
rahulsplunk: @rahulsplunk : undo this file change . It will fix CI

helm-charts/splunk-otel-collector/values.schema.json
rahulsplunk: @rahulsplunk : Revert this change and see if CI fails

rahulsplunk:
PR's 
1. Auto instrumentation only 
2. Logs big fix - all logs are sent to secure app , correct this 
3. enhancement - logs pipeline enhancement to add more attributes 
4. Workflow - where secureapp image reelase with otel agent takes place 
5. Integration test addition for SecureApp , this adds on top on otel IT test 

repo changes : https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/auto-instrumentation-install.md

PR #2441, opened 2026-06-02
feat: mounting secret as a file instead of using env variables

dhruvanshus-crest:
**Description:** mounting secret as a file instead of using env variables
 Mounting secret as a file instead of env variables for both Splunk HEC token and Splunk observability access token to fix the security vulnerability - [SITG](https://www.stigviewer.com/stigs/kubernetes/2025-05-16/finding/V-242415)

**Link to ADDON Jira:** [ADDON-87370](https://splunk.atlassian.net/browse/ADDON-87370)


dhruvanshus-crest:
> Can we move this to workload's podSecurityContext setting instead? gateway.podSecurityContext/clusterReceiver.podSecurityContext already exist and are the conventional home for fsGroup. secret.fsGroup reads as scoped to just the secret but it is pod-wide and every fsGroup-eligible volume in the pod is affected. The agent only has a container-level securityContext, which can't carry fsGroup, so it needs a new agent.podSecurityContext (this also makes the agent consistent with the other workloads).

@jinja2 I've moved this to podSecurityContext, removed secret.fsGroup and added new agent.podSecurityContext.

dhruvanshus-crest:
> We have a set of tests that run the chart in hosted clusters, but a fork PR won't trigger them. I'd prefer these changes run against those tests; we should look into what's needed internally for proper access.
> 
> A few things these tests may answer (we have an AKS cluster with Windows nodes and an OpenShift cluster wired into CI) - or, if you've already looked into them, please comment:
> 
> Windows - I understand fsGroup isn't supported on Windows. Have we verified that ${file:/otel/etc/...} resolves on Windows and is readable by the collector process?
> OpenShift - OpenShift admits every pod through a [SecurityContextConstraints](https://docs.redhat.com/en/documentation/openshift_container_platform/4.11/html/authentication_and_authorization/managing-pod-security-policies) (SCC). The default SCC uses fsGroup: MustRunAs with a per-namespace allocated GID range, and OpenShift normally auto-assigns one from that range. The [SCC our chart creates allows any fsGroup](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/securityContextConstraints.yaml#L28-L29), but that's configurable. I want to confirm whether GID 999 falls within the default OpenShift SCC's allowed range. At a minimum we should document this for OpenShift users.

I've closed the previous fork PR as the test were not executed, the tests are executed on this PR for both windows and Openshift and are passing.

GID 999 falls within the default Openshift SCC's allowed range, unable to verify this information but I've documented this for Openshift users. The changes can be found in the PR.

cc: @jinja2

examples/distribution-gke-autopilot/rendered_manifests/daemonset.yaml
```diff
@@ -156,6 +158,9 @@ spec:
         - mountPath: /hostfs/var/run/utmp
           name: host-var-run-utmp
           readOnly: true
+        - name: secret
+          mountPath: /otel/etc
+          readOnly: true
```
jinja2: Looking at the exact fields matched as part of [workload allowlisting](https://docs.cloud.google.com/kubernetes-engine/docs/reference/crds/workloadallowlist) - this new volumeMount seems to be rejected with warden on gke-autopilot. We'll have to get this allowlisted before we can make this change for autopilot. Let me check internally on the process for this and if we should skip for autopilot for now.

PS - from the allowlist doc, the other 2 changes look okay
1. Pod securityContext matches only appArmorProfile - https://docs.cloud.google.com/kubernetes-engine/docs/reference/crds/workloadallowlist#PodSecurityContext
2. Volumes does not call out the `secret` field - https://docs.cloud.google.com/kubernetes-engine/docs/reference/crds/workloadallowlist#volumes

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -448,6 +450,12 @@ agent:
   # Please note that on uninstallation of the chart, the permissions added to the
   # host log directories for given uid/gid are not reverted.
 
+  # Pod security context applied to the agent daemonset. On Linux, the chart
+  # injects fsGroup: 999 when mounting group-readable Splunk Secret files unless
+  # fsGroup is set here. OpenShift users should set a value allowed by their SCC
+  # if the chart-created SCC is disabled or overridden.
+  podSecurityContext: {}
```
jinja2: Any reason we can't directly set `fsGroup: 999` under `agent.podSecurityContext` / `gateway.podSecurityContext` / `clusterReceiver.podSecurityContext`? We can strip it for windows with the securityContext helper. I think having it in values yaml is better for ux/discoverability and the chart templating will also simplify.

Can you check if there any drawbacks to having it set even when the secret is not mounted? I think hostVolume's don't get fsgroup chown but better to verify this. If a user does complain about pod startup latency due to fsgroup related chown on their extra volumes (emptyDir might be the only one in our chart), setting `fsGroupChangePolicy: OnRootMismatch` should help and we can document this.

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
```diff
@@ -22,7 +22,7 @@ extensions:
       - action: upsert
         key: X-SF-TOKEN
         from_context: X-SF-TOKEN
-        default_value: "${SPLUNK_OBSERVABILITY_ACCESS_TOKEN}"
+        default_value: "${file:/otel/etc/splunk_observability_access_token}"
```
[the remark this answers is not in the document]
jinja2: This looks like a legit concern.

docs/advanced-configuration.md
```diff
@@ -99,6 +99,35 @@ secret:
...
+gateway:
+  podSecurityContext:
+    fsGroup: 1234
+```
+
+On OpenShift, explicit `fsGroup` values are validated by Security Context
+Constraints (SCCs). The chart-created SCC allows any `fsGroup` by default, but
+if you disable or override that SCC, make sure the configured `fsGroup` is
+allowed by the SCC used to admit the pod.
+```
```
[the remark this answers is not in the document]
dhruvanshus-crest: done

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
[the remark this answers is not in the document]
dhruvanshus-crest: changes done

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -459,7 +459,7 @@ Splunk Platform Logs exporter
 {{- define "splunk-otel-collector.splunkPlatformLogsExporter" -}}
 splunk_hec/platform_logs:
   endpoint: {{ .Values.splunkPlatform.endpoint | quote }}
-  token: "${SPLUNK_PLATFORM_HEC_TOKEN}"
+  token: "${file:/otel/etc/splunk_platform_hec_token}"
   index: {{ .Values.splunkPlatform.index | quote }}
```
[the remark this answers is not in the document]
dhruvanshus-crest: I think this can be ignored as it is the same way implemented what it is for cert_file, ca_file. I checked mountPaths are fine hostPaths need to follow C standard, so we are good

helm-charts/splunk-otel-collector/values.yaml
dhruvanshus-crest: - I didn't set it directly because I kept it conditional based on secret mount required. But I think we can set it directly. 
- You are correct regarding the setting fsGroup directly if secret is not mounted, there can be startup latency for on extra Volumes and documented the same how to mitigate it.
- Verified it based on kubernetes doc - https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods
- Applied the requested changes

examples/distribution-gke-autopilot/rendered_manifests/daemonset.yaml
dhruvanshus-crest: okay, please keep me updated on the outcome to skip for GKE-autopilot for now or not?

dmitryax (approved):
LGTM

dmitryax:
@dhruvanshus-crest sorry for the delay. can you please rebase?

docs/advanced-configuration.md
```diff
@@ -99,6 +99,52 @@ secret:
   name: your-secret
 ```
 
+Files mounted from the Splunk Secret use `secret.defaultMode: "0440"` by
```
dmitryax, line 107: The wording around this section still says like this is an optional approach, please update that

dmitryax:
@dhruvanshus-crest can you also please exclude this change from GKE/Autopilot for now. We need some extra steps to unblock it on that platform which take some time

docs/advanced-configuration.md
dhruvanshus-crest: done

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -25,6 +25,47 @@ If release name contains chart name it will be used as a full name.
...
 {{- end -}}
 
+{{/*
+Warn if a collector config override still references Splunk token environment variables.
+*/}}
+{{- define "splunk-otel-collector.warnOnTokenEnvVarRefs" -}}
+{{- if and .enabled .config -}}
+{{- $source := .source -}}
+{{- if regexMatch "\\$\\{SPLUNK_[A-Z0-9_]*_TOKEN\\}" (toYaml .config) }}
+{{- printf "[WARNING] %s references a Splunk token environment variable (${SPLUNK_*_TOKEN}). Built-in chart configuration now reads tokens from mounted Secret files. Please update custom collector config to use ${file:/otel/etc/splunk_observability_access_token} or ${file:/otel/etc/splunk_platform_hec_token}. Token environment variables are still injected for compatibility but will be removed in a future release.\n" $source }}
```
jinja2: Since we are using a featuregate now which a user will opt into, lets stop injecting these environment variables. Custom configurations that still reference them should produce render-time failure instead of warning.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -1452,6 +1475,9 @@ featureGates:
   # and target allocator deployments.
   # AKS clusters with Azure Policy enabled might need to enable this to resolve automountServiceAccountToken policy.
   explicitMountServiceAccountToken: false
+  # Mount Splunk Observability and Splunk Platform HEC tokens from the Splunk
+  # Secret and configure the Collector to read them from files instead of environment variables
```
jinja2:
```suggestion
  # Secret and configure the default Collector configuration to read them from files.
  # Token environment variables are not created when this is enabled. Custom
  # Collector configuration must replace:
  #   ${SPLUNK_OBSERVABILITY_ACCESS_TOKEN} with ${file:/otel/etc/splunk_observability_access_token}
  #   ${SPLUNK_PLATFORM_HEC_TOKEN} with ${file:/otel/etc/splunk_platform_hec_token}
  # This option is not currently supported on GKE Autopilot.
```

Verify the env var/file paths in the suggestion, in case i have missed anything.

Please add a note about gke-autopilot support in docs and changelog too. Maybe add a schema check like below (not tested) - 
```
{
  "if": {
    "properties": {
      "distribution": {
        "const": "gke/autopilot"
      }
    },
    "required": ["distribution"]
  },
  "then": {
    "properties": {
      "featureGates": {
        "properties": {
          "mountSplunkSecretAsFile": {
            "const": false
          }
        }
      }
    }
  }
}
```

docs/advanced-configuration.md
```diff
@@ -86,19 +86,75 @@ the chart stores them in the Splunk Platform Secret, mounts that Secret at
...
 
 ```yaml
 secret:
   create: false
   name: your-secret
 ```
 
+Avoid using this pre-created Secret option for new deployments. Prefer the
+default chart-created Secret because the pre-created Secret option will be
+unsupported in a future release.
```
jinja2: Please remove this wording about pre-create secrets. Secrets created externally from helm release is a legit and often preferred k8s administrative model. It should be documented as supported and tested to make sure PR changes are compatible.

.chloggen/mount-secret-as-file.yaml
```diff
@@ -0,0 +1,5 @@
+change_type: enhancement
+component: chart
+note: Mount Splunk tokens as files and use file references in the default collector configuration.
+issues: [2425]
```
jinja2:
```suggestion
issues: [2441]
```

This should be PR number so user can look up the full change from changelog

.chloggen/mount-secret-as-file.yaml
```diff
@@ -0,0 +1,5 @@
+change_type: enhancement
```
jinja2, line 1: Can you add another changelog entry with a note like - `Set mounted Splunk Secret files to mode `0440` by default and set `fsGroup: 999` on Linux collector pods. Both settings remain configurable.`? This will separate the defatult changes from the featuregate

.chloggen/mount-secret-as-file.yaml
```diff
@@ -0,0 +1,5 @@
...
+component: chart
+note: Mount Splunk tokens as files and use file references in the default collector configuration.
```
jinja2:
```suggestion
note: Add an opt-in setting to read Splunk tokens from mounted Secret files.
```

.chloggen/mount-secret-as-file.yaml
```diff
@@ -0,0 +1,5 @@
...
+issues: [2425]
+subtext:
```
jinja2:
```suggestion
subtext:  |
  Enable `featureGates.mountSplunkSecretAsFile` to use this mode with either a chart-created or existing Secret. This option is disabled by default and is not currently supported on GKE Autopilot.
```

jinja2 (changes requested):
@dhruvanshus-crest thanks for the updates. Can you update the hosted functional_test's [values files for different k8s distros](https://github.com/signalfx/splunk-otel-collector-chart/tree/main/functional_tests/functional/testdata/values) to have the feature gate enabled so we can test the latest in these clusters? Especially, the AKS windows cluster to make sure we don't need the file path to have anything windows specific

helm-charts/splunk-otel-collector/values.yaml
dhruvanshus-crest: Applied the suggestion. Added a note about gke-autopilot support in docs and changelog too. Also, added a schema check with unit test for the same.

docs/advanced-configuration.md
dhruvanshus-crest: Updated the wording

.chloggen/mount-secret-as-file.yaml
dhruvanshus-crest: Yes, it was my previous PR number the fork one but thanks for spotting, I've updated it now.

.chloggen/mount-secret-as-file.yaml
dhruvanshus-crest: done

.chloggen/mount-secret-as-file.yaml
dhruvanshus-crest: updated

.chloggen/mount-secret-as-file.yaml
dhruvanshus-crest: updated

helm-charts/splunk-otel-collector/templates/_helpers.tpl
dhruvanshus-crest: yeah it make sense. I've updated the logic.

dhruvanshus-crest:
> @dhruvanshus-crest thanks for the updates. Can you update the hosted functional_test's [values files for different k8s distros](https://github.com/signalfx/splunk-otel-collector-chart/tree/main/functional_tests/functional/testdata/values) to have the feature gate enabled so we can test the latest in these clusters? Especially, the AKS windows cluster to make sure we don't need the file path to have anything windows specific

I've updated the hosted functional_test's values files for different distribution with feature gate enabled. I skipped or kept it same for distribution `eks/auto-mode` and `eks/fargate` so that we can also be sure the legacy (secret as env variable) is working as expected. Let me know if that makes sense or require any changes.

jinja2 (approved):

PR #2442, opened 2026-06-02
[chart] Disable k8s.node.condition_ready in clusterReceiver

rohit-sonawane:
Set `node_conditions_to_report: []` in the default `k8s_cluster` receiver config to stop emitting the legacy `k8s.node.condition_ready` metric. The `k8s.node.condition` metric (enabled in #2151) already covers node conditions via a `condition` attribute, making `k8s.node.condition_ready` redundant and adding avoidable MTS/cardinality.

Updates expected test data and all rendered example manifests.

Resolves OTL-4324

**Description:** Before this change, the collector emitted two overlapping metrics for the node `Ready` condition — the legacy `k8s.node.condition_ready` (standalone gauge, no attributes) alongside the newer `k8s.node.condition` (attribute-based gauge covering all conditions). This PR disables the legacy metric by setting `node_conditions_to_report: []` in the default cluster receiver config, eliminating the duplicate MTS while keeping full node condition coverage via `k8s.node.condition{condition="Ready|MemoryPressure|DiskPressure|PIDPressure"}`.

**Link to Splunk idea:** N/A

**Testing:** Verified locally with an E2E test on minikube using a `debug` exporter to inspect emitted metrics directly from the collector logs.

- `k8s.node.condition` IS emitted with the `condition` attribute covering all node conditions:

  ```
  Name: k8s.node.condition
    condition=MemoryPressure  →  Value: 0
    condition=DiskPressure    →  Value: 0
    condition=PIDPressure     →  Value: 0
    condition=Ready           →  Value: 1
  ```

- `k8s.node.condition_ready` is **NOT** emitted (grep returns empty):

  ```bash
  $ kubectl logs <cluster-receiver-pod> | grep "condition_ready"
  # (no output)
  ```

- Rendered ConfigMap confirms the setting:

  ```yaml
  k8s_cluster:
    node_conditions_to_report: []   # disables legacy metric
    metrics:
      k8s.node.condition:
        enabled: true               # new metric still active
  ```

**Documentation:** No user-facing documentation changes needed. The `k8s.node.condition` metric was already documented in #2151. The removal of `k8s.node.condition_ready` is captured in `.chloggen/disable-k8s-node-condition-ready.yaml`.

### DIFF
--- a/functional_tests/k8sentities/testdata/expected_k8sentities.yaml
+++ b/functional_tests/k8sentities/testdata/expected_k8sentities.yaml
@@ -864,7 +864,9 @@ resourceLogs:
                       - key: k8s.node.condition_pid_pressure
                         value:
                           stringValue: "false"
-
+                      - key: k8s.node.condition_ready
+                        value:
+                          stringValue: "true"
                       - key: beta.kubernetes.io/arch
                         value:
                           stringValue: arm64
--- a/functional_tests/k8sentities/testdata/expected_k8sentities.yaml
+++ b/functional_tests/k8sentities/testdata/expected_k8sentities.yaml
@@ -864,9 +864,7 @@ resourceLogs:
                       - key: k8s.node.condition_pid_pressure
                         value:
                           stringValue: "false"
-                      - key: k8s.node.condition_ready
-                        value:
-                          stringValue: "true"
+
                       - key: beta.kubernetes.io/arch
                         value:
                           stringValue: arm64
--- a/.chloggen/disable-k8s-node-condition-ready.yaml
+++ b/.chloggen/disable-k8s-node-condition-ready.yaml
@@ -1,0 +1,15 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: clusterReceiver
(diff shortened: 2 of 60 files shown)


functional_tests/k8sentities/testdata/expected_k8sentities.yaml
```diff
@@ -864,9 +864,7 @@ resourceLogs:
                       - key: k8s.node.condition_pid_pressure
                         value:
                           stringValue: "false"
-                      - key: k8s.node.condition_ready
-                        value:
-                          stringValue: "true"
+
```
jinja2: I don't think this attribute on the `k8s.node` entity would be dropped with the config change. Can you confirm if this is correct? The [k8sentities test](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/functional_tests/k8sentities/k8sentities_test.go) does not seem to be checking attributes within the entity so the test won't fail atm if this is not the case. We should update the test to compare attr keys too; just making a note and not related to this PR.

rohit-sonawane: You're right — the k8s.node.condition_ready entity attribute is emitted by GetMetadata() in the k8s cluster receiver, which is independent of node_conditions_to_report. Setting node_conditions_to_report: [] only disables the legacy metric via CustomMetrics(), not the entity attribute sent via metadata_exporters. I've restored the attribute in the test file. Thanks for catching this!

PR #2444, opened 2026-06-03
[chart] Deprecate cert-manager subchart and document migration

agoallikmaa:
**Description:** Adds documentation for migrating from the chart-managed cert-manager subchart to self-managed cert-manager ahead of the planned removal in https://github.com/signalfx/splunk-otel-collector-chart/pull/2049 . Also adds a Helm warning when the cert-manager subchart is enabled.

**Link to Splunk idea:** N/A

**Testing:** Manually verified the documented migration path on a local kind cluster.

**Documentation:** The changes are mainly documentation that is already described in description.

Resolves OTL-4265

PR #2451, opened 2026-06-08
[chart] Deprecate cert-manager subchart and document migration

agoallikmaa:
Replaces https://github.com/signalfx/splunk-otel-collector-chart/pull/2444 since I shouldn't have PR-ed that from a fork.

**Description:** Adds documentation for migrating from the chart-managed cert-manager subchart to self-managed cert-manager ahead of the planned removal in https://github.com/signalfx/splunk-otel-collector-chart/pull/2049 . Also adds a Helm warning when the cert-manager subchart is enabled.

**Link to Splunk idea:** N/A

**Testing:** Manually verified the documented migration path on a local kind cluster.

**Documentation:** The changes are mainly documentation that is already described in description.

Resolves OTL-4265

Also addresses the request to add this to changelog from previous PR.

### DIFF
--- a/.chloggen/deprecate-cert-manager-subchart-migration-guide.yaml
+++ b/.chloggen/deprecate-cert-manager-subchart-migration-guide.yaml
@@ -3,4 +3,4 @@ component: chart
 note: Deprecate installing cert-manager as a subchart with `certmanager.enabled=true`
 issues: [2451]
 subtext: |
-  Installing cert-manager through this chart will be removed in a future release. See the [upgrade guidelines](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md#01530-to-01540) for migration details.
+  Installing cert-manager through this chart will be removed in a future release. See the [upgrade guidelines](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md#01540-to-01550) for migration details.
--- a/.chloggen/deprecate-cert-manager-subchart-migration-guide.yaml
+++ b/.chloggen/deprecate-cert-manager-subchart-migration-guide.yaml
@@ -1,6 +1,6 @@
 change_type: deprecation
 component: chart
 note: Deprecate installing cert-manager as a subchart with `certmanager.enabled=true`
-issues: []
+issues: [2451]
 subtext: |
   Installing cert-manager through this chart will be removed in a future release. See the [upgrade guidelines](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md#01530-to-01540) for migration details.
--- a/.chloggen/deprecate-cert-manager-subchart-migration-guide.yaml
+++ b/.chloggen/deprecate-cert-manager-subchart-migration-guide.yaml
@@ -1,0 +1,6 @@
+change_type: deprecation
+component: chart
+note: Deprecate installing cert-manager as a subchart with `certmanager.enabled=true`
+issues: []
+subtext: |
+  Installing cert-manager through this chart will be removed in a future release. See the [upgrade guidelines](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md#01530-to-01540) for migration details.
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,6 +1,6 @@
 # Upgrade guidelines
 
-## 0.153.0 to 0.154.0
+## 0.154.0 to 0.155.0
 
 ### cert-manager subchart is deprecated
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,76 +1,5 @@
 # Upgrade guidelines
 
-## 0.153.1 to 0.154.0
-
-### Target allocator functionality now uses upstream chart as subchart
-
-Instead of relying on local versioning and implementation of target allocator functionality, the
-chart has moved to using the
-[upstream Target Allocator helm chart](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-target-allocator)
-directly as a subchart. This will help keep up to date with upstream features and functionality.
-
-This resulted in breaking changes to the helm chart's configuration for the target allocator, as outlined below.
-
-| Old option                       | New option                                                                                                  |
-|----------------------------------|-------------------------------------------------------------------------------------------------------------|
-| `targetAllocator`                | `targetallocator`                                                                                           |
-| `image.imagePullSecrets`         | `targetallocator.targetAllocator.imagePullSecrets`                                           |
-| `targetAllocator.image`          | `targetallocator.targetAllocator.image.repository` + `targetallocator.targetAllocator.image.tag`           |
-| `targetAllocator.resources`      | `targetallocator.targetAllocator.resources`                                                                 |
-| `targetAllocator.serviceAccount` | `targetallocator.targetAllocator.serviceAccount`                                                            |
-| `targetAllocator.config`         | `targetallocator.targetAllocator.config`                                                                    |
-
-> [!NOTE]
-> The `image.imagePullSecrets` option is still valid for non-target allocator service accounts, but the
-> new option must ALSO be configured to attach secrets to the target allocator's service account.
-
-Example old config:
-```
-image:
-  imagePullSecrets:
-    - my-registry-secret
-
-targetAllocator:
-  enabled: true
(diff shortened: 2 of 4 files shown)


UPGRADING.md
```diff
@@ -1,5 +1,180 @@
...
+```
+
+The examples in this section show only the values related to cert-manager and the
+operator webhook certificate. Keep your existing required chart values, such as
+`clusterName` and your Splunk destination configuration, when running these
+commands.
+
+#### Why the migration uses a maintenance window
+
+It might seem possible to avoid downtime by first installing a standalone
```
jinja2: The framing of this subsection implies the OpenTelemetry operator webhook is at risk of downtime, but that's not the case, it keeps serving off the existing Secret with the already-injected CA bundle. The only real risk is the webhook cert expiring during the short window with no cert-manager controller to renew it. Suggest trimming this to a couple of sentences and dropping the internal Helm explainer to keep it relevant and avoid overwhelming users, e.g.:

> The operator webhook keeps serving with its current certificate throughout the migration. cert-manager cannot issue or renew certificates in the short gap between disabling the subchart and installing standalone cert-manager, so only proceed if the webhook certificate is not close to expiry:
>
> ```bash
> kubectl get certificate -n <namespace> \
>   -l "app.kubernetes.io/instance=<release>,app.kubernetes.io/component=webhook" \
>   -o jsonpath='{.items[0].status.notAfter}'
> ```

UPGRADING.md
```diff
@@ -1,5 +1,180 @@
...
+setting `certmanager.enabled=true` is deprecated and will be removed in a future
+release. Use a separately managed cert-manager installation instead.
+
+This affects deployments that use all of these settings:
+
+- `operator.enabled=true`
+- `operator.admissionWebhooks.certManager.enabled=true`
+- `certmanager.enabled=true`
+
+For new installs that use cert-manager for the operator webhook certificate,
```
jinja2: The operator docs already documents new install guidance [here](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/auto-instrumentation-install.md#2-using-a-cert-manager-certificate). I would suggest just linking to it instead of having install related steps in the upgrading guide.

Also think that the above mentioned install doc can be improved further, like we can link to the official cert manager docs for steps to install, etc.

UPGRADING.md
```diff
@@ -1,5 +1,180 @@
...
+by Helm.
+
+Upgrade the Splunk chart with the cert-manager subchart disabled, while keeping
+the operator configured to use cert-manager:
+
+```bash
+helm upgrade <release> splunk-otel-collector-chart/splunk-otel-collector \
+  --namespace <namespace> \
+  --reuse-values \
+  --set certmanager.enabled=false \
```
jinja2, line 65: Worth noting this clean handoff relies on the cert-manager CRDs being retained. This subchart templates its CRDs as regular resources with `helm.sh/resource-policy: keep` and If anyone runs this upgrade cmd with `crds.keep=false`, it would delete the CRDs and cascade-delete CRs.

UPGRADING.md
```diff
@@ -1,5 +1,180 @@
...
+kept from the old subchart release. It is only for retained resources such as
+CRDs. Use a Helm version that supports `--take-ownership`. If cert-manager CRDs
+are managed separately in your cluster, do not adopt them into this standalone
+release. Install cert-manager with `crds.enabled=false` and follow your existing
+CRD management process instead.
+
+If your existing values use `certificateAnnotations` or `issuerAnnotations` with
+Helm hook annotations for the operator webhook certificate resources, keep those
+values during the migration. Remove them later only after testing a separate
+upgrade without the subchart.
```
jinja2: This para is referring to the `post-install,post-upgrade` hooks we suggest in our docs [here](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/auto-instrumentation-install.md#option-2-deploy-cert-manager-and-the-operator-together-deprecated), right?

Helm's default hook delete policy is `before-hook-creation`, so Helm deletes and recreates the `Certificate`/`Issuer` on the `helm upgrade` that removes the subchart. Noting my understanding of why that recreate is harmless:
- the serving Secret isn't deleted with the `Certificate` so it survives untouched;
- the `Certificate` has no finalizer, so the delete completes immediately even with the subchart's cert-manager controller already gone;
- the webhook serves off that Secret + the already-injected `caBundle`, so it never stops. The recreated CR just sits un-reconciled until the standalone cert-manager adopts it.

Questions and feedback:

1. Can we confirm in the doc that this recreate is expected and safe? A sentence would help so readers aren't alarmed if they notice the `Certificate`/`Issuer` get recreated.

2. The migration uses `--reuse-values`, so the annotations carry through untouched - "keep them during the migration" is already the default behavior; worth saying so (and linking Option 2 for where they came from).

3. "Remove them later" is a separate cleanup with a gotcha: the hook resources lack helm's meta annotations, so just dropping the hook annotations fails Helm's ownership import on the next upgrade. Could we make it actionable (re-apply the ownership annotations, then upgrade) and pull it into a short optional "removing the hook workaround" note as the last step?

4. There's an edge case worth explicitly calling out. If cert-manager runs with the upstream chart value [`enableCertificateOwnerRef=true`](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/values.yaml#L277), the serving Secret gets an ownerReference to the Cert, so this recreate deletes the Secret too. The operator should continue serving the cert from memory and new will cert-manager reissue the Secret once up. But if the operator pod restarts while the Secret is gone (and cert-manager can't reissue), it can't bootstrap TLS and the webhook fails. The flag defaults to `false`, so most users aren't affected, but a one-line caution would help anyone who flipped it.

UPGRADING.md
agoallikmaa: Done, replaced the maintenance/downtime related explanations with this.

UPGRADING.md
agoallikmaa: Added a link to cert-manager documentation in the install doc, and referred to the install doc from here instead of providing explicit yaml.

UPGRADING.md
agoallikmaa: Added a warning not to use `crds.keep=false` for this migration.

UPGRADING.md
agoallikmaa: Addressed 1, 2, 4 at the end of the migration steps section, and added the section for the optional removal of the hook annotations.

UPGRADING.md
```diff
@@ -1,5 +1,209 @@
 # Upgrade guidelines
 
+## 0.153.0 to 0.154.0
```
jinja2:
```suggestion
## 0.154.0 to 0.155.0
```

.chloggen/deprecate-cert-manager-subchart-migration-guide.yaml
```diff
@@ -0,0 +1,6 @@
+change_type: deprecation
+component: chart
+note: Deprecate installing cert-manager as a subchart with `certmanager.enabled=true`
+issues: [2451]
+subtext: |
+  Installing cert-manager through this chart will be removed in a future release. See the [upgrade guidelines](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md#01530-to-01540) for migration details.
```
jinja2:
```suggestion
  Installing cert-manager through this chart will be removed in a future release. See the [upgrade guidelines](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md#01540-to-01550) for migration details.
```

helm-charts/splunk-otel-collector/templates/NOTES.txt
```diff
@@ -48,6 +48,11 @@ Splunk OpenTelemetry Collector is installed and configured to send data to Splun
   https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/auto-instrumentation-install.md#troubleshooting-the-operator
 {{- end }}
 {{- end }}
+{{- if .Values.certmanager.enabled }}
+[WARNING] Installing cert-manager as a subchart with "certmanager.enabled=true" is deprecated and will be removed in a future release.
+          Install cert-manager separately or use the default Helm-generated operator webhook certificate.
+          Upgrade guidelines: https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md#01530-to-01540
```
jinja2:
```suggestion
          Upgrade guidelines: https://github.com/signalfx/splunk-otel-collector-chart/blob/main/UPGRADING.md#01540-to-01550
```

jinja2 (approved):

jinja2:
Branch needs to be rebased on main

PR #2453, opened 2026-06-09
Add support for deploying the gateway as a StatefulSet

dloucasfx:


examples/collector-gateway-statefulset-only/rendered_manifests/service.yaml
```diff
@@ -0,0 +1,55 @@
...
+    app.kubernetes.io/version: "0.153.0"
+    app: splunk-otel-collector
+    component: otel-collector
+    chart: splunk-otel-collector-0.153.1
+    release: default
+    app.kubernetes.io/component: otel-collector
+  annotations:
+    gateway: test-value
+spec:
+  type: ClusterIP
```
jinja2, line 22: I think we want to switch to a headless service (clusterIP: None) to get the per-pod DNS (`<pod>.<svc>.<ns>.svc.cluster.local`) published. Can you check on this? I am 80% certain that coredns will not publish these if we don't have a headless svc

examples/collector-gateway-statefulset-only/collector-gateway-only-values.yaml
```diff
@@ -0,0 +1,32 @@
...
+      cpu: 2
+      memory: 4Gi
+  service:
+    annotations:
+      gateway: "test-value"
+  mode: statefulset
+  statefulsetSpec:
+    volumeClaimTemplates:
+      - metadata:
+          name: data
```
jinja2: Can we update the example to include the config for the container to mount it? The current config only adds a pvc but the pod is not consuming it and the example feels incomplete. Maybe we should also add an example of collector config using the disk?

examples/collector-gateway-statefulset-only/rendered_manifests/service.yaml
dloucasfx: @jinja2 
Right, in order to get per-pod DNS we need to run headless service, but why do we need Pod-level addressability (ei: per-pod DNS)? 
As this will take away kube-proxy load balancing and user has to do client side load balancing. Also how will we configure the agents -> gws communication?
Finally, the multiple gw pods don't communicate among each others/aware of each other, so split brain risk, deterministic bootstrapping etc.. are not in play, and simply using the service name DNS should still work.

Now if user wants to deal with load balancing, then we can add it as a separate feature?

examples/collector-gateway-statefulset-only/collector-gateway-only-values.yaml
dloucasfx: The scope was not to address the persistent queue through volume claims, but I can see benefits to add a full on example as this question will absolutely come in the future

examples/collector-gateway-statefulset-only/rendered_manifests/service.yaml
jinja2: Right, I saw some discussion about using the per-pod dns with loadbalancing exporter (i assume for tailsampling?) and was suggesting to update the example to set `gateway.service.clusterIP: None`, but that field isn't exposed in values. And on 2nd thought clusterIP is immutable so it'd work on install, but switching it on an upgrade would force the user to delete and recreate it. We could still expose clusterIP in values, and document this for user in the example for anyone wanting a headless svc for .

The other option is, in statefulset mode, to add an additional headless service alongside the existing ClusterIP one. Our config stays the same and keeps using the VIP service, so default load balancing is unaffected; the headless service is purely opt-in. A user who wants it points their lb exporter at the headless service via a custom agent.config override.

dloucasfx: > The other option is, in statefulset mode, to add an additional headless service alongside the existing ClusterIP one. Our config stays the same and keeps using the VIP service, so default load balancing is unaffected; the headless service is purely opt-in. A user who wants it points their lb exporter at the headless service via a custom agent.config override.

Yes, if it is needed for this PR, then I would introduce another headless service which user can opt-in.
Shall we address it in this PR?

jinja2: The only issue with not adding it in the same release as statefulset mode is that `serviceName` is immutable. If we want per-pod stable DNS, serviceName has to point at the headless service, and we can't repoint it later without recreating the sts.

For the LB use case we only need the service to resolve to the list of pod IPs (?), so serviceName can stay on the VIP and the headless service can be a follow-up PR.

My preference is to add it in this PR cause I think per-pod DNS with an STS + headless service is a reasonable expectation for k8s users, and `serviceName` immutability means it is cleaner to decide/add now. That said, it's not blocking so feel free to resolve this if it feels beyond the scope of the PR

examples/collector-gateway-statefulset-only/collector-gateway-only-values.yaml
jinja2: Got it, i think even if we do not add the persistent queue collector config to the example, we should still update it with the correct config to mount the pvc (does gateway have extraMounts exposed?), otherwise the example provisions a volume which is never attached to a container

jinja2 (approved):

jinja2:
@dloucasfx can you add a values file similar to your example's values (ref this [existing](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/ci/use-custom-gateway-values.yaml) one for gw) to `ci/` folder in our chart? That should give us some more test coverage as ct install will run helm install/upgrade and test that the sts obj remains valid going forward

helm-charts/splunk-otel-collector/templates/_helpers.tpl
```diff
@@ -307,6 +307,31 @@ The name of the gateway service.
 {{  (include "splunk-otel-collector.fullname" . ) | trunc 63 | trimSuffix "-" }}
 {{- end -}}
 
+{{/*
+The name of the gateway headless service used for StatefulSet identity.
+*/}}
+{{- define "splunk-otel-collector.gatewayHeadlessServiceName" -}}
+{{ printf "%s-headless" (include "splunk-otel-collector.gatewayServiceName" .) | trunc 63 | trimSuffix "-" }}
+{{- end -}}
```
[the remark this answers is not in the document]
dloucasfx: helper is function is updated

dloucasfx:
> @dloucasfx can you add a values file similar to your example's values (ref this [existing](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/ci/use-custom-gateway-values.yaml) one for gw) to `ci/` folder in our chart? That should give us some more test coverage as ct install will run helm install/upgrade and test that the sts obj remains valid going forward

We discussed it in yesterday's standup and @dmitryax suggested that the scope of this PR is met and I will push another PR to add the policy and this example of persistent queue

PR #2458, opened 2026-06-11
Title: functional_tests: add SecureApp Java events integration test

rahulsplunk:
## Summary

  Adds a fully wired functional test for SecureApp Java security events. The test deploys the standard `java_test` workload with `secureAppEnabled: true`, triggers attack-pattern HTTP requests, and
  asserts that OTLP log records with `instrumentation_scope.name == "secureapp"` arrive at the local sink — proving the full pipeline end-to-end.

  ### What SecureApp emits

  SecureApp emits **OTLP logs**, not traces. The Java CSA agent sends log records with `instrumentation_scope.name == "secureapp"`. The `routing/logs` connector routes those to the `logs/secureapp`
  pipeline, which exports to `otlp_http/secureapp` (`logs_endpoint: <ingestUrl>/v3/event`).

  ### What's included
  
  - **`functional_tests/secureapp/`** — isolated test package (same pattern as `k8sentities`/`obi`), with its own Helm values file pointing `ingestUrl` at port 4320, its own kind job in 
  `ci-matrix.json`, and no shared state with the `functional` suite.
  - **`Test_SecureApp_JavaAttackEvents`** — deploys the chart with `secureAppEnabled: true`, waits for the OTel operator webhook to be ready, deploys `java_test`, verifies the CSA init-container was 
  injected, fires SQL injection and path traversal requests via `kubectl exec`, then asserts `instrumentation_scope.name == "secureapp"` log records arrive within 3 minutes.
  - **`ExecInPod`** — exported wrapper around the existing private `streamExec` in `internal/common.go`, used to run curl commands inside the java-test pod.
  - **`SecureAppLogsReceiverPort = 4320`** — new port constant in `internal/sinks.go`.
  - No changelog entry — test-only change.

  ### Design notes

  - Isolation is via a separate Helm values file that points `ingestUrl` at the port-4320 sink. Safe because the deployment is isolated from the `functional` suite — no shared Instrumentation CR.
  - The CSA Java agent detects attack patterns in raw HTTP requests before they reach servlet code, so the standard `java_test` Tomcat app is sufficient to trigger events.
  - Assertions are scope name + stable resource attributes (`service.name`, `telemetry.sdk.language`) — intentionally not a golden file, since exact event payload details vary across CSA versions.

  ## Files changed

  | File | Change |
  |---|---|
  | `functional_tests/secureapp/secureapp_test.go` | New isolated test package with full end-to-end test |
  | `functional_tests/secureapp/testdata/secureapp_values.yaml.tmpl` | Helm values for secureapp suite |
  | `functional_tests/internal/common.go` | `ExecInPod` exported wrapper |
  | `functional_tests/internal/sinks.go` | `SecureAppLogsReceiverPort = 4320` constant |
  | `ci-matrix.json` | Added `secureapp` to `functional_test_v2` test-job array |
  | `functional_tests/functional/auto_instrumentation_test.go` | Removed old skipped skeleton |
  | `functional_tests/functional/functional_test.go` | Removed shared-suite secureapp sink/test wiring |


rahulsplunk:
@dmitryax  and @jinja2 : I would really need help for option B in the PR description above . Please review and we can discuss

.chloggen/secureapp-it-tests.yaml
```diff
@@ -0,0 +1,6 @@
+change_type: enhancement
```
jinja2: Changelog is not needed for test updates. Please remove this and either set the label `Skip Changelog` on your PR or add `[chore]` to you PR description at start. Either of these will disable the changelog check

jinja2:
@rahulsplunk signal/sink/format look right. Main thing: I wouldn't add this to the shared functional suite - give SecureApp its own test package with its own namespace + release (like obi/k8sentities). That takes care of most of the risks you listed, and at that point I'd just do the full test in this PR instead of a skipped skeleton.

1. The collision is really just that the chart emits one Instrumentation CR per release and the apps inject via the namespace-default CR, so two CRs in default is what breaks things - not SecureApp itself. Run it as its own suite/namespace and it won't touch the existing tests.
2. You don't need a new image. CSA isn't baked in, the operator injects it through the secureAppImage swap, so reusing `java_test` with `secureAppEnabled: true` should work. The open question is whether the test Tomcat app actually does anything that makes CSA emit an event; if not, there's nothing to assert on. I'd assume the team has a demo app we can drop in if ours doesn't trigger one.
3. The template var is fine, it's additive. There's no dedicated chart value for the endpoint - the chart builds `logs_endpoint` from `splunkObservability.ingestUrl` and tacks on /v3/event, so just point `ingestUrl` at the 4320 sink in the SecureApp values file. This only works because the deployment is isolated; `ingestUrl` is shared with traces/metrics, so we couldn't repoint it like that in the shared suite.
4. Setup time isn't really a concern once it's its own package, it runs as its own job and doesn't slow the main suite.
5. For the golden file, I'd assert on `instrumentation_scope.name == "secureapp"` plus any stable attributes (or whatever's SecureApp-specific in the record) rather than an exact CompareLogs match.

rahulsplunk:
CLA accepted

functional_tests/secureapp/secureapp_test.go
```diff
@@ -0,0 +1,254 @@
...
+			}
+
+			svcName, ok := res.Get("service.name")
+			assert.True(t, ok, "resource must have service.name")
+			if ok {
+				assert.Equal(t, "java-test", svcName.Str())
+			}
+
+			assert.Positive(t, sl.At(j).LogRecords().Len(),
+				"secureapp scope must contain at least one log record")
```
dmitryax, line 251: Should we check that the event(s) associated with the attacks applied before?

dmitryax (approved):
LGTM. Comment isn't blocking

functional_tests/secureapp/secureapp_test.go
rahulsplunk: that might be more of functional check which i wanted to avoid . If something changes in the functional behavior than it will break here . Just simple test would be ok here . If needed we will enhance it later :)

PR #2459, opened 2026-06-11
[chore] ignore windows STATUS_CONTROL_C_EXIT on file streaming

jinja2:

Windows AKS functional tests intermittently fail when streaming files off the collector pod with:
`error stream protocol error: invalid exit code value "3221225786"`

3221225786 is 0xC000013A (STATUS_CONTROL_C_EXIT), which Windows containers are known to return for `cmd.exe /c type <file>` even on otherwise-successful reads. client-go's SPDY v4 protocol decodes the exec exit code as a uint8, so this out-of-range value becomes `error stream protocol error: invalid exit code value "3221225786"` and fails the test. If the stream returns this only after stdout has been fully copied, the file is already written; the fix ignores this specific code when bytes were received, and downstream parsing still guards against truncation

### DIFF
--- a/functional_tests/internal/common.go
+++ b/functional_tests/internal/common.go
@@ -47,6 +47,9 @@ const (
 	TargetAllocatorLabelSelector = "app.kubernetes.io/name=targetallocator"
 	waitTimeout                  = 3 * time.Minute
 	fileCopyHelperImage          = "busybox:1.38.0"
+
+	// winControlCExitCode is Windows STATUS_CONTROL_C_EXIT (0xC000013A).
+	winControlCExitCode = "3221225786"
 )
 
 func HostEndpoint(t *testing.T) string {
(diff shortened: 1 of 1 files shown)


functional_tests/internal/common.go
```diff
@@ -263,6 +271,15 @@ func CopyFileFromPod(t *testing.T, clientset *kubernetes.Clientset, config *rest
 		Stderr: os.Stderr, // Direct stderr to local stderr for debugging
 		Tty:    false,
 	})
+
+	// Scope the STATUS_CONTROL_C_EXIT tolerance to the Windows
+	// path so we never mask a genuine failure on the Linux path.
+	if err != nil && isWindows && strings.Contains(err.Error(), winControlCExitCode) {
+		if info, statErr := localFile.Stat(); statErr == nil && info.Size() > 0 {
+			t.Logf("ignoring benign Windows exit code error while streaming %s from pod: %v", podFilePath, err)
+			return
```
crobert-1: Probably not a big deal, but I'm slightly concerned about returning early in the case of some additional functionality being added to the end of the method in the future. We could just set err to nil and proceed as if everything's normal, but I'll leave it up to you.

functional_tests/internal/common.go
```diff
@@ -231,11 +236,14 @@ func CopyFileToPod(t *testing.T, clientset *kubernetes.Clientset, config *rest.C
 func CopyFileFromPod(t *testing.T, clientset *kubernetes.Clientset, config *rest.Config,
 	namespace, podName, containerName, podFilePath, localFilePath string,
 ) {
+	// An absolute Unix-style path means a Linux container (use `cat`); otherwise
+	// assume a Windows container (use `cmd.exe /c type`).
```
crobert-1:
```suggestion
```
This comment isn't super helpful, it's pretty clear what's going on without it.

crobert-1 (approved):
Just some nits, I'll leave it up to you whether they're important enough or not. 👍

functional_tests/internal/common.go
jinja2: done

PR #2457, opened 2026-06-11
fix(agent): correct SecureApp log routing so only secureapp-scoped logs are routed

rahulsplunk:
Fixes three bugs in the OTel agent log routing pipeline that caused incorrect
  behaviour when `secureAppEnabled=true`, particularly in clusters where logs or
  profiling pipelines were not enabled.
  
  ### Bugs fixed
  
  **1. Connector gated on wrong condition**
  
  The `routing/logs` connector was only rendered when both `secureAppEnabled` AND
  (`logsEnabled` OR `profilingEnabled`) were true. In a SecureApp-only deployment
  (no logs, no profiling), the connector was absent — but `logs/secureapp` still
  referenced it, producing an invalid collector config.
  
  **Fix:** gate the connector on `secureAppEnabled` alone.
  
  **2. `default_pipelines` hardcoded to `[logs]`**
  
  When the `logs` pipeline didn't exist, the router tried to forward non-SecureApp
  logs to a non-existent pipeline.
  
  **Fix:** `default_pipelines: [logs]` when logs or profiling is enabled;
  `default_pipelines: []` otherwise.
  
  **3. `logs/secureapp` fell back to `otlp` receiver**
  
  When logs/profiling was disabled, `logs/secureapp` used `receivers: [otlp]`
  directly — meaning SecureApp logs arrived on the same receiver as everything
  else, completely bypassing the routing connector.
  
  **Fix:** `logs/secureapp` always uses `receivers: [routing/logs]`. No fallback.
  
  **Bonus:** `logs/split` now includes `processors: [memory_limiter, batch]`
  (previously missing) and is correctly scoped inside the `secureAppEnabled` gate.
  
  ## Files changed
  
  | File | Change |
  |---|---|
  | `templates/config/_otel-agent.tpl` | Three routing bug fixes + `logs/split` processor fix |
  | `unittests/secureapp_test.yaml` | 12 new unit tests covering S=0, S=1/L=0/P=0, S=1/L=1, S=1/P=1, S=1/L=1/P=1 |
  | `.chloggen/secureapp-logs-routing-fix.yaml` | Changelog entry |
  
  ## Testing
  
  helm unittest --strict helm-charts/splunk-otel-collector
  12/12 new tests pass, 0 regressions

PR #2480, opened 2026-06-23
feat(agent): enrich SecureApp logs pipeline with k8s and resource met…

rahulsplunk:
## Summary
  
  Enriches the `logs/secureapp` pipeline in the OTel agent config with Kubernetes metadata and resource attributes, so that SecureApp events carry the same context as other log pipelines.
  
  ### What changes
  
  - **`_otel-agent.tpl`** — `logs/secureapp` processors extended from `[memory_limiter, batch]` to:
    `memory_limiter → k8s_attributes → resourcedetection → resource → resource/add_environment (when set) → batch`
  - **`unittests/secureapp_test.yaml`** — 3 new tests: enrichment processors present, `resource/add_environment` absent without `environment` value, present when set
  - **`unittests/resource_detection_test.yaml`** — updated existing test that previously asserted `logs/secureapp` has no `resourcedetection`; now correctly asserts it does (agent only — gateway
  pipeline is unchanged)
  - **`.chloggen/secureapp-agent-enrichment-processors.yaml`** — changelog entry
  
  ### Depends on
  
  #2457 (merged) — SecureApp log routing fix
  
  ### Notes
  
  - Gateway (`_otel-collector.tpl`) `logs/secureapp` pipeline is intentionally unchanged — enrichment is agent-side only
  - `resource/add_environment` is guarded by `.Values.environment` consistent with all other pipelines

### DIFF
--- a/.chloggen/secureapp-agent-enrichment-processors.yaml
+++ b/.chloggen/secureapp-agent-enrichment-processors.yaml
@@ -4,4 +4,4 @@ note: >
   Enrich SecureApp logs pipeline with Kubernetes metadata, resource detection,
   and environment attributes so that SecureApp events carry the same resource
   context as other log pipelines in the agent.
-issues: []
+issues: [2480]
--- a/.chloggen/secureapp-agent-enrichment-processors.yaml
+++ b/.chloggen/secureapp-agent-enrichment-processors.yaml
@@ -1,5 +1,5 @@
 change_type: enhancement
-component: agent
+component: agent, gateway
 note: >
   Enrich SecureApp logs pipeline with Kubernetes metadata, resource detection,
   and environment attributes so that SecureApp events carry the same resource
--- a/.chloggen/secureapp-agent-enrichment-processors.yaml
+++ b/.chloggen/secureapp-agent-enrichment-processors.yaml
@@ -1,0 +1,7 @@
+change_type: enhancement
+component: agent
+note: >
+  Enrich SecureApp logs pipeline with Kubernetes metadata, resource detection,
+  and environment attributes so that SecureApp events carry the same resource
+  context as other log pipelines in the agent.
+issues: []
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -1120,6 +1120,9 @@ service:
       processors:
         - memory_limiter
         - k8s_attributes
+        {{- if eq (include "splunk-otel-collector.autoDetectClusterName" .) "true" }}
+        - resourcedetection/k8s_cluster_name
+        {{- end }}
         - resource
         - batch
       exporters:
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -1117,7 +1117,15 @@ service:
     # secure application events — always routed via routing/logs connector
     logs/secureapp:
       receivers: [routing/logs]
-      processors: [memory_limiter, batch]
+      processors:
+        - memory_limiter
+        - k8s_attributes
+        - resourcedetection
+        - resource
+        {{- if .Values.environment }}
+        - resource/add_environment
+        {{- end }}
+        - batch
       exporters:
         {{- if .Values.gateway.enabled }}
         - otlp_grpc
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -1120,11 +1120,7 @@ service:
       processors:
         - memory_limiter
         - k8s_attributes
-        - resourcedetection
         - resource
-        {{- if .Values.environment }}
-        - resource/add_environment
-        {{- end }}
         - batch
       exporters:
         {{- if .Values.gateway.enabled }}
--- a/unittests/secureapp_test.yaml
+++ b/unittests/secureapp_test.yaml
@@ -193,3 +193,50 @@ tests:
       - matchRegex:
           path: data["relay"]
           pattern: "logs/secureapp:"
+
+  # --------------------------------------------------------------------------
+  # PR3: logs/secureapp enrichment processors
+  # --------------------------------------------------------------------------
+
+  - it: "S=1 — logs/secureapp has k8s_attributes, resourcedetection, resource processors"
+    set:
+      splunkObservability:
+        secureAppEnabled: true
+        profilingEnabled: false
+    template: configmap-agent.yaml
+    asserts:
+      - matchRegex:
+          path: data["relay"]
+          pattern: "logs/secureapp:"
+      - matchRegex:
+          path: data["relay"]
+          pattern: "- k8s_attributes"
+      - matchRegex:
+          path: data["relay"]
+          pattern: "- resourcedetection"
+      - matchRegex:
+          path: data["relay"]
+          pattern: "- resource\n"
+
+  - it: "S=1 — logs/secureapp omits resource/add_environment when environment not set"
+    set:
+      splunkObservability:
+        secureAppEnabled: true
+        profilingEnabled: false
+    template: configmap-agent.yaml
+    asserts:
+      - notMatchRegex:
+          path: data["relay"]
+          pattern: "logs/secureapp:[\\s\\S]*?- resource/add_environment[\\s\\S]*?logs/split:"
+
+  - it: "S=1 — logs/secureapp includes resource/add_environment when environment is set"
+    set:
+      splunkObservability:
+        secureAppEnabled: true
+        profilingEnabled: false
+      environment: production
+    template: configmap-agent.yaml
+    asserts:
+      - matchRegex:
+          path: data["relay"]
+          pattern: "- resource/add_environment"
--- a/unittests/secureapp_test.yaml
+++ b/unittests/secureapp_test.yaml
@@ -207,10 +207,10 @@ tests:
     asserts:
       - matchRegex:
           path: data["relay"]
-          pattern: "- k8s_attributes"
+          pattern: "logs/secureapp:\\n\\s+exporters:\\n(?:\\s+- .+\\n)+\\s+processors:\\n(?:\\s+- .+\\n)*\\s+- k8s_attributes\\n(?:\\s+- .+\\n)*\\s+receivers:"
       - matchRegex:
           path: data["relay"]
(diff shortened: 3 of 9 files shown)


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -1117,7 +1117,15 @@ service:
     # secure application events — always routed via routing/logs connector
     logs/secureapp:
       receivers: [routing/logs]
-      processors: [memory_limiter, batch]
+      processors:
+        - memory_limiter
+        - k8s_attributes
+        - resourcedetection
+        - resource
```
tarora-splunk, line 1126: Rahul, We want only k8s_attributes and resource processors in SecureApp pipeline. Please remove resourcedetection and resource/add_environment

rahulsplunk: sure let me do that

tarora-splunk (approved):

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -1117,7 +1117,11 @@ service:
     # secure application events — always routed via routing/logs connector
     logs/secureapp:
       receivers: [routing/logs]
-      processors: [memory_limiter, batch]
+      processors:
+        - memory_limiter
+        - k8s_attributes
```
jinja2, line 1122: FYI, `k8s.cluster.name` attr could be missing on auto-detect clusters

`logs/secureapp` only sets `k8s.cluster.name` when `clusterName` is set (via the `resource` processor). When it's unset (the config is [optional for certain distributions](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/values.schema.json#L1558)) nothing in this pipeline supplies it (`k8s_attributes` doesn't, and there's no resourcedetection). Other pipelines get it from detection; secureapp will ship without it.

Assuming the host/cloud attrs aren't wanted in this pipeline, you can add the cluster-name-only `resourcedetection/k8s_cluster_name` detector guarded by `autoDetectClusterName` to get the cluster name when `clusterName` is not set in config:

```
        - memory_limiter
        - k8s_attributes
        {{- if eq (include "splunk-otel-collector.autoDetectClusterName" .) "true" }}
        - resourcedetection/k8s_cluster_name
        {{- end }}
        - resource
        - batch
```

.chloggen/secureapp-agent-enrichment-processors.yaml
```diff
@@ -0,0 +1,7 @@
+change_type: enhancement
+component: agent
```
jinja2: In gateway mode, the secureapp flow is: agent `logs/secureapp` -> `otlp_grpc` -> gateway `logs/secureapp` -> o11y. The agent's `k8s_attributes` runs with [passthrough: true](https://github.com/signalfx/splunk-otel-collector-chart/blob/splunk-otel-collector-0.154.0/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl#L781-L787) whenever `gateway.enabled=true` (tags pod ip only, defers real enrichment to the gateway). The gateway completes this for the `logs` pipeline (it has a full `k8s_attributes`), but the `logs/secureapp` pipeline does not.

To make the enrichment effective for gateway users, the gateway `logs/secureapp` pipeline should add the `k8s_attribute` processor too.

unittests/secureapp_test.yaml
```diff
@@ -193,3 +193,32 @@ tests:
...
+
+  - it: "S=1 — logs/secureapp has k8s_attributes and resource processors"
+    set:
+      splunkObservability:
+        secureAppEnabled: true
+        profilingEnabled: false
+    template: configmap-agent.yaml
+    asserts:
+      - matchRegex:
+          path: data["relay"]
```
jinja2, line 209: The unscoped `matchRegex` for `- k8s_attributes` / `- resource` can pass even if the secureapp block is reverted since these processors are common to other pipelines too. A positive assertion scoped to the `logs/secureapp:` block would be a better guard for these tests.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
tarora-splunk: SecureApp does not use k8s.cluster.name at this time. When needed, we can add this?

.chloggen/secureapp-agent-enrichment-processors.yaml
tarora-splunk: SecureApp uses only the resource attributes injected bu=y collector when running in agent mode. The gateway mode attributes are not interesting for current SecureApp features.

jinja2: To clarify, this isn't about gateway's own attributes. The attributes in question (k8s.namespace.name, k8s.pod.name, k8s.pod.uid, pod labels/annotations) are the same ones this PR's k8s_attributes injects on the agent when `gateway.enabled=false`. With a gateway, the agent runs that same processor in passthrough (pod IP only), so those attributes are not added on agent and the gateway's pipeline would just re-add the identical set, not anything specific to gateway.

So as-is, the same SecureApp deployment gets pod metadata without a gateway but loses it with one. If SecureApp only relies on the resource-processor attributes, then k8s_attributes may not be needed in this pipeline at all. If it is needed on the agent, then imo gateway pipeline should carry it for parity.

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
jinja2: > SecureApp does not use k8s.cluster.name at this time.

I assume you mean that secureapp UI does not show the `k8s.cluster.name` attribute? or is it dropped on ingest completely? From a k8s user pov, being able to filter on a specific cluster is a useful feature. But if in its current form there is no way to have this attribute show up for a user in UI, then we can leave it out and add it if/when support for it lands in backend. The attribute will still exist on secureapp logs when the `clusterName` config is set, so there will a data parity issue leaving it out.

unittests/secureapp_test.yaml
rahulsplunk: fixed this thanks

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
tarora-splunk: @jinja2 so your concern is secureapp will not receive cluster name unless user provides it whereas other pipelines can still get it based on auto detect? Couple of questions:

Is the order you specified the correct order? If resource follows autodetect, it will always override using user provided value since it uses upsert. That is what the intention is here?

Shall we add same configuration for  gateway mode also?

.chloggen/secureapp-agent-enrichment-processors.yaml
tarora-splunk: @jinja2 if I understand you correctly, the collector in agent mode will not do any enrichment since it will be in passthru mode when gateway is enabled? That's why we must add these processors in gateway mode - otherwise SecureApp backend will not receive any of these attributes?

tarora-splunk: @jinja2 Given that at this time, SecureApp needs only k8s.node.name and host.name (which is set by agent), shall we keep it simple and add only k8s_attribute processor in both agent and gateway collectors?

And when we need more attributes, we can revisit this?

tarora-splunk: One more question @jinja2 or @atoulme I noticed that the k8s_attributes processor is not injecting anything for a service that is manually configured (and not auto instrumented with operator). 

What does the k8s_attribute processor need from the service deployment in order to be able to inject these k8s attributes?

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
jinja2: Yes, the order is correct. User-set clusterName take precedence.

Yes, adding to gateway would tag any secureapp logs which arrive directly at gateway (bypass agent). Here's a snippet for gateway https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl#L262-L267

.chloggen/secureapp-agent-enrichment-processors.yaml
jinja2: > @jinja2 if I understand you correctly, the collector in agent mode will not do any enrichment since it will be in passthru mode when gateway is enabled? That's why we must add these processors in gateway mode - otherwise SecureApp backend will not receive any of these attributes?

Yes, that's correct.

> shall we keep it simple and add only k8s_attribute processor in both agent and gateway collectors? And when we need more attributes, we can revisit this?

Did you mean the opposite, i.e. remove the processor from both for now? I have no preference here, I would just like the processor to be consistently included or excluded. And we can definitely make a future enhancement safely to include a new processor when needed.

jinja2: > One more question @jinja2 or @atoulme I noticed that the k8s_attributes processor is not injecting anything for a service that is manually configured (and not auto instrumented with operator). What does the k8s_attribute processor need from the service deployment in order to be able to inject these k8s attributes?

Take a look at the pod_association rules [here](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/config/_common.tpl#L197-L208). The processor uses these to identify the sending pod. I'd suggest setting the `k8s.pod.uid` using the k8s downward api to inject it as env var. There is a fallback of inspecting the connection ip but it will only work if pods are assigned unique ips

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
tarora-splunk: @rahulsplunk Please make changes as suggested by Jina. Namely:

1. Have following processors in secureapp pipeline in this order
        - memory_limiter
        - k8s_attributes
        {{- if eq (include "splunk-otel-collector.autoDetectClusterName" .) "true" }}
        - resourcedetection/k8s_cluster_name
        {{- end }}
        - resource
        - batch

 2. Configure these for both agent mode and gateway mode of collector

jinja2 (approved):

.chloggen/secureapp-agent-enrichment-processors.yaml
jinja2:
```suggestion
component: agent, gateway
```

jinja2 (approved):

PR #2482, opened 2026-06-23
feat(clusterReceiver): enable NetworkPolicy and CRD collection via k8s_objects receiver

rohit-sonawane:
**Description:**
Adds NetworkPolicy (`networking.k8s.io/v1`) and CustomResourceDefinition (`apiextensions.k8s.io/v1`) to the default `clusterReceiver.k8sObjects` list, collected via the `k8sobjects` receiver in **pull mode** (6h interval), consistent with the existing `pods` object. Also adds the required read permissions to the ClusterRole so the collector can list these API groups.

Data flows through the existing logs/objects pipeline and is sent to whichever destination is enabled:
- `splunkPlatform.logsEnabled` — Splunk Platform logs (HEC)
- `featureGates.sendK8sEventsToSplunkO11y` — Splunk Observability `/v3/event`

To opt out, remove these entries from `clusterReceiver.k8sObjects`.

Assisted-by: Claude Sonnet 4.6

**Testing:**
Rendered example manifests updated via `make render` to reflect the new objects and RBAC rules.

**Documentation:**
No doc changes needed. The existing `values.yaml` comment already notes that custom RBAC rules may be required for additional object types.

### DIFF
--- a/.chloggen/otl-4323-enable-networkpolicy-crd-collection.yaml
+++ b/.chloggen/otl-4323-enable-networkpolicy-crd-collection.yaml
@@ -1,0 +1,15 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: clusterReceiver
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Enable collection of NetworkPolicy and CustomResourceDefinition objects via k8s_objects receiver, routed to /v3/event endpoint.
+# One or more tracking issues related to the change
+issues: [4323]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Adds NetworkPolicy (networking.k8s.io/v1) and CustomResourceDefinition (apiextensions.k8s.io/v1) to the default
+  k8sObjects list using pull mode, consistent with the existing pods object. Also adds the required RBAC rules to
+  the ClusterRole so the collector can query these API groups.
--- a/.chloggen/otl-4323-enable-networkpolicy-crd-collection.yaml
+++ b/.chloggen/otl-4323-enable-networkpolicy-crd-collection.yaml
@@ -3,13 +3,16 @@ change_type: enhancement
 # The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
 component: clusterReceiver
 # A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Enable collection of NetworkPolicy and CustomResourceDefinition objects via k8s_objects receiver, routed to /v3/event endpoint.
+note: Enable collection of NetworkPolicy and CustomResourceDefinition objects by default via k8s_objects pipeline.
 # One or more tracking issues related to the change
 issues: [4323]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
-  Adds NetworkPolicy (networking.k8s.io/v1) and CustomResourceDefinition (apiextensions.k8s.io/v1) to the default
-  k8sObjects list using pull mode, consistent with the existing pods object. Also adds the required RBAC rules to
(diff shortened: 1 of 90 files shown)


helm-charts/splunk-otel-collector/templates/clusterRole.yaml
```diff
@@ -121,6 +121,22 @@ rules:
...
+  - list
+  - watch
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - get
+  - list
+  - watch
```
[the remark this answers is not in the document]
rohit-sonawane: The existing ClusterRole in this chart doesn't gate RBAC rules by feature — events.k8s.io, autoscaling, batch, and extensions are all unconditional even when their respective features are disabled. The convention here is to only use conditionals for distribution-specific rules (openshift, eks, gke/autopilot). Adding feature-based conditions for these two rules would be inconsistent with the rest of the file and add complexity for read-only permissions.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -611,6 +611,12 @@ clusterReceiver:
     - name: pods
       mode: pull
       interval: 6h
+    - name: networkpolicies
+      mode: watch
```
jinja2: Pull and watch mode generate different data sets. Given we have been sending the full obj inventory with the existing pod object config, we should keep the same mode for additional objects

.chloggen/otl-4323-enable-networkpolicy-crd-collection.yaml
```diff
@@ -0,0 +1,15 @@
...
+note: Enable collection of NetworkPolicy and CustomResourceDefinition objects via k8s_objects receiver, routed to /v3/event endpoint.
+# One or more tracking issues related to the change
+issues: [4323]
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Adds NetworkPolicy (networking.k8s.io/v1) and CustomResourceDefinition (apiextensions.k8s.io/v1) to the default
+  k8sObjects list using pull mode, consistent with the existing pods object. Also adds the required RBAC rules to
+  the ClusterRole so the collector can query these API groups.
```
jinja2:
```suggestion
  Added to the default `clusterReceiver.k8sObjects` list in pull mode (6h interval), with the
  required read permissions granted in the chart's ClusterRole.

  Collected only when a destination is enabled: `splunkPlatform.logsEnabled` (Splunk Platform logs)
  and/or `featureGates.sendK8sEventsToSplunkO11y` (Splunk Observability `/v3/event`); sent to each
  that is enabled. To opt out, remove these entries from `clusterReceiver.k8sObjects`.
```

.chloggen/otl-4323-enable-networkpolicy-crd-collection.yaml
```diff
@@ -0,0 +1,15 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: clusterReceiver
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Enable collection of NetworkPolicy and CustomResourceDefinition objects via k8s_objects receiver, routed to /v3/event endpoint.
```
jinja2:
```suggestion
note: Enable collection of NetworkPolicy and CustomResourceDefinition objects by default via k8s_objects pipeline.
```

see the subtext suggestion below for more accurate destination info

examples/add-filter-processor/rendered_manifests/clusterRole.yaml
```diff
@@ -89,3 +89,19 @@ rules:
...
   - list
   - watch
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - networkpolicies
+  verbs:
+  - get
+  - list
+  - watch
```
jinja2: is watch needed for pull mode? not a big deal to keep it, but i think it might be redundant when in pull mode

jinja2:
@rohit-sonawane can you please also update the PR description to reflect the latest changes?

examples/add-filter-processor/rendered_manifests/clusterRole.yaml
rohit-sonawane: Good catch — removed watch from both. They're pull mode (list only), so watch wasn't doing anything. Fixed in 3b79d6bf.

jinja2 (approved):

PR #2488, opened 2026-06-26
Add block_on_overflow to noDropLogs featuregate

omrozowicz-splunk:
**Description:** 
Sets `block_on_overflow`: true on the exporter sending queue to apply backpressure to the filelog receiver when the queue is full, preventing data loss. Also, change the `queue_size` depending on the featureGate `noDropLogsPipeline`. When it is on, the size should be bigger (as it is number of individual events, no the batches). Added validation on setting it to be at least 10000.

**Link to Splunk idea:** N/A

**Testing:** Manual tests, benchmarking and functional tests

**Documentation:** Added to `values.yaml`


functional_tests/logs/no_drop_logs_test.go
```diff
@@ -46,43 +48,57 @@ func Test_NoDropLogs(t *testing.T) {
...
+		}
+		// wait to ensure the queue fills and backpressure is applied before HEC starts
+		time.Sleep(10 * time.Second)
+		logsConsumer := internal.SetupHECLogsSink(t)
+
+		// test log file contains 600 log lines, min_size=100 so expect 6 batches
+		internal.WaitForLogs(t, 6, logsConsumer)
+		podLogs, podLogsErr := internal.GetPodLogs(t, clientset, internal.DefaultNamespace, podName, internal.CollectorContainerName, 100)
+		require.NoError(t, podLogsErr, "failed to get logs for pod: %s", podName)
+		require.NotContains(t, podLogs, "Exporting failed. Rejecting data.", "drop log message not found — records shouldn't be dropped with noDropLogsPipeline feature gate")
```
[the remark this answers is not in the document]
omrozowicz-splunk: done

functional_tests/logs/no_drop_logs_test.go
```diff
@@ -46,43 +48,57 @@ func Test_NoDropLogs(t *testing.T) {
...
-		// entriesNum calculation logic:
-		// - test log file contains 600 log lines
-		// - file_log receiver batches 100 lines/request
-		// - for this test, I set the sender queue size to 3 requests; ie: 3 * 100 log messages, before the queue is full.
-		// entriesNum = 600 / (3 * 100) = 2
-		internal.WaitForLogs(t, 2, logsConsumer)
-		require.Equal(t, 600, logsConsumer.LogRecordCount(), "expected number of log records does not match what received")
+		podLogs, podLogsErr := internal.GetPodLogs(t, clientset, internal.DefaultNamespace, podName, internal.CollectorContainerName, 100)
+		require.NoError(t, podLogsErr, "failed to get logs for pod: %s", podName)
+		require.Contains(t, podLogs, "Exporting failed. Rejecting data.", "expected drop log message not found — records should be dropped without noDropLogsPipeline feature gate")
```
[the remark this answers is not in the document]
omrozowicz-splunk: done

functional_tests/logs/no_drop_logs_test.go
```diff
@@ -46,43 +48,57 @@ func Test_NoDropLogs(t *testing.T) {
...
+		require.NotContains(t, podLogs, "Exporting failed. Rejecting data.", "unexpected drop log message found — records shouldn't be dropped with noDropLogsPipeline feature gate")
+		require.Equal(t, testLogLineCount, logsConsumer.LogRecordCount(), "expected number of log records does not match what received")
+		if os.Getenv("SKIP_TEARDOWN") != "true" {
+			teardown(t)
+		}
+	})
+
+	// DropLogs: without noDropLogsPipeline feature gate, queue fills and records are dropped.
+	// HEC is started after a delay so the queue fills while HEC is unavailable, triggering drops.
+	t.Run("DropLogsWithoutFeatureGate", func(t *testing.T) {
```
jinja2, line 80: I don't mind it but just thinking why we are testing this behavior. Is this meant more like a demo/proof for justifying enabling the featuregate?

helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -492,16 +495,19 @@ splunk_hec/platform_logs:
...
     storage: file_storage/persistent_queue
     {{- end }}
     num_consumers: {{ .Values.splunkPlatform.sendingQueue.numConsumers }}
     {{- if .Values.featureGates.noDropLogsPipeline }}
+    block_on_overflow: true
     batch:
       flush_timeout: 200ms
       min_size: 2048
       sizer: items
+    queue_size: {{ max 10000 .Values.splunkPlatform.sendingQueue.queueSize }}
```
jinja2, line 508: Shouldn't we set `sending_queue::sizer` to `items` ([field](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/queuebatch/config.schema.yaml#L56))  if we want queue size for featuregate to be individual records? Its default is [requests](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/queue_sender.go#L27)

For my understanding - is the reason for switching the sizer mostly about bounding the queue's footprint more predictably? Items (and bytes) give a real cap on the actual data volume, whereas requests is effectively unbounded since a batch can hold any number of records.

The README flags requests as the most performant sizer with items/bytes being progressively more expensive. Do you think it worth benchmarking the throughput/CPU impact of items sizing at some point? Since it's behind the feature gate, I don't think it needs to block this PR.

functional_tests/logs/no_drop_logs_test.go
omrozowicz-splunk: Yes, this is not necessary. I added this test case to compare the default behaviour and one with the feature gate. I think it helps to understand how this works, but I can delete it.

helm-charts/splunk-otel-collector/templates/config/_common.tpl
omrozowicz-splunk: Right, I moved sizer to the sending_queue level so now batch should inherit the same and both queue_size and min_size should rely on items. 

> For my understanding - is the reason for switching the sizer mostly about bounding the queue's footprint more predictably? 

Yes, the main goal is to no drop any logs, we'd like to avoid keeping too many records in memory and instead rely on checkpoint mechanism whenever we struggle to deliver data to Splunk.

functional_tests/logs/no_drop_logs_test.go
jinja2: Nah, we can keep it (unless it becomes flaky)

jinja2 (approved):

dmitryax (approved):
LGTM

PR #2509, opened 2026-07-14
feat(chart): add tokenPassthrough option for agent and gateway

dmitryax:
Add `agent.tokenPassthrough` and `gateway.tokenPassthrough` boolean values (default `false`) that enable passthrough of the client's authentication token to Splunk Observability and the OpAMP HTTP forwarder egress. Current behavior is preserved by default — when both flags are `false`, rendered configs are unchanged.

When enabled:
- The agent's OTLP receiver adds `include_metadata: true` on both grpc and http protocols so the incoming `X-SF-Token` flows through the `headers_setter` extension. (The gateway's OTLP receiver already has `include_metadata: true` unconditionally.)
- The OpAMP HTTP forwarder egress omits the static `X-SF-Token` header, letting the caller-supplied token flow upstream (gateway or Splunk Observability ingest).

Passthrough for Splunk Platform is supported over OTLP only; the Splunk Platform HEC exporters continue to use the token configured on the chart. This is called out in the values.yaml comments and in the docs update.

The existing `examples/otlp-token-passthrough` example was updated to use the new flags in place of the manual `include_metadata: true` overrides, and the rendered manifests were regenerated.

### DIFF
--- a/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
+++ b/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
@@ -5,7 +5,7 @@ The values can be overridden in .Values.agent.config
 {{- define "splunk-otel-collector.agentConfig" -}}
 extensions:
   {{- include "splunk-otel-collector.opampExtension" . | nindent 2 }}
-  {{- include "splunk-otel-collector.o11yIngestHttpForwarderExtension" . | nindent 2 }}
+  {{- include "splunk-otel-collector.o11yIngestHttpForwarderExtension" (merge (dict "tokenPassthrough" .Values.agent.tokenPassthrough) .) | nindent 2 }}
   {{- if eq (include "splunk-otel-collector.logsEnabled" .) "true" }}
   file_storage:
     directory: {{ .Values.logsCollection.checkpointPath }}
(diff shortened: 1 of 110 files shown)


jinja2:
@dmitryax can you update [this test's](https://github.com/signalfx/splunk-otel-collector-chart/tree/main/functional_tests/token_passthrough/testdata) values yaml to use the new config options too

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -49,9 +49,15 @@ receivers:
     protocols:
       grpc:
         endpoint: 0.0.0.0:4317
+        {{- if .Values.agent.tokenPassthrough }}
+        include_metadata: true
+        {{- end }}
```
jinja2: Hmm, why do we not have a similar conditional in the gateway config? Is this config always there in gateway? In which case, isn't this changing the default behavior for gateway

dmitryax:
Sorry this should've been a draft PR

helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
dmitryax: You right. It's true by default for gateway for data. https://github.com/signalfx/splunk-otel-collector-chart/pull/2508 brings inconsistency, but it's not released yet. So we can change it here

jinja2 (approved):

PR #2513, opened 2026-07-17
[gateway] Enable SignalFx telemetry forwarding on port 9943

crobert-1:

This allows SignalFx data to be sent to the gateway and forwarded to Splunk O11y Ingest. Also adds tests to ensure the gateway can run successfully by itself, as well as validating the http forwarder introduced in this PR works as expected.


helm-charts/splunk-otel-collector/templates/config/_common.tpl
```diff
@@ -667,9 +667,9 @@ opamp/splunk_o11y:
 {{- end }}
 
 {{/*
-Common config for Splunk O11Y Ingest HTTP Forwarder extension
+Common config for Splunk O11Y OpAmp Ingest HTTP Forwarder extension
 */}}
-{{- define "splunk-otel-collector.o11yIngestHttpForwarderExtension" -}}
+{{- define "splunk-otel-collector.o11yOpAmpHttpForwarderExtension" -}}
```
crobert-1, line 672: Renamed the variable here since the http forwarder here is OpAmp-specific, shown by the ingress port. Now that the gateway needs a generic SignalFx telemetry http forwarder it should be renamed to avoid confusion.

functional_tests/discovery/discovery_test.go
```diff
@@ -156,23 +155,6 @@ func assertRedisMetrics(t *testing.T, sink *consumertest.MetricsSink) {
 	}
 }
 
-func installCollectorChart(t *testing.T, kubeConfig, valuesTmpl string) {
```
crobert-1, line 165: Moving duplicated test setup to be in a single central location

functional_tests/gateway/passthrough_test.go
crobert-1, line 1: Moving to a gateway-specific test directory as there's a lot of overlap with the newly introduced gateway only tests, and the passthrough functionality is also focused on the gateway itself.

.chloggen/gatewaysfxforwarder.yaml
```diff
@@ -0,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+# The name of the component, or a single word describing the area of concern, (e.g. agent, clusterReceiver, gateway, operator, chart, other)
+component: gateway
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Enable SignalFx telemetry forwarding on port 9943
```
jinja2, line 6: Is there any additional context on why are adding this forwarder? I see the [PR](https://github.com/signalfx/splunk-otel-collector-chart/pull/2191/changes) which removed the receiver mentioned that users should set a receiver themself on port 9943. I am trying to understand the motivation for adding this especially since this might require customers who added the recv to change stuff in values

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
```diff
@@ -5,13 +5,18 @@ The values can be overridden in .Values.gateway.config
...
   {{- include "splunk-otel-collector.opampExtension" (merge (dict "forceDirectEndpoint" true) .) | nindent 2 }}
-  {{- include "splunk-otel-collector.o11yIngestHttpForwarderExtension" (merge (dict "forceDirectEndpoint" true "tokenPassthrough" .Values.gateway.tokenPassthrough) .) | nindent 2 }}
+  {{- include "splunk-otel-collector.o11yOpAmpHttpForwarderExtension" (merge (dict "forceDirectEndpoint" true "tokenPassthrough" .Values.gateway.tokenPassthrough) .) | nindent 2 }}
   health_check:
     endpoint: 0.0.0.0:13133
   {{- if (eq (include "splunk-otel-collector.splunkO11yEnabled" .) "true") }}
   http_forwarder:
     egress:
       endpoint: {{ include "splunk-otel-collector.o11yApiUrl" . }}
+  http_forwarder/o11y_ingest:
```
jinja2, line 15: Should this forwarder respect the `gateway.tokenPassthrough` config? In its current form, it is always passing the caller's token i think

.chloggen/gatewaysfxforwarder.yaml
crobert-1: This is meant to support users running collectors as agents on VMs/hosts, independently from the helm chart deployment. Our [default agent config uses the Signalfx exporter](https://github.com/signalfx/splunk-otel-collector/blob/3ea2971feea92d8545abac46894a570a6796174c/cmd/otelcol/config/collector/agent_config.yaml#L235), so adding the `http_forwarder` here is what's needed to support the use case.

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
crobert-1: Good point, updated.

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
```diff
@@ -5,13 +5,23 @@ The values can be overridden in .Values.gateway.config
...
+    ingress:
+      endpoint: 0.0.0.0:9943
+      include_metadata: {{ .Values.gateway.tokenPassthrough }}
+    egress:
+      endpoint: {{ include "splunk-otel-collector.o11yIngestUrl" . }}
+      {{- if not .Values.gateway.tokenPassthrough }}
+      headers:
+        X-SF-Token: "{{ include "splunk-otel-collector.splunkObservabilityAccessToken" . }}"
```
dloucasfx: Why are we introducing a new setting instead of using `headers_setter` ?

```
egress:
   endpoint: {{ include "splunk-otel-collector.o11yIngestUrl" . }}
   auth:
      authenticator: headers_setter
```

dloucasfx (changes requested):

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
crobert-1: Sorry, I missed that. My only clarification is how should we handle `.Values.gateway.tokenPassthrough` with this solution? Should we introduce an `if/else` in the template that would use `auth` if token passthrough is enabled, then just manually set the header when disabled?

dloucasfx: No need for `if/else`, the token passthrough is implied by the sender through the upsert definition in the headers_setter

.chloggen/gatewaysfxforwarder.yaml
jinja2: Thanks, the use case makes sense. I suggest adding a clarifying note like below to the subtext:

> This adds a transparent HTTP forwarder for SignalFx exporters explicitly configured to use the gateway. It does not send requests through gateway service pipelines, so gateway processors are not applied.

functional_tests/gateway/gateway_only_test.go
```diff
@@ -0,0 +1,93 @@
...
+			runKubectlFileCommand(t, testKubeConfig, kubectlApply, podManifestPath)
+			t.Cleanup(func() {
+				if os.Getenv("SKIP_TEARDOWN") == "true" {
+					return
+				}
+				runKubectlFileCommand(t, testKubeConfig, kubectlDelete, podManifestPath)
+				internal.ChartUninstall(t, testKubeConfig)
+			})
+
+			require.Eventually(t, func() bool {
```
jinja2, line 73: Can we make this test verify the token behavior as well as metric arrival? I'd suggest adding below test cases -

- `tokenPassthrough: true`: use distinct standalone and gw tokens (both are currently CHANGEME) and verify the fake ingest endpoint receives the standalone token.
- `tokenPassthrough: false`: send a different token from the sender and verify the fake ingest endpoint receives the gw token.

I think we can use the `bearertokenauth` in the SFx receiver sink in the test to validate the right token makes it through

helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl
jinja2: Iirc, the conditional behavior comes from `ingress.include_metadata`, which is mapped to `gateway.tokenPassthrough`. When true, it makes the incoming request headers available to the egress `headers_setter`; when disabled, `headers_setter` should fall back to the chart token even if the sender included one

dloucasfx: Unless I am missing it, there is currently no `gateway.tokenPassthrough` , when I added support for token passthrough through metadata we decided to just document it like this
https://github.com/signalfx/splunk-otel-collector-chart/pull/1870/changes#diff-f4ca3fd04aacbc7c0a102016a08850e9f9992e87229cd6c2dea48e2719e99791

````
### OTLP Token Passthrough

The collector's configurations for the agent and gateway support OTLP token passthrough.
This allows the collector to forward the authentication token received from the client, OTEL agent or instrumented application,
to the backend services.

To enable OTLP token passthrough, use the following configuration in your values.yaml:

For the agent:
```yaml
agent:
  config:
    receivers:
      otlp:
        protocols:
          grpc:
            include_metadata: true
          http:
            include_metadata: true
```

For the gateway:
```yaml
gateway:
  config:
    receivers:
      otlp:
        protocols:
          grpc:
            include_metadata: true
          http:
            include_metadata: true
```
````

Comment (dloucasfx) [reply helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl]: update:
I see that I was on an older branch, looks like this was recently added https://github.com/signalfx/splunk-otel-collector-chart/pull/2509 , but yes, @crobert-1 back to original ask, still no need for `if/else`

Comment (jinja2) [reply helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl]: right, I was trying to clarify my understanding of how the config option is still respected w/o the block

File: functional_tests/gateway/gateway_only_test.go (hunk 6 above)
Comment (crobert-1) [reply functional_tests/gateway/gateway_only_test.go:73]: Done 👍

File: helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl (hunk 5 above)
Comment (crobert-1) [reply helm-charts/splunk-otel-collector/templates/config/_otel-collector.tpl]: _Updated_

File: functional_tests/internal/sinks.go [hunk 7]
```diff
@@ -245,23 +245,52 @@ func SetupOTLPLogsSinkOnPort(t *testing.T, port int, logsPath string) *consumert
...
+		extensions: map[component.ID]component.Component{
+			component.MustNewIDWithName("bearertokenauth", "passthroughValidation"): btaExt,
+		},
+	}
+
+	return setupSignalfxReceiverSink(t, host, port, new(configoptional.Some(confighttp.AuthConfig{
+		Config: configauth.Config{
+			AuthenticatorID: component.MustNewIDWithName("bearertokenauth", "passthroughValidation"),
+		},
+	})))
```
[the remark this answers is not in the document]
crobert-1: This succeeded in the build. go1.26 linting says to use `new(variable)` instead of `&variable`.

helm-charts/splunk-otel-collector/values.yaml
```diff
@@ -969,6 +969,12 @@ gateway:
       containerPort: 14250
       protocol: TCP
       enabled_for: [traces]
+    signalfx:
+      containerPort: 9943
+      protocol: TCP
+      # SignalFx metrics enabled in gateway for all telemetry types since there may be
+      # bundled metrics.
+      enabled_for: [metrics, traces, logs]
```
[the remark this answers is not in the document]
crobert-1: This config was copied from a previous definition before being removed, and matches other existing port definitions, so I don't think it's relevant.

dloucasfx (approved):

jinja2 (approved):

.chloggen/gatewaysfxforwarder.yaml
```diff
@@ -0,0 +1,14 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
crobert-1, line 2: Discussed with @jinja2, we were concerned that this would break users who followed guidance from #2191 to use the `signalfx` receiver on this port. However, the receiver was removed from the Splunk distribution of the OTel Collector in[ v0.153.0](https://github.com/signalfx/splunk-otel-collector/releases/tag/v0.153.0), so this shouldn't break anyone.

PR #2514, opened 2026-07-20
enable compaction by default in persistent queue

omrozowicz-splunk:
**Description:** Enable compaction in persistent queues by default. This prevents unbounded disk growth after traffic spikes — queue files no longer accumulate in system's memory when the queue is empty. Compaction triggers when the file is ≥200 MiB and live data has dropped to ≤100 MiB. Users who need different thresholds can override via `agent.config.extensions.file_storage/persistent_queue.compaction`.

**Testing:** Manual tests


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -21,6 +21,13 @@ extensions:
     {{- if not (eq (toString .Values.splunkPlatform.fsyncEnabled) "<nil>") }}
     fsync: {{ .Values.splunkPlatform.fsyncEnabled }}
     {{- end }}
+    compaction:
+      on_start: true
```
jinja2: With on_start: true, [upstream seems to call Compact() synchronously](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.156.0/extension/storage/filestorage/extension.go#L96-L103) for each client during initialization unconditionally. I think this will delay pipeline startup, so the health endpoint will return 503 until compaction completes. I’m concerned that a sufficiently large queue could cause the default k8s liveness probe to restart the pod before startup completes.

Have we tested this config with our chart defaults by restarting a Collector with a reasonably large, realistic queue database? If not, should we default only `on_rebound` for now and leave `on_start` opt-in as a comment? Alternatively, we could add startup probe support to accommodate longer startup compaction safely.

omrozowicz-splunk: Yes, this concern is valid, with this option on on the start we could go into a restart loop. I moved it to a comment and added a short description in the template. Do you think it requires an additional documentation in md files?

jinja2: The template comment won’t be visible to most users. Since `values.yaml` already links to the [Data Persistence](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/docs/advanced-configuration.md#data-persistence) section, could we document the default rebound compaction and the `on_start` option there?

Suggested addition:

```
### Persistent queue compaction

The chart enables online (`on_rebound`) compaction by default. Compaction runs when the queue database has allocated at least 200 MiB and its live data has dropped to 100 MiB or less.

Startup (`on_start`) compaction is disabled by default because compacting a large queue delays Collector readiness and can cause K8s liveness probe failures. Before enabling it, set `livenessProbe.initialDelaySeconds` high enough to cover the expected compaction time.

To enable startup compaction, override the Collector configuration:

```yaml
agent:
  config:
    extensions:
      file_storage/persistent_queue:
        compaction:
          on_start: true
```

_Since the chart doesn’t currently expose a startup probe, i included a note about updating the initialDelaySeconds for liveness prode instead._

jinja2 (approved):

PR #2531, opened 2026-07-29
[chore] add k8s 1.36 support

jinja2:
Changes in the PR -

- Onboards K8s 1.36 in the functional test matrix and removes 1.32 to keep the latest four versions.
- Updates Kind to v0.32.0
- Migrates the control-plane histogram suite to `pmetricassert` snapshots
- Stabilizes CoreDNS assertions by generating cache-hit traffic
- Stabilizes kubeletstats assertions by comparing datapoints from the collector agent pod


functional_tests/functional/functional_test.go
```diff
@@ -1117,11 +1124,22 @@ func testAgentMetricsTemplate(t *testing.T, metricsSink *consumertest.MetricsSin
 }
 
 // tryMetricsComparison performs metric comparison using pmetrictest.CompareMetrics and returns error
-func tryMetricsComparison(expected pmetric.Metrics, actual pmetric.Metrics) error {
+func tryMetricsComparison(expected pmetric.Metrics, actual pmetric.Metrics, podNamePrefix ...string) error {
+	if len(podNamePrefix) > 0 {
```
crobert-1: A short comment about podNamePrefix might be helpful here, it's not obvious from the name why it would require special logic, or what would be required if it's included.

Also, it may make sense to just pull this functionality out into its own method to clean up `tryMetricsComparison` a bit.

crobert-1 (approved):
Just one nit, looks good otherwise

functional_tests/functional/functional_test.go
jinja2: done

PR #2542, opened 2026-08-07
[chore][histogram test] Re-check expected metric contents to address flaky failures

crobert-1:

Use `require.EventuallyWithT` to retry flaky metric check. The test has passed 3/3 times on this PR.

### DIFF
--- a/functional_tests/internal/metricassert.go
+++ b/functional_tests/internal/metricassert.go
@@ -148,6 +148,7 @@ func AssertMetricsSnapshot(t *testing.T, sink *consumertest.MetricsSink, targetM
 	require.NoError(t, err, "Failed to read expected counts from %s", assertionFile)
 
 	require.Eventually(t, func() bool {
+		firstFailure := true
 		sinkMetrics := sink.AllMetrics()
 		for i := len(sinkMetrics) - 1; i >= 0; i-- {
 			m := sinkMetrics[i]
@@ -166,7 +167,10 @@ func AssertMetricsSnapshot(t *testing.T, sink *consumertest.MetricsSink, targetM
 			if assertErr == nil {
 				return true
 			}
-			t.Logf("metrics assertion failed: %v", assertErr)
+			if firstFailure {
+				t.Logf("metrics assertion failed: %v", assertErr)
+				firstFailure = false
+			}
 		}
 		return false
 	}, timeout, interval, "Failed to find expected metrics in allotted time")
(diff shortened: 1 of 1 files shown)


dloucasfx (approved):

functional_tests/internal/metricassert.go
```diff
@@ -147,21 +148,22 @@ func AssertMetricsSnapshot(t *testing.T, sink *consumertest.MetricsSink, targetM
...
+					wantResources, wantMetrics, selected.MetricCount())
+			}
+			assert.NoError(c, assertErr, "Metric assertion failed for %s. Error: %v", assertionFile, assertErr)
 		}
-		require.NoError(t, assertErr, "Metric assertion failed for %s. Error: %v", assertionFile, assertErr)
-	}
-
-	t.Logf("Metric assertion passed for %d metrics (%s)", selected.MetricCount(), assertionFile)
+		t.Logf("Metric assertion passed for %d metrics (%s)", selected.MetricCount(), assertionFile)
+	}, 5*time.Minute, 1*time.Second, "Failed to find expected metrics in allotted time")
```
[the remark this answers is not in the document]
crobert-1: Removed existing `selectMetricSetByCountsWithTimeout` and `richestMetricSet` as a result of this comment. The two step process of "selecting metrics by count" then "compare expected metrics vs. selected" is prone to flakiness, we've seen this many times in the collector and chart repos. It's much more stable to just recheck all received metrics within the timeout limit, since the "selected" group of metrics causes problems.

functional_tests/internal/metricassert.go
```diff
@@ -147,73 +147,31 @@ func AssertMetricsSnapshot(t *testing.T, sink *consumertest.MetricsSink, targetM
...
+			if m.MetricCount() != wantMetrics {
+				continue
+			}
+			actual := prepareMetricsAssertion(m, cfg)
+			maybeUpdateExpectedMetricsAssertion(t, assertionFile, actual, opts...)
+			assertErr := pmetricassert.AssertMetrics(assertionFile, actual)
+			if assertErr == nil {
+				return true
+			}
+			t.Logf("metrics assertion failed: %v", assertErr)
```
crobert-1: This might be a little spammy since it'll log for every comparison that fails, rather than just one. I think it's better to be a bit spammy here to have an idea of why the comparison is failing if we never find the metric in the format we're looking for, rather than one generic log message saying "Failed to find expected metrics in allotted time" after the timeout.

The alternative would be to simply log the first failure every tick. This would be the most recent resource group that is found with the target metric, so it would be a good indicator of why the comparison is failing.

crobert-1: I modified it to just log the first failure per tick.

crobert-1:
Closing in favor of https://github.com/signalfx/splunk-otel-collector-chart/pull/2546

PR #2544, opened 2026-08-07
[chore] Upgrade Prometheus receiver versions as part of chart upgrade

crobert-1:

This is intended to resolve [failures that will happen every upgrade](https://github.com/signalfx/splunk-otel-collector-chart/actions/runs/31213550790/job/92981907376?pr=2543).

The alternative to this is to simply update the tests to match a regex version. This may allow prometheus receiver versions to be different than what's expected, but it may be simpler logic than what I've added here.

**Testing:**
Tested locally and worked as expected


crobert-1:
I'm going to update to just match version by regex rather than part of the upgrade process.

PR #2546, opened 2026-08-07
[chore][histogram test] Pin API server request datapoints

jinja2:

Reduce flakiness in API-server non-histogram test assertion by selecting the deterministic `apiserver_request_total` datapoint generated by the test’s k8s service request. This prevents the test from arbitrarily selecting `/healthz`, whose empty Prometheus labels are omitted during OTLP conversion and lack component, resource, scope, and version.

Once pmetricassert supports datapoints/include, this selector can be replaced with a YAML-only assertion.

### DIFF
--- a/functional_tests/internal/metricassert.go
+++ b/functional_tests/internal/metricassert.go
@@ -26,6 +26,8 @@ type metricsAssertionConfig struct {
 	regexAttrs                     map[string]string
 	firstDatapointOnly             []string
 	maxDatapointsPerMetric         int
+	selectedNumberMetric           string
+	selectedNumberDatapoint        map[string]string
 	exactDatapointAttrs            map[string]struct{}
 	includeHistogramExplicitBounds bool
 }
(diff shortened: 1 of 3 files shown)


functional_tests/internal/metricassert.go
```diff
@@ -118,6 +120,15 @@ func WithMaxDatapointsPerMetric(maxDatapoints int) MetricsAssertionOption {
 	}
 }
 
+// WithSelectedNumberDatapoint retains datapoints for a number metric that
+// contain the provided attributes.
```
crobert-1, line 124: I'm pretty lost on the wording and method name here. "number metric" and "NumberDatapoint" aren't obvious to me, can you expand what this is supposed to mean?

jinja2: Its shared datapoint type for gauges and sums, pmetric uses the same. I can rename it though

crobert-1 (approved):

PR #2547, opened 2026-08-08
Add support for cloud management of otel workloads

owais:
When `cloudManagement.enabled` is set to `true`, the chart installs the opamp bridge component and configures it to report configmaps for all workloads installed by the helm chart to the remote opamp server. It also accept new config from the server and applies that locally to the workload.

**Testing:** 
Added tests and tested manually on a kind cluster

**Documentation:**
updated README.md


functional_tests/cloud_management/testdata/cloud_management_values.yaml.tmpl
```diff
@@ -0,0 +1,23 @@
+splunkObservability:
+  realm: test
+  accessToken: test-token
+  apiUrl: {{ .ApiURL }}
+  ingestUrl: {{ .IngestURL }}
+
+clusterName: cloud-management-test
+
+cloudManagement:
```
dmitryax: Maybe just `opampBridge`? `cloud` within this repo typically refers to AWS/GCP/Azure where the helm chart is installed so it might be a bit confusing.


cc @aunshc @atoulme WDYT?

dmitryax:
Please resolve the CI failures related to this PR

helm-charts/splunk-otel-collector/templates/opamp-bridge.yaml
```diff
@@ -0,0 +1,316 @@
...
+    chart: {{ template "splunk-otel-collector.chart" . }}
+    release: {{ .Release.Name }}
+  {{- if $bridge.serviceAccount.annotations }}
+  annotations:
+    {{- toYaml $bridge.serviceAccount.annotations | nindent 4 }}
+  {{- end }}
+---
+{{- end }}
+{{- if $bridge.rbac.create }}
+apiVersion: rbac.authorization.k8s.io/v1
```
jinja2: Can we break these up into their individual yamls? Imo, that'll make it easier to maintain and review.

functional_tests/cloud_management/testdata/cloud_management_values.yaml.tmpl
owais: changed to `remoteManagement`. I think that's what we've been calling it internally and when talking to customers as well.

aunshc: Yes, best focused on the capability users can relate to - `remoteManagement`makes sense.

helm-charts/splunk-otel-collector/templates/opamp-bridge-configmap.yaml
```diff
@@ -0,0 +1,105 @@
...
+            non_identifying_attributes:
+              {{- include "splunk-otel-collector.opampBridgeWorkloadAttributes" (dict "root" . "workload" $bridge.workloads.agent) | nindent 14 }}
+          workloadRef:
+            apiVersion: apps/v1
+            kind: DaemonSet
+            name: {{ printf "%s-agent" (include "splunk-otel-collector.fullname" .) | quote }}
+          config:
+            {{ $bridge.workloads.agent.configName }}:
+              kind: configmap
+              name: {{ printf "%s-otel-agent" (include "splunk-otel-collector.fullname" .) | quote }}
```
jinja2, line 67: Does the bridge write `relay` into the same configmap Helm manages? If so, what is intended to happen on a later `helm upgrade`? Does Helm replace the remotely managed config, and the bridge reconciles it back afterward?

eks_addon/run_tests.sh
```diff
@@ -74,17 +74,24 @@ test_no_chart_dependencies() {
 test_images() {
   echo "⏳ Running test_images ..."
 
-  # Find all repository entries in the values file
-  local image_repos=$(yq e '.image | .. | select(has("repository")) | .repository' "${EKS_CHART_DIR}/values.yaml" | sort -u)
-  if [[ -z "${image_repos}" ]]; then
-    echo "❌ Image verification failed: No image repositories found in values.yaml"
+  # Check rendered workload images. Some optional features have image defaults in values.yaml,
```
jinja2: We exclude unsupported features through [`disabled_properties`](https://github.com/signalfx/splunk-otel-collector-chart/blob/1ab08b3fc7fc5d7b5e3ab23cd7a59dc99cf3aa45/eks_addon/common.sh#L54-L62); could we add `remoteManagement` there?

functional_tests/remote_management/remote_management_test.go
```diff
@@ -0,0 +1,184 @@
...
+	if os.Getenv("SKIP_TESTS") == "true" {
+		t.Log("Skipping tests as SKIP_TESTS is set to true")
+		return
+	}
+
+	clientset, err := internal.GetKubeClient(testKubeConfig)
+	require.NoError(t, err)
+	internal.CheckPodsReady(t, clientset, internal.DefaultNamespace, bridgeLabelSelector, 3*time.Minute, 0)
+
+	t.Run("bridge deployment is configured", func(t *testing.T) {
```
jinja2: This is useful as an in-cluster wiring smoke test, but it does not seem to provide any functional opamp coverage atm. Could the current coverage live in the Helm unit tests plus a `ct install` values file, with a dedicated functional workflow added later when it can verify more opamp-specific functionality?

helm-charts/splunk-otel-collector/values.schema.json
```diff
@@ -1536,6 +1705,11 @@
         }
       }
     },
+    "opentelemetry-target-allocator": {
```
jinja2: Is this unrelated `opentelemetry-target-allocator` schema addition intentional in this PR?

README.md
```diff
@@ -196,6 +196,13 @@ Sending data to Splunk Observability Cloud
 helm install my-splunk-otel-collector --set="splunkObservability.realm=us0,splunkObservability.accessToken=xxxxxx,clusterName=my-cluster" splunk-otel-collector-chart/splunk-otel-collector
 ```
 
+To enable Splunk Remote Management for the collectors installed by this chart, add
```
jinja2, line 199: Could we add a short Remote Management section to `advanced-configuration.md` and link it here? It would be good to document the operational behavior of this new workload for users who are not familiar with these concepts. If this is already covered in public-facing documentation, linking to that here would work too.

eks_addon/run_tests.sh
owais: Thanks moving to this also improved another wart

helm-charts/splunk-otel-collector/templates/opamp-bridge-configmap.yaml
owais: Yes, it writes back into the same configmap. This isn't handled yet in this PR. It is tracked separately in JIRA and I have another subsequent PR after this that takes care of this case.

owais: But to answer your question, no helm will not override config set by the remote server unless explicitly directed to do so.

jinja2: Ack, Helm 4 uses server-side apply, so my concern was that helm upgrade might fail if there is ownership conflict with bridge

jinja2 (approved):

PR #2549, opened 2026-08-10
[chore] Add scope.version regex for pmetricassert tests

jinja2:

Pins pdatatest to the commit from open-telemetry/opentelemetry-collector-contrib#49336 and uses scope-version regex matching in the histogram metric assertions. Regenerated snapshots now preserve the matcher, avoiding release-by-release scope version updates.


functional_tests/internal/metricassert.go
```diff
@@ -343,6 +352,10 @@ func markFlexibleAttrs(file string, volatile []string, regex map[string]string,
 			if !scopeOK {
 				return fmt.Errorf("parse assertion file %s: resources[%d].scopes[%d] must be a map", file, i, j)
 			}
+			if scopeVersionRegex != "" {
```
crobert-1, line 355: Where do we actually check to make sure the regex matches the received version?

jinja2: The actual check happens in `pmetricassert.AssertMetrics(assertionFile, actual)`, where pmetricassert reads version/regex and full-string matches it against the received scope version.

This method is only for snapshot auto-generation with `UPDATE_EXPECTED_RESULTS=true`. Since the pmetricassert writer emits an exact version, we rewrite it to `version/regex` when generating snapshots.

crobert-1 (approved):

PR #2552, opened 2026-08-10
[chore] Fix flaky k8s-clusters tests

jinja2:

Run the k8s cluster metrics assertion after chart installation but before deploying instrumentation test workloads. This prevents Node.js, Java, .NET, or Python resources from adding timing-dependent datapoints to the exact pmetricassert snapshot.
Normal Kind runs execute the assertion once at this isolated point. The existing `SKIP_SETUP=true` behavior is retained for manual runs that reuse an already-deployed cluster, although it is not used by current CI workflows.

Also add an opt-in `WithWaitForSnapshotMatch` option because matching resource and metric counts does not always guarantee that datapoints have converged. It retries the full `pmetricassert` comparison until the snapshot matches.

### DIFF
--- a/functional_tests/functional/functional_test.go
+++ b/functional_tests/functional/functional_test.go
@@ -116,7 +116,7 @@ func setupSinks(t *testing.T) {
 	}
 }
 
-func deployChartsAndApps(t *testing.T, testKubeConfig string) {
+func deployCharts(t *testing.T, testKubeConfig string) {
 	kubeTestEnv, setKubeTestEnv := os.LookupEnv("KUBE_TEST_ENV")
 	require.True(t, setKubeTestEnv, "the environment variable KUBE_TEST_ENV must be set")
 	kubeConfig, err := clientcmd.BuildConfigFromFlags("", testKubeConfig)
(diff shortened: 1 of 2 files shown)


functional_tests/functional/functional_test.go
```diff
@@ -204,6 +201,30 @@ func deployChartsAndApps(t *testing.T, testKubeConfig string) {
...
+func deployTestResources(t *testing.T, testKubeConfig string) {
+	kubeTestEnv := requireEnv(t, "KUBE_TEST_ENV")
+	kubeConfig, err := clientcmd.BuildConfigFromFlags("", testKubeConfig)
+	require.NoError(t, err)
+	client, err := kubernetes.NewForConfig(kubeConfig)
+	require.NoError(t, err)
+	dynamicClient, err := dynamic.NewForConfig(kubeConfig)
+	require.NoError(t, err)
+	decode := scheme.Codecs.UniversalDeserializer().Decode
+
```
dloucasfx, line 219: Would we use `deployTestResources` without `deployCharts`, as I am wondering if we should reuse the same instance of kubernetes client .

dloucasfx (approved):
Overall, LGTM! 
Just one non-blocking observation regarding the reuse of the k8s client

functional_tests/functional/functional_test.go
jinja2: I think i can pass it in, let me check

PR #2565, opened 2026-08-18
[docs] Document gateway routing behavior

jinja2:

Document the gateway routing contract, clarifying which telemetry is processed by gateway pipelines, transparently forwarded, or sent directly to backends. Include an example for routing k8s cluster metrics through the gateway.

### DIFF
--- a/docs/advanced-configuration.md
+++ b/docs/advanced-configuration.md
@@ -28,14 +28,17 @@ clusterName: my-k8s-cluster
 
 ## Gateway routing
 
-Setting `gateway.enabled: true` deploys the gateway and automatically sends the
-agent's metrics, traces, logs, profiling data, SecureApp events, and entity events
-to its OTLP pipelines. It does not route every chart-managed data path through
-gateway processors.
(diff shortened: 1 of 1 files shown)


docs/advanced-configuration.md
```diff
@@ -26,6 +26,46 @@ splunkObservability:
 clusterName: my-k8s-cluster
 ```
 
+## Gateway routing
+
+Setting `gateway.enabled: true` deploys the gateway and automatically sends the
+agent's metrics, traces, logs, profiling data, SecureApp events, and entity events
+to its OTLP pipelines. It does not route every chart-managed data path through
```
dloucasfx: What do you mean by "OTLP pipelines" , I don't think you mean using the `otlp` data format? as not all exporters send to otlp receiver (as you stated in below table). Can you make it clearer, otherwise, the rest LGTM

dloucasfx (approved):

docs/advanced-configuration.md
jinja2: Updated

PR #2575, opened 2026-08-26
Preserve remote-managed ConfigMaps during upgrades

owais:
**Description:**
Preserve OpAMP-managed collector ConfigMap data.relay during Helm upgrades so remote configuration applied by the bridge is not overwritten after takeover. Adds `remoteManagement.collectorConfig.upgradeStrategy=resetFromHelm` as an escape hatch to force Helm-rendered collector config. Also added `service.version` non-identifying attr as an opportunistic update.

**Testing:** Tested manually & added functional tests.

**Documentation:** Mentioned the escape hatch in advanced configuration.

### DIFF
--- a/docs/advanced-configuration.md
+++ b/docs/advanced-configuration.md
@@ -65,6 +65,20 @@ can apply remote configuration and roll out workload changes. When Remote
 Management is enabled, chart-created collectors do not open their own direct
 OpAMP sessions to Splunk Observability Cloud; the bridge owns that connection.
 
+When Remote Management manages a chart-created collector, Helm marks the
+collector ConfigMap and preserves its live `data.relay` value on later upgrades
+so remote configuration applied by the bridge is not overwritten. Set
+`remoteManagement.collectorConfig.upgradeStrategy: resetFromHelm` to force Helm
+to render collector config from chart values again. Because OpAMP Bridge owns the
+live `data.relay` field after applying remote config, use Helm's
+`--force-conflicts` flag with `resetFromHelm`:
+
+```bash
+helm upgrade <release> <chart> \
+  --set remoteManagement.collectorConfig.upgradeStrategy=resetFromHelm \
+  --force-conflicts
+```
+
 By default, the bridge is configured to manage every enabled collector workload
 installed by this chart: the agent, gateway, and cluster receiver. If you do not
 override `remoteManagement.opampBridge.workloads`, every installed chart-managed
--- a/unittests/remote_management_test.yaml
+++ b/unittests/remote_management_test.yaml
@@ -238,6 +238,9 @@ tests:
           pattern: 'otelcol.service.mode: agent'
       - matchRegex:
           path: data["config.yaml"]
+          pattern: 'service.version: [0-9]+\.[0-9]+\.[0-9]+'
+      - matchRegex:
+          path: data["config.yaml"]
           pattern: 'otelcol.service.mode: clusterReceiver'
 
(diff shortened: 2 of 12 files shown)


unittests/remote_management_test.yaml
```diff
@@ -236,6 +236,9 @@ tests:
       - matchRegex:
           path: data["config.yaml"]
           pattern: 'otelcol.service.mode: agent'
+      - matchRegex:
+          path: data["config.yaml"]
+          pattern: 'service.version: 0\.158\.0'
```
jinja2: Is this version part of input to the test fixture? If it is instead from a chart default (like collector image), can we loosen this check so this assertion will not need to be updated on every chart release

docs/advanced-configuration.md
```diff
@@ -65,6 +65,20 @@ can apply remote configuration and roll out workload changes. When Remote
 Management is enabled, chart-created collectors do not open their own direct
 OpAMP sessions to Splunk Observability Cloud; the bridge owns that connection.
 
+When Remote Management manages a chart-created collector, Helm marks the
+collector ConfigMap and preserves its live `data.relay` value on later upgrades
+so remote configuration applied by the bridge is not overwritten. Set
+`remoteManagement.collectorConfig.upgradeStrategy: resetFromHelm` to force Helm
+to render collector config from chart values again. Because OpAMP Bridge owns the
+live `data.relay` field after applying remote config, use Helm's
+`--force-conflicts` flag with `resetFromHelm`:
```
jinja2, line 74: Since we support [Helm 3 as well as Helm 4](https://github.com/signalfx/splunk-otel-collector-chart#prerequisites), could you confirm that this preserve/reset upgrade flow works with Helm 3? The functional tests use the v4 SDK, and --force-conflicts is a Helm 4 flag. A brief version-specific note here may be enough if you can verify the behavior with Helm 3.

owais: I initially built it on 3 and that worked for both manual testing and functional tests but functional were failing in CI because of helm 4 so I added this. I'll check if the matrix allows running tests per helm version so we can cover both.

owais: @jinja2 updated functional tests to run against helm 3 and 4 both so we can catch regressions or differences. I can rebase this on that once merged. PR: https://github.com/signalfx/splunk-otel-collector-chart/pull/2577

owais: @jinja2 I'll update the helm 3/4 test harness changes but this change is a bit urgent as it needs to land for customers to have a smooth upgrade experience later so can we merge it now with just helm 4 test coverage? I'll work on helm 3 coverage in parallel

jinja2: Sounds good, but to clarify, my og comment was a suggestion to update the document and a manual check with helm 3 would have been good enough. If we are not sure about helm 3 behavior, can we just update the document with a note that these instructions are for helm 4?

> this change is a bit urgent as it needs to land for customers to have a smooth upgrade experience

Jfyi, merge to this repo does not release new chart versions. If a patch needs to go out sooner than the next scheduled release, ktlo can help make one.

jinja2 (approved):

docs/advanced-configuration.md
owais: I did all manual verification needed for helm 3. I actually was using helm 3 during development and verified 4 only at the end once the feature was ready so I'm confident about 3 working fine.

PR #2577, opened 2026-08-27
Run functional tests with Helm 3 and 4

owais:
**Description:**
Run helm functional tests against both helm 3 and 4. Added test build tags to build the functional tests with either v3 or v4. Test matrix selects the version with vars.

**Testing:** CI is green

**Documentation:** None needed.

Assisted-by: OpenAI Codex


jinja2 (commented):
I’m supportive of adding the Helm 3 coverage needed for the remote-management upgrade flow, but do we need to run every functional suite against both Helm versions? This doubles the Kind matrix and adds a dual-SDK compatibility layer that we'll need to maintain, while most suites test collector behavior which are not impacted much by Helm-version-specific semantics. The existing Helm 3 lint and install coverage seems sufficient for the rest of the chart.

Could we scope the additional Helm 3 coverage to version-sensitive upgrade tests, such as the remote-management preserve/reset flow, and run it against a single k8s version? A focused CLI-based test might also avoid maintaining both Helm SDKs as dependencies. Since Helm 3 reaches EOL in early 2027 and we’ll likely drop support then, carrying its SDK could create ongoing dependency and vulnerability-maintenance overhead for relatively limited coverage. WDYT?

PR #2593, opened 2026-09-11
[chore] eks privileged workflow foundation

jinja2:

Add an inactive reusable workflow for the privileged EKS functional tests, preserving the existing six install and upgrade test variants.

Before any credential-bearing job can start, an unprivileged validation job checks GitHub-generated event metadata. It accepts only internal, non-Dependabot PRs targeting main on the expected PR actions and merge ref, or manual runs dispatched from main. Repository names and immutable IDs must match, PRs cannot override the collector image, and manual image overrides are restricted to approved Splunk repositories.

The EKS matrix depends on that validation and retains the protected environment, minimal GitHub permissions, pinned actions, fixed cluster/region combinations and one-hour timeout.

No existing workflow calls this reusable workflow yet, so this foundation PR does not change current CI behavior, credentials or AWS access. It prepares the follow-up GitHub OIDC migration.

### DIFF
--- a/.github/workflows/eks-privileged-tests.yaml
+++ b/.github/workflows/eks-privileged-tests.yaml
@@ -9,8 +9,7 @@ on:
         default: ''
         type: string
 
-# Callers must grant only the permissions required by the jobs below. A called
-# workflow cannot elevate GITHUB_TOKEN permissions beyond its caller.
+# Deny GitHub token access by default; EKS jobs grant only what they need.
 permissions: {}
 
 env:
@@ -20,8 +19,8 @@ env:
   SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
 
 jobs:
-  authorize-caller:
-    name: Authorize EKS test caller
+  validate-caller:
+    name: Validate EKS test caller
     permissions: {}
     runs-on: ubuntu-latest
     timeout-minutes: 5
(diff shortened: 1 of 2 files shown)


.github/workflows/eks-privileged-tests.yaml
```diff
@@ -0,0 +1,212 @@
...
+          PR_AUTHOR: ${{ github.event.pull_request.user.login || '' }}
+          REF: ${{ github.ref }}
+          REPOSITORY: ${{ github.repository }}
+          REPOSITORY_ID: ${{ github.repository_id }}
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          if [[ "$REPOSITORY" != "signalfx/splunk-otel-collector-chart" || "$REPOSITORY_ID" != "286847478" ]]; then
+            echo "::error::This workflow only authorizes the splunk-otel-collector-chart repository."
```
pjanotti, line 47: Just curious: is this `::error::This ...` some kind of convention?

pjanotti (approved):

.github/workflows/eks-privileged-tests.yaml
jinja2: yeah, just for more visibility. It will show up under annotations https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message

PR #2606, opened 2026-09-16
Merge histogram buckets to fit cloud limit as part of metric processing pipeline

agoallikmaa:
Previously if histogram bucket count exceeded the 31 limit, the excess was dropped using prometheus receiver metric relabel rules. This is now replaced by using merge_histogram_buckets function as part of the metric processing pipeline instead.


helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl
```diff
@@ -548,9 +548,6 @@ receivers:
                     kubeproxy_sync_proxy_rules_service_changes_pending|\
                     kubeproxy_sync_proxy_rules_duration_seconds|\
                     kubeproxy_network_programming_duration_seconds)(?:_sum|_count|_bucket)?"
-                - action: drop
-                  regex: 'kubeproxy_network_programming_duration_seconds_bucket;([1-3][1-46-9]|[4-9][1-9]|100|110|115|270)\.0'
-                  source_labels: [__name__, le]
```
jinja2, line 553: Is there a reason to drop this? We added this intentionally to bring this under the limit and the buckets were chosen for optimal resolution while still being under 32. Passing this through the transform will change the buckets retained.

helm-charts/splunk-otel-collector/templates/config/_otel-k8s-cluster-receiver-config.tpl
```diff
@@ -152,6 +152,15 @@ processors:
       - set(resource.attributes["k8s.deployment.name"], resource.attributes["k8s.hpa.scaletargetref.name"])
         where IsMatch(resource.attributes["k8s.hpa.scaletargetref.kind"], "Deployment")
 
+  {{- if and (and (hasPrefix "eks" .Values.distribution) .Values.featureGates.enableEKSApiServerMetrics) (eq (include "splunk-otel-collector.o11yMetricsEnabled" .) "true") }}
+  transform/limit_histogram_buckets:
+    error_mode: ignore
+    metric_statements:
+      - context: datapoint
+        statements:
+          - merge_histogram_buckets(31, method="limit_buckets")
```
jinja2, line 161: I think the first arg here is # bucket and not bounds in which case it should be 32. Please check this config against a similar change we already made in splunk-otel-collector.

PR #2608, opened 2026-09-17
Add talos distribution and skip os-release mount for it. Parallel to #2089's "Skip os-release mount for GKE", but names the enum addition too, since that's the user-facing part.

buchanandc:
Description: 

Talos Linux does not ship /usr/lib/os-release and mounts its root filesystem as a read-only squashfs. The agent DaemonSet bind-mounts that host path whenever metrics are enabled, so the container runtime tries to create the missing source, fails on the read-only filesystem, and never generates the container spec. Every agent pod stays in CreateContainerError indefinitely:

Error: failed to generate container spec: failed to apply OCI options:
failed to mkdir "/usr/lib/os-release": mkdir /usr/lib/os-release: read-only file system

Confirmed on the node:

talosctl ls /usr/lib/os-release   → lstat /usr/lib/os-release: no such file or directory
talosctl ls -l /etc/os-release    → -rw-r--r--  232 bytes  (a real file, not a symlink)
/proc/mounts                      → /  squashfs  ro

This adds talos to the distribution enum and skips the mount for it, following the existing gke/autopilot and gke exemptions. Nothing is lost: the mount exists only for distributions where /etc/os-release is a symlink into /usr/lib, and the hostmetrics receiver reads host OS details through the /hostfs/etc mount the chart already creates — where Talos does provide os-release.

It is not fixable outside the chart. Talos refuses machine.files writes outside /var (create operation not allowed outside of /var), and a bind mount cannot target a path that does not exist.

Unlike the gke exemption in #2089, this also adds talos to the two checks in _otel-agent.tpl that treat distribution: "" as vanilla Kubernetes. GKE sits outside that allowlist by design because its control plane is managed, so #2089 needed no change there. Talos runs a self-managed control plane as static pods in kube-system, and operators currently run it as distribution: "" — so without this, adopting distribution: talos would silently drop all control plane receivers and change the kube-proxy scrape failure log level. With it, rendering distribution: talos is byte-identical to distribution: "" apart from the removed volume and mount.

Link to Splunk Idea: N/A — bug fix for an unsupported-distribution failure, not a feature request.

Testing: 
- helm lint passes.
- Deployed to a live 4-node Talos v1.13.9 cluster (containerd 2.2.7, Kubernetes v1.36.3) with metrics enabled. All agent pods are healthy with no os-release volume on the DaemonSet, and kubelet_stats collecting normally. 

No unit tests added — no existing test covers the os-release mount or enumerates the distribution enum, and this adds no new logic branch beyond one enum value. Happy to add one if you'd like coverage.

Documentation: 
- docs/advanced-configuration.md — added talos to the supported distribution list, plus a short "Talos Linux" subsection explaining the mount, that hostmetrics still reports OS details, and that clusterName must be set since it is not auto-detected.
- helm-charts/splunk-otel-collector/values.yaml — added # - "talos" (Talos Linux) to the distribution comment block.
- examples/distribution-talos/README.md — new example, matching the other seven distribution examples.
- .chloggen/skip-osrelease-mount-talos.yaml — bug_fix / agent.


buchanandc:

.chloggen/skip-osrelease-mount-talos.yaml
```diff
@@ -0,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
jinja2, line 2:
```suggestion
change_type: enhancement
```

PR #2610, opened 2026-09-18
Add optional Splunk agent-not-ready startup taint support

husne-fairwinds:
This adds `agent.removeNotReadyTaint` so the agent can land on a node with the `splunk.net/agent-not-ready` NoSchedule taint and take that taint off after it schedules. Enable it with `agent.removeNotReadyTaint.enabled=true`. Off by default.
Enabling this also grants `nodes/patch`.


husne-fairwinds:

